PR #21, opened 2023-04-30
feat: add pylint and black as poetry dev dependencies

collindutter:


collindutter:
Flake8 might be preferred after pylint.

PR #24, opened 2023-05-06
feat: add & shift operators for Workflows & Tasks

atcol:
Having used DAG-based Python libraries for very large workflows such as Hera & Argo Workflows I thought the following would be neat to add to griptape.

This PR adds support for `+` for `Worklfows`:

```python
workflow + task == workflow.add_task(task)
```

 and `>>` and `<<` for assembling child/parent task relationships:

```python
task1 >> task2 == task.add_child(task2)
task3 << task4 == task4.add_parent(task3)
```

Lists are also supported for workflow `+`.

### DIFF
--- a/griptape/structures/pipeline.py
+++ b/griptape/structures/pipeline.py
@@ -22,6 +22,9 @@ class Pipeline(StructureWithMemory):
     def finished_tasks(self) -> list[BaseTask]:
         return [s for s in self.tasks if s.is_finished()]
 
+    def __add__(self, other: BaseTask) -> BaseTask:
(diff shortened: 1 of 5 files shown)


vasinov (changes requested):
Nice! Could you also add those to `Pipeline`?

atcol:
> Nice! Could you also add those to `Pipeline`?

Oh good idea. Yep I'll do that!

atcol:
Done (I squashed the commits).

vasinov (approved):

Issue #43, opened 2023-05-16
AwsLambdaExecutor


vasinov:
We want to enable developers who want to deploy tools in the cloud serverlessly by introducing `AwsLambdaExecutor`—one of the first serverless platform integrations.

atcol:
I'd like to pick this up. Do you have further details on how this should be structured / approached?

vasinov:
I think @collindutter might've done some work on this already. @collindutter, could you share some details? Thanks!

collindutter:
Yes, I already created an executor (https://github.com/griptape-ai/griptape/issues/43) that works in a lambda environment. Though this is basically a modification of the LocalExecutor. A proper AWSLambdaExecutor would likely invoke a lambda function configured with a tool?

kyleroche:
Can you throw a quick doc PR up along with that for testing?

collindutter:
@atcol the changes I had made on #48 are being moved to `LocalExecutor` in case you wanted to pick this up again.

atcol:
Would you like me to move that to LocalExecutor - is that what you mean?

vasinov:
Now that the executors are gone, this issue is no longer relevant.

PR #48, opened 2023-05-16
feat: add lambda executor

collindutter:


collindutter:
Relevant to https://github.com/griptape-ai/griptape/issues/43

griptape/executors/lambda_executor.py
```diff
@@ -0,0 +1,76 @@
...
+            cwd=self.tool_dir(tool),
+            stdout=None if self.verbose else subprocess.DEVNULL,
+            stderr=None if self.verbose else subprocess.DEVNULL
+        )
+
+    def run_subprocess(self, env: dict[str, str], tool_activity: callable, value: Optional[dict]) -> subprocess.CompletedProcess:
+        tool = tool_activity.__self__
+        tool_name = tool.class_name
+        input_value = value if value else ""
+        deps_dir = f'/tmp/deps/{tool_name}/' if self.install_dependencies else '/var/task/'
```
vasinov, line 62: Can we move the `/tmp/deps` and `/var/task` into a class-level const?

griptape/executors/lambda_executor.py
```diff
@@ -0,0 +1,76 @@
...
+
+        command = [
+            "python",
+            "-c",
+            f'from sys import path; path.insert(1, "{deps_dir}"); from tool import {tool_name}; print({tool_name}().{tool_activity.__name__}({input_value}))'
```
vasinov, line 67: I think those are starting to get long. We should just use string joins. Can you break this up into a list, join it with `str.join(";", <list name>)`, and apply the same pattern to other executors?

vasinov (changes requested):
So, overall, this would only work if the whole structure is executed inside of a lambda function. It doesn't seem like we need a standalone executor to achieve whatever this new class is doing. All of those changes can be applied to the `LocalExecutor` to make it more versatile.

I think we should still consider building a standalone `AwsLambdaExecutor` that runs tools inside of lambda functions separately from where structures are executed. Basically, the same idea as the `DockerExecutor`.

Issue #37, opened 2023-05-15
Get rid of the transformers warning


vasinov:
Any time Griptape is run, the following warning from the transformers library is displayed:

> None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.

We need to get rid of it either by disabling it at the library level of importing transformers lazily in tokenizers and prompt drivers.

collindutter:
I think this is roughly the change that needs to be made:

```
from transformers import logging

logging.set_verbosity_error()
```

PR #63, opened 2023-06-02
Suppress transformers pytorch warning

collindutter:
Fixes https://github.com/griptape-ai/griptape/issues/37

This warning happens in [__init__](https://github.com/huggingface/transformers/blob/f49a3453caa6fe606bb31c571423f72264152fce/src/transformers/__init__.py#L7046) so even importing the logging module (to set verbosity) triggers it. Setting the environment variable seems like the best alternative.


vasinov (commented):
I forgot that this `__init__.py` can't have anything apart from the `__path__ assignment due to how Python packages work. Will move the env var setting to all files that import transformers.

PR #67, opened 2023-06-07
add Azure OpenAI support

AlexHiesch:
Azure OpenAI support is achieved by
using the same drivers for prompts and
embeddings in case additional os env
variables are set:

API_TYPE=azure
API_BASE="https://YOUR_RESOURCE_NAME.openai.azure.com"
API_VERSION=2023-03-15-preview
ENGINE=<your deployment model>
OPENAI_API_KEY=<your api key>
ENGINE_EMBEDDING=<your deployed embedding model>
API_VERSION_EMBEDDING=2023-05-15

### DIFF
--- a/griptape/drivers/embedding/openai_embedding_driver.py
+++ b/griptape/drivers/embedding/openai_embedding_driver.py
@@ -1,3 +1,4 @@
+import os
 from typing import Optional, Union
 import numpy as np
 import openai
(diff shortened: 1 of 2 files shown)


vasinov:
After much discussion, we ended up adding Azure OpenAI support via a [standalone driver](https://docs.griptape.ai/en/latest/griptape-framework/structures/prompt-drivers/#azure-openai). Thanks for kicking this off!

Issue #50, opened 2023-05-18
Support regular inserts in `BaseVectorStorageDriver`


vasinov:
`BaseVectorStorageDriver` should support regular inserts in addition to upserts for `TextArtifact`s

PR #74, opened 2023-06-14
Support batch inserts in BaseVectorStorageDriver

trish11953:
Implemented the following methods- 
insert_text_artifacts
insert_texts
insert_vector
[Issue #50](https://github.com/griptape-ai/griptape/issues/50)

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ If you are running into `gpt-4` rate limiting issues, specify a custom number of
 
 ```python
 OpenAiPromptDriver(
+    model="gpt-4",
     max_tokens=200
 )
 ```


trish11953:
fixed

griptape/drivers/vector/base_vector_driver.py
```diff
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from concurrent import futures
 from dataclasses import dataclass
-from typing import Optional
+from typing import Optional, List, Tuple, Dict
```
vasinov, line 4: Since we support Python 3.9+ only, there is no need for importing types. Per [PEP 585](https://peps.python.org/pep-0585/), we can use built-in types directly, which we do everywhere else in the framework.

griptape/drivers/vector/base_vector_driver.py
```diff
@@ -36,9 +36,36 @@ class Entry:
...
+    def insert_text_artifacts(self, artifacts: List[Tuple[TextArtifact, str, Optional[dict]]], **kwargs) -> None:
+        with self.futures_executor as executor:
+            utils.execute_futures_dict({
+                artifact[1]:
+                    executor.submit(self.upsert_text_artifact, artifact[0], artifact[1],
+                                    artifact[2] if len(artifact) > 2 else None, **kwargs)
+                for artifact in artifacts
+            })
+
+    def insert_texts(self, texts: List[Tuple[str, str, Optional[str], Optional[dict]]], **kwargs) -> None:
```
vasinov, line 48: Don't need this method anymore.

griptape/drivers/vector/base_vector_driver.py
```diff
@@ -36,9 +36,36 @@ class Entry:
...
+        with self.futures_executor as executor:
+            utils.execute_futures_dict({
+                text[1]:
+                    executor.submit(self.upsert_text, text[0], text[1], text[2] if len(text) > 2 else None,
+                                    text[3] if len(text) > 3 else None, **kwargs)
+                for text in texts
+            })
+
+    def insert_vectors(self, vectors: List[Tuple[List[float], str, Optional[str], Optional[dict]]], **kwargs) -> None:
```
vasinov, line 57: Don't need this method anymore.

vasinov (changes requested):
Thanks for the PR! Because of the recent changes to the framework, the original issue was a bit outdated. I updated it and reduced the scope :)

trish11953:
Should we merge the branches?

griptape/drivers/vector/base_vector_driver.py
```diff
@@ -36,6 +36,15 @@ class Entry:
         kw_only=True
     )
 
+    def insert_text_artifacts(self, artifacts: list[tuple[TextArtifact, str, Optional[dict]]], **kwargs) -> None:
```
vasinov, line 39: The signature should be the same as `upsert_text_artifacts`. Is there a reason you decided to go for a list of tuples?

griptape/drivers/vector/base_vector_driver.py
```diff
@@ -36,6 +36,15 @@ class Entry:
...
+        with self.futures_executor as executor:
+            utils.execute_futures_dict({
+                artifact[1]:
+                    executor.submit(self.upsert_text_artifact, artifact[0], artifact[1],
+                                    artifact[2] if len(artifact) > 2 else None, **kwargs)
```
vasinov, line 44: This should be similar to the `upsert_text_artifacts` logic except for when the artifact is already in the namespace. Only in that case it should be inserted. Perhaps, we need another abstract method (similar to `load_entry`) that checks whether a vector with a specific ID and namespace already exist in the vector store.

vasinov (changes requested):

PR #78, opened 2023-06-23
Event Observer

collindutter:
~~Missing tests~~


griptape/drivers/prompt/base_prompt_driver.py
```diff
@@ -19,10 +21,14 @@ class BasePromptDriver(ABC):
     temperature: float = field(default=0.1, kw_only=True)
     model: str
     tokenizer: BaseTokenizer
+    structure: Optional[Structure] = field(default=None, init=False)
```
vasinov, line 24: No need for `init=False` here.

griptape/drivers/prompt/base_prompt_driver.py
```diff
@@ -19,10 +21,14 @@ class BasePromptDriver(ABC):
...
 
     def run(self, value: str) -> TextArtifact:
         for attempt in range(0, self.max_retries + 1):
             try:
+                token_count = self.tokenizer.token_count(value)
```
vasinov, line 29: Move this under `if`, so it's only executed if there is a structure.

griptape/events/base_event.py
```diff
@@ -0,0 +1,6 @@
+from attr import define
+
+
+@define
+class BaseEvent:
```
vasinov, line 5: Should be `BaseEvent(ABC)`.

griptape/events/prompt_event.py
```diff
@@ -0,0 +1,6 @@
+from attrs import define, field
+from griptape.events.base_event import BaseEvent
+
+@define
+class PromptEvent(BaseEvent):
+    token_count = field(kw_only=True)
```
vasinov, line 6: Add type.

vasinov (changes requested):

vasinov (approved):

Issue #42, opened 2023-05-16
Memory should auto-load when instantiated


vasinov:
It would be useful not to have to load memory by hand after object initialization. I propose adding memory auto-loading logic in `BaseMemoryDriver().__attrs_post_init__()`.

PR #83, opened 2023-07-03
Add ability to autoload conversation memory

collindutter:
Addresses https://github.com/griptape-ai/griptape/issues/42

### DIFF
--- a/griptape/memory/structure/conversation_memory.py
+++ b/griptape/memory/structure/conversation_memory.py
@@ -20,8 +20,7 @@ class ConversationMemory:
 
     def __attrs_post_init__(self) -> None:
         if self.driver and self.autoload:
-            autoloaded_memory = self.driver.load()
-            self.runs = autoloaded_memory.runs
+            [self.add_run(r) for r in self.driver.load().runs]
 
     def add_run(self, run: Run) -> ConversationMemory:
         self.before_add_run()
--- a/griptape/memory/structure/conversation_memory.py
+++ b/griptape/memory/structure/conversation_memory.py
@@ -16,7 +16,7 @@ class ConversationMemory:
     driver: Optional[BaseConversationMemoryDriver] = field(default=None, kw_only=True)
     runs: list[Run] = field(factory=list, kw_only=True)
     structure: Structure = field(init=False)
-    autoload = field(default=False, kw_only=True)
+    autoload: bool = field(default=True, kw_only=True)
 
     def __attrs_post_init__(self) -> None:
         if self.driver and self.autoload:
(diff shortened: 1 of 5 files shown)


griptape/memory/structure/conversation_memory.py
```diff
@@ -16,6 +16,12 @@ class ConversationMemory:
     driver: Optional[BaseConversationMemoryDriver] = field(default=None, kw_only=True)
     runs: list[Run] = field(factory=list, kw_only=True)
     structure: Structure = field(init=False)
+    autoload = field(default=False, kw_only=True)
```
collindutter, line 19: @vasinov do you think it should default to `True`? Not sure if having to specify `autoload=True` takes the "auto" out of autoloading 😅

griptape/memory/structure/conversation_memory.py
```diff
@@ -16,6 +16,12 @@ class ConversationMemory:
...
+
+    def __attrs_post_init__(self) -> None:
+        if self.driver and self.autoload:
+            autoloaded_memory = self.driver.load()
+            self.runs = autoloaded_memory.runs
```
collindutter, line 24: @vasinov should any of the other fields be loaded in?

griptape/memory/structure/conversation_memory.py
vasinov: I _think_ it makes sense to set to `True` by default.

griptape/memory/structure/conversation_memory.py
vasinov: We should use `add_run`, so that before and after hooks get triggered correctly:

```python
[self.add_run(r) for r in self.driver.load().runs]
```

vasinov: We should use `add_run`, so that before and after hooks get triggered correctly:

```python
[self.add_run(r) for r in self.driver.load().runs]
```

griptape/memory/structure/conversation_memory.py
vasinov: I _think_ it makes sense to set to `True` by default...

And let's add `Bool` to autoload :)

vasinov (changes requested):

vasinov (approved):

griptape/memory/structure/conversation_memory.py
```diff
@@ -16,6 +16,11 @@ class ConversationMemory:
     driver: Optional[BaseConversationMemoryDriver] = field(default=None, kw_only=True)
     runs: list[Run] = field(factory=list, kw_only=True)
     structure: Structure = field(init=False)
+    autoload: bool = field(default=True, kw_only=True)
+
+    def __attrs_post_init__(self) -> None:
+        if self.driver and self.autoload:
+            [self.add_run(r) for r in self.driver.load().runs]
```
vasinov, line 23: Perhaps we should change the return type `load()` to `Optional[ConversationMemory]` and check here if it's `None` or not.

vasinov (changes requested):

vasinov (approved):

Issue #39, opened 2023-05-16
Memory should store structure task results


vasinov:
There are some scenarios where it's useful to capture not just structure run inputs and outputs but also intermediate task and substask run inputs and outputs.

To store tasks and subtasks in memory, we'll need to add Marshmallow models for serialization and deserialization and integrate them with memory.

PR #84, opened 2023-07-04
Store subtask information in memory

collindutter:
Work in progress for https://github.com/griptape-ai/griptape/issues/39

Still need to modify pipelines to store intermediary task results before adding subtasks.

### DIFF
--- a/griptape/schemas/__init__.py
+++ b/griptape/schemas/__init__.py
@@ -10,6 +10,7 @@ from .artifacts.blob_artifact_schema import BlobArtifactSchema
 from .artifacts.csv_row_artifact_schema import CsvRowArtifactSchema
 
 from .memory.run_schema import RunSchema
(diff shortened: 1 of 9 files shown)


griptape/schemas/__init__.py
```diff
@@ -27,6 +28,7 @@
     "CsvRowArtifactSchema",
 
     "RunSchema",
```
collindutter, line 30: @vasinov should this change to `TaskRunSchema`?

collindutter:
@vasinov if we take this current approach, I'm not sure how to best add to pipelines. Do we use the subtasks of the first task? The last task? I think pipeline memory first needs to be improved to include the runs of all tasks.

https://github.com/griptape-ai/griptape/blob/ee7f5006224038bb0a47cd1aa10fe040bfdd1836/griptape/structures/pipeline.py#L55-L61

PR #88, opened 2023-07-05
Fix autoloading when local memory file doesn't exist

collindutter:


griptape/drivers/memory/conversation/local_conversation_memory_driver.py
```diff
@@ -13,9 +13,12 @@ def store(self, memory: ConversationMemory) -> None:
             file.write(memory.to_json())
 
     def load(self) -> Optional[ConversationMemory]:
-        with open(self.file_path, "r") as file:
-            memory = ConversationMemory.from_json(file.read())
+        try:
```
vasinov, line 16: Let's use `os.path.exists`, so it's cleaner.

vasinov (changes requested):

vasinov (approved):

PR #90, opened 2023-07-07
Add Prompt Driver for the text-generation-webui

stavsap:
Add support for PromptDriver for the text-generation-webui API

https://github.com/oobabooga/text-generation-webui

### DIFF
--- a/griptape/drivers/prompt/text_gen_driver.py
+++ b/griptape/drivers/prompt/text_gen_driver.py
@@ -1,7 +1,8 @@
 import copy
 import requests
 from attr import define, field
-
+from typing import Optional
+from urllib.parse import urljoin
 from griptape.artifacts import TextArtifact
 from griptape.drivers import BasePromptDriver
 
@@ -14,7 +15,7 @@ class TextGenPromptDriver(BasePromptDriver):
     params usage example:
     https://github.com/oobabooga/text-generation-webui/blob/main/api-examples/api-example.py
     """
-    preset: str = field(default=None, kw_only=True)
+    preset: Optional[str] = field(default=None, kw_only=True)
     params: dict = field(factory=dict, kw_only=True)
 
     model_url: str = field(default="http://localhost:5000", kw_only=True)
@@ -25,7 +26,7 @@ class TextGenPromptDriver(BasePromptDriver):
 
     def try_run(self, value: str) -> TextArtifact:
 
-        url = f"{self.model_url}{self.generate_uri}"
+        url = urljoin(self.model_url, self.generate_uri)
 
         if self.preset is None:
             request = copy.deepcopy(self.params)
--- a/griptape/drivers/prompt/text_gen_driver.py
+++ b/griptape/drivers/prompt/text_gen_driver.py
@@ -1,0 +1,43 @@
+import copy
+import requests
+from attr import define, field
+
+from griptape.artifacts import TextArtifact
+from griptape.drivers import BasePromptDriver
+
+
+@define
+class TextGenPromptDriver(BasePromptDriver):
+    """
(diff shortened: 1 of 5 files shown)


griptape/drivers/prompt/text_gen_driver.py
```diff
@@ -0,0 +1,43 @@
...
+
+@define
+class TextGenPromptDriver(BasePromptDriver):
+    """
+    Prompt Driver for the text-generation-webui https://github.com/oobabooga/text-generation-webui
+
+    params usage example:
+    https://github.com/oobabooga/text-generation-webui/blob/main/api-examples/api-example.py
+    """
+    preset: str = field(default=None, kw_only=True)
```
vasinov, line 17: `str` -> `Optional[str]`

griptape/drivers/prompt/text_gen_driver.py
```diff
@@ -0,0 +1,43 @@
...
+    params: dict = field(factory=dict, kw_only=True)
+
+    model_url: str = field(default="http://localhost:5000", kw_only=True)
+    generate_uri: str = field(default="/api/v1/generate", kw_only=True)
+
+    model = field(default=None, kw_only=True)
+    tokenizer = field(default=None, kw_only=True)
```
vasinov, line 24: The tokenizer shouldn't default to `None` here as it's required by other parts of the framework. We can make it a required param instead.

griptape/drivers/prompt/text_gen_driver.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def try_run(self, value: str) -> TextArtifact:
+
+        url = f"{self.model_url}{self.generate_uri}"
```
vasinov, line 28: Let's use `urljoin` from `urllib.parse`.

griptape/drivers/prompt/text_gen_driver.py
```diff
@@ -0,0 +1,43 @@
...
+
+        if self.preset is None:
+            request = copy.deepcopy(self.params)
```
vasinov, line 31: Why do we need to copy `params`?

vasinov (changes requested):
Thanks for the PR! Could you also add an example of how it can be used to the [docs repo](https://github.com/griptape-ai/griptape-docs)? Thanks!

griptape/drivers/prompt/text_gen_driver.py
stavsap: the copy is used because the params used as a request body with addition of the prompt. dont want to add the prompt to the params field. 
the copy result is the body of the request + prompt.

griptape/drivers/prompt/text_gen_driver.py
stavsap: i wasn't sure what to set here, i saw others do have, but i dont know which one to set for text gen, based on what the tokenizer is paired? for example for opanai why tiktoken is used?

griptape/drivers/prompt/text_gen_driver.py
stavsap: fixed

griptape/drivers/prompt/text_gen_driver.py
stavsap: fixed

griptape/drivers/prompt/text_gen_driver.py
vasinov: The tokenizer here should match the tokenizer the underlying model is using. OpenAI prompt driver is paired with tiktoken because that's what all OpenAI models use.

vasinov (approved):

PR #92, opened 2023-07-10
Adding Amazon Redshift SQL Driver

zachgiordano:
### Description

Adding an Amazon Redshift SQL Driver for use with the SqlClient tool.

Example with a Redshift Serverless workgroup:

```
sql_driver=AmazonRedshiftSqlDriver(
            database="dev",
            workgroup_name="dev"
)
```

### DIFF
--- a/griptape/drivers/sql/amazon_redshift_sql_driver.py
+++ b/griptape/drivers/sql/amazon_redshift_sql_driver.py
@@ -2,54 +2,62 @@ import time
 from typing import Optional
 import boto3
 from griptape.drivers import BaseSqlDriver
-from attr import define, field
+from attr import Factory, define, field
 
 
 @define
 class AmazonRedshiftSqlDriver(BaseSqlDriver):
     database: str = field(kw_only=True)
-    cluster_identifier: str = field(default=None, kw_only=True)
-    workgroup_name: str = field(default=None, kw_only=True)
-    db_user: str = field(default=None, kw_only=True)
-    secret_arn: str = field(default=None, kw_only=True)
-    region_name: str = field(default="us-east-1", kw_only=True)
-    wait_sec: float = field(default=0.3, kw_only=True)
-    client = field(default=None, kw_only=True)
+    session: boto3.session = field(kw_only=True)
+    cluster_identifier: Optional[str] = field(default=None, kw_only=True)
+    workgroup_name: Optional[str] = field(default=None, kw_only=True)
+    db_user: Optional[str] = field(default=None, kw_only=True)
+    database_credentials_secret_arn: Optional[str] = field(default=None, kw_only=True)
+    wait_for_query_completion_sec: Optional[float] = field(default=0.3, kw_only=True)
+    client: boto3.client = field(
+        default=Factory(
+            lambda self: self.session.client("redshift-data"), takes_self=True
+        ),
+        kw_only=True,
+    )
 
-    def __attrs_post_init__(self):
-        if self.client == None:
-            session = boto3.Session(region_name=self.region_name)
-            self.client = session.client("redshift-data")
+    @workgroup_name.validator
+    def validate_params(self, _, workgroup_name: Optional[str]) -> None:
+        if not self.cluster_identifier and not self.workgroup_name:
+            raise ValueError(
+                "Provide a value for one of `cluster_identifier` or `workgroup_name`"
+            )
+        if self.cluster_identifier and self.workgroup_name:
+            raise ValueError(
+                "Provide a value for either `cluster_identifier` or `workgroup_name`, but not both"
+            )
 
-    @staticmethod
-    def _process_rows_from_records(records) -> list[list]:
-        rows = []
-        for r in records:
-            tmp = [c[list(c.keys())[0]] for c in r]
-            rows.append(tmp)
-        return rows
+    @classmethod
+    def _process_rows_from_records(cls, records) -> list[list]:
+        return [[c[list(c.keys())[0]] for c in r] for r in records]
 
-    @staticmethod
+    @classmethod
     def _process_cells_from_rows_and_columns(
-        columns: list, rows: list[list]
+        cls, columns: list, rows: list[list]
     ) -> list[dict[str, any]]:
         return [{column: r[idx] for idx, column in enumerate(columns)} for r in rows]
 
-    @staticmethod
-    def _process_columns_from_column_metadata(meta) -> list:
+    @classmethod
+    def _process_columns_from_column_metadata(cls, meta) -> list:
         return [k["name"] for k in meta]
 
-    @staticmethod
-    def _post_process(meta, records) -> list[dict[str, any]]:
-        columns = AmazonRedshiftSqlDriver._process_columns_from_column_metadata(meta)
-        rows = AmazonRedshiftSqlDriver._process_rows_from_records(records)
-        return AmazonRedshiftSqlDriver._process_cells_from_rows_and_columns(
-            columns, rows
-        )
+    @classmethod
+    def _post_process(cls, meta, records) -> list[dict[str, any]]:
+        columns = cls._process_columns_from_column_metadata(meta)
+        rows = cls._process_rows_from_records(records)
+        return cls._process_cells_from_rows_and_columns(columns, rows)
 
     def execute_query(self, query: str) -> Optional[list[BaseSqlDriver.RowResult]]:
         rows = self.execute_query_raw(query)
-        return [BaseSqlDriver.RowResult(row) for row in rows]
+        if rows:
+            return [BaseSqlDriver.RowResult(row) for row in rows]
+        else:
+            return None
 
     def execute_query_raw(self, query: str) -> Optional[list[dict[str, any]]]:
         function_kwargs = {"Sql": query, "Database": self.database}
@@ -59,8 +67,8 @@ class AmazonRedshiftSqlDriver(BaseSqlDriver):
             function_kwargs["ClusterIdentifier"] = self.cluster_identifier
         if self.db_user:
             function_kwargs["DbUser"] = self.db_user
-        if self.secret_arn:
-            function_kwargs["SecretArn"] = self.secret_arn
+        if self.database_credentials_secret_arn:
+            function_kwargs["SecretArn"] = self.database_credentials_secret_arn
 
         response = self.client.execute_statement(**function_kwargs)
         response_id = response["Id"]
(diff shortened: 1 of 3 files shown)


griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,106 @@
...
+from griptape.drivers import BaseSqlDriver
+from attr import define, field
+
+
+@define
+class AmazonRedshiftSqlDriver(BaseSqlDriver):
+    database: str = field(kw_only=True)
+    cluster_identifier: str = field(default=None, kw_only=True)
+    workgroup_name: str = field(default=None, kw_only=True)
+    db_user: str = field(default=None, kw_only=True)
```
vasinov, line 13: Should probably not have defaults.

For other cases when we do want the default to be `None`, let's always wrap the type into `Optional`. In this particular case it would be `Optional[str]`.

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,106 @@
...
+    secret_arn: str = field(default=None, kw_only=True)
+    region_name: str = field(default="us-east-1", kw_only=True)
```
vasinov, line 15: How do secret ARNs work here? Can we swap both of those out with session? (similar to how we [do it](https://github.com/griptape-ai/griptape-tools/blob/main/griptape/tools/base_aws_client.py#L11) in tools)

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,106 @@
...
+    wait_sec: float = field(default=0.3, kw_only=True)
```
vasinov, line 16: Let's give it a more descriptive name like `wait_query_completion_sec` or something better.

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,106 @@
...
+    client = field(default=None, kw_only=True)
```
vasinov, line 17: If we use session then we can instantiate a default client from that like we [do it](https://github.com/griptape-ai/griptape-tools/blob/main/griptape/tools/aws_s3_client/tool.py#L12) in tools.

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def __attrs_post_init__(self):
+        if self.client == None:
+            session = boto3.Session(region_name=self.region_name)
+            self.client = session.client("redshift-data")
```
vasinov, line 22: If use session, then this becomes unnecessary also.

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @staticmethod
```
vasinov, line 24: I know that it's technically the right way to do it but it results in too many `AmazonRedshiftSqlDriver`s all over the code. We could make it a class method instead:

```python
@classmethod
def _process_rows_from_records(cls, records) -> list[list]:
   ...
```

Then it can be called from other class or instance methods by referring to `cls` or `self`.

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,106 @@
...
+    def _process_rows_from_records(records) -> list[list]:
+        rows = []
+        for r in records:
+            tmp = [c[list(c.keys())[0]] for c in r]
+            rows.append(tmp)
```
vasinov, line 29: Just make stuff like this a single statement:

```python
rows.append(
    [c[list(c.keys())[0]] for c in r]
)
```

or better yet replace the whole thing with a one-liner:

```python
return [c[list(c.keys())[0]] for r in records for c in r]
```

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @staticmethod
+    def _post_process(meta, records) -> list[dict[str, any]]:
+        columns = AmazonRedshiftSqlDriver._process_columns_from_column_metadata(meta)
+        rows = AmazonRedshiftSqlDriver._process_rows_from_records(records)
+        return AmazonRedshiftSqlDriver._process_cells_from_rows_and_columns(
+            columns, rows
+        )
+
+    def execute_query(self, query: str) -> Optional[list[BaseSqlDriver.RowResult]]:
```
vasinov, line 50: The output here will always be a `list`, not an `Optional`.

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,106 @@
...
+                statement_result = self.client.get_statement_result(
+                    Id=response_id, NextToken=statement_result["NextToken"]
+                )
+                results = results + response.get("Records", [])
+
+            return AmazonRedshiftSqlDriver._post_process(
+                statement_result["ColumnMetadata"], results
+            )
+
+        if statement["Status"] in ["FAILED", "ABORTED"]:
```
vasinov, line 88: `elif` for readability

tests/unit/drivers/sql/test_amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,163 @@
+import pytest
+import boto3
+from botocore.stub import Stubber
+from griptape.drivers import BaseSqlDriver, AmazonRedshiftSqlDriver
+
+
+class TestAmazonRedshiftSqlDriver:
```
vasinov, line 7: An impressive set of tests! :)

vasinov (changes requested):

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: A SecretArn can be provided which hosts the Database credentials. That is separate from the AWS credentials for calling the Redshift Data API via boto. Yes, we can swap out `region_name` and just use session, but keep `secret_arn` as an optional parameter.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/redshift-data/client/execute_statement.html#RedshiftDataAPIService.Client.execute_statement

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: Sounds good! I'll update this

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: Got it, I like that. I'll update this to use `session`. I have the `client` as an optional parameter so that the unit tests can pass in a stubbed client.

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: Yes, I'll get rid of the `session` creation here. I plan to keep the `client` as an optional parameter, and create it here if not supplied.

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: Sounds good, I'll update those to be classmethods

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: Cool, I'll update this to the one-liner 😃

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: Good catch, I'll update this to return `None` if there are no rows, like the [SqlDriver](https://github.com/griptape-ai/griptape/blob/main/griptape/drivers/sql/sql_driver.py#L24)

griptape/drivers/sql/amazon_redshift_sql_driver.py
vasinov: No need to supply it here either. Just use a default `Factory` in the field constructor.

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: Makes sense! I'll update the params to use Optional, and create a validator to check that one of `cluster_identifier` or `workgroup_name` are supplied

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,112 @@
...
+from attr import Factory, define, field
+
+
+@define
+class AmazonRedshiftSqlDriver(BaseSqlDriver):
+    database: str = field(kw_only=True)
+    session: boto3.session = field(kw_only=True)
+    cluster_identifier: Optional[str] = field(default=None, kw_only=True)
+    workgroup_name: Optional[str] = field(default=None, kw_only=True)
+    db_user: Optional[str] = field(default=None, kw_only=True)
```
vasinov, line 14: Should this also be required? Or are there cases where it can be `None`?

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,112 @@
...
+    database_credentials_secret_arn: Optional[str] = field(default=None, kw_only=True)
+    wait_for_query_completion_sec: Optional[float] = field(default=0.3, kw_only=True)
```
vasinov, line 16: Remove `Optional`.

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -0,0 +1,112 @@
...
+        kw_only=True,
+    )
+
+    @workgroup_name.validator
+    def validate_params(self, _, workgroup_name: Optional[str]) -> None:
+        if not self.cluster_identifier and not self.workgroup_name:
+            raise ValueError(
+                "Provide a value for one of `cluster_identifier` or `workgroup_name`"
+            )
+        if self.cluster_identifier and self.workgroup_name:
```
vasinov, line 30: `elif` for readability

vasinov (changes requested):
Apart from a couple of nits, looks good to me!

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: The db_user is optional and can be None. If it is None, then the [database user name is derived from the IAM identity](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/redshift-data/client/execute_statement.html).

vasinov: Gotcha

griptape/drivers/sql/amazon_redshift_sql_driver.py
zachgiordano: Will do!

PR #95, opened 2023-07-11
Refactor test table

zachgiordano:
### Details

Small refactor to test cases contrived data tables


kyleroche (approved):
when I'm not CEOing, I review code

PR #96, opened 2023-07-11
Adding Snowflake SQL Driver

zachgiordano:
### Details

Adding a Snowflake SQL Driver

Example usage:

```
import snowflake.connector
from snowflake.connector import SnowflakeConnection

PWD = os.environ.get('PWD')

def get_snowflake_connection() -> SnowflakeConnection:
    return snowflake.connector.connect(
                account = 'account',
                user = 'user',
                password = PWD,
                database = 'database',
                schema = 'schema',
    )

sql_driver=SnowflakeSqlDriver(
            snowflake_connection_function=get_snowflake_connection
        )
```

Tested with the SqlClient tool and `gpt-4`

### DIFF
--- a/griptape/drivers/sql/snowflake_sql_driver.py
+++ b/griptape/drivers/sql/snowflake_sql_driver.py
@@ -9,30 +9,26 @@ from snowflake.connector import SnowflakeConnection
 
 @define
 class SnowflakeSqlDriver(BaseSqlDriver):
-    snowflake_connection_function: Callable[[], SnowflakeConnection] = field(
-        kw_only=True
-    )
+    connection_func: Callable[[], SnowflakeConnection] = field(kw_only=True)
     engine: Engine = field(
         default=Factory(
             # Creator bypasses the URL param
             # https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine.params.creator
             lambda self: create_engine(
-                "snowflake://not@used/db", creator=self.snowflake_connection_function
+                "snowflake://not@used/db", creator=self.connection_func
             ),
             takes_self=True,
         ),
         kw_only=True,
     )
 
-    @snowflake_connection_function.validator
-    def validate_snowflake_connection_function(
-        self, _, snowflake_connection_function: Callable[[], SnowflakeConnection]
+    @connection_func.validator
+    def validate_connection_func(
+        self, _, connection_func: Callable[[], SnowflakeConnection]
     ) -> None:
-        snowflake_connection = snowflake_connection_function()
+        snowflake_connection = connection_func()
         if not isinstance(snowflake_connection, SnowflakeConnection):
-            raise ValueError(
-                "The snowflake_connection_function must return a SnowflakeConnection"
-            )
+            raise ValueError("The connection_func must return a SnowflakeConnection")
         if not snowflake_connection.schema or not snowflake_connection.database:
             raise ValueError(
                 "Provide a schema and database for the Snowflake connection"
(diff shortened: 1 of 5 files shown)


griptape/drivers/sql/snowflake_sql_driver.py
```diff
@@ -0,0 +1,85 @@
...
+from attr import Factory, define, field
+from sqlalchemy import create_engine, text, MetaData, Table
+from sqlalchemy.engine import Engine
+from sqlalchemy.exc import NoSuchTableError
+from snowflake.connector import SnowflakeConnection
+
+
+@define
+class SnowflakeSqlDriver(BaseSqlDriver):
+    snowflake_connection_function: Callable[[], SnowflakeConnection] = field(
```
vasinov, line 12: This feels a bit long. Should we shorten it? May be `connection_func`?

griptape/drivers/sql/snowflake_sql_driver.py
```diff
@@ -0,0 +1,85 @@
...
+    @snowflake_connection_function.validator
+    def validate_snowflake_connection_function(
+        self, _, snowflake_connection_function: Callable[[], SnowflakeConnection]
+    ) -> None:
+        snowflake_connection = snowflake_connection_function()
+        if not isinstance(snowflake_connection, SnowflakeConnection):
+            raise ValueError(
+                "The snowflake_connection_function must return a SnowflakeConnection"
+            )
+        if not snowflake_connection.schema or not snowflake_connection.database:
```
vasinov, line 36: Is the schema name really always required in Snowflake?

griptape/drivers/sql/snowflake_sql_driver.py
```diff
@@ -0,0 +1,85 @@
...
+                results = con.execute(text(query))
+
+                if results.returns_rows:
+                    results = [
+                        {column: value for column, value in result.items()}
+                        for result in results
+                    ]
+            finally:
+                con.close()
+        self.engine.dispose()
```
vasinov, line 67: Is `dispose` not an automatic part of the `with` structure?

tests/unit/drivers/sql/test_snowflake_sql_driver.py
```diff
@@ -0,0 +1,146 @@
...
+import pytest
+from sqlalchemy import create_engine
+from snowflake.connector import SnowflakeConnection
+from griptape.drivers import BaseSqlDriver, SnowflakeSqlDriver
+
+
+class TestSnowflakeSqlDriver:
+    TEST_ROWS = [
+        {"first_name": "Tony", "last_name": "Hawk"},
+        {"first_name": "Bob", "last_name": "Ross"},
```
vasinov, line 12: 🎨

vasinov (changes requested):

griptape/drivers/sql/snowflake_sql_driver.py
zachgiordano: I like that, yes I'll shorten it in the next commit

griptape/drivers/sql/snowflake_sql_driver.py
zachgiordano: It seems that the sqlalchemy snowflake dialect requires the fully qualified object, so we need to require the schema and database name here.

griptape/drivers/sql/snowflake_sql_driver.py
zachgiordano: Discussed offline: 
* The Python context manager will automatically call `Connection.close()`, so the try block can be removed
* The `engine.dispose()` method is unnecessary based on the [docs](https://docs.sqlalchemy.org/en/14/core/connections.html#engine-disposal), so that will be removed

vasinov (approved):

PR #97, opened 2023-07-12
Marqo Vector Store Driver Implementation

RaynorChavez:
This PR introduces a vectorstore integration for Marqo as well as accompanying unit tests.

Integration tests will be provided in a separate PR.

A PR into `griptape-docs` will also be created referencing this one.


griptape/drivers/vector/marqo_vector_store_driver.py
```diff
@@ -0,0 +1,253 @@
...
+@define
+class MarqoVectorStoreDriver(BaseVectorStoreDriver):
+    api_key: str = field(kw_only=True)
+    url: str = field(kw_only=True)
+    mq: marqo.Client = field(kw_only=True)
+    index: str = field(kw_only=True)
+
+    def __attrs_post_init__(self):
+        """Initialize the Marqo client with the given API key and URL."""
+        self.mq = marqo.Client(self.url, self.api_key)
```
vasinov, line 19: Could we set this one in the attribute factory? (here is an [example](https://github.com/griptape-ai/griptape/blob/06b119181f53e7fa34aabd597a8a9a4129712cd5/griptape/drivers/sql/amazon_redshift_sql_driver.py#L17))

griptape/drivers/vector/marqo_vector_store_driver.py
```diff
@@ -0,0 +1,253 @@
...
+    def set_index(self, index):
+        """Set the index for the Marqo client.
+
+        Args:
+            index (str): The index to set for the Marqo client.
+        """
+        indexes = self.get_indexes()
+
+        if index not in indexes:
+            self.create_index(index)
```
vasinov, line 31: Should we log here in case the user expects the index to exist?

vasinov (changes requested):
Great PR—thanks so much!

Could you please resolve the `poetry.lock` conflict and address a couple of minor nits before I merge?

griptape/drivers/vector/marqo_vector_store_driver.py
RaynorChavez: fixed :)

griptape/drivers/vector/marqo_vector_store_driver.py
RaynorChavez: Good idea, thanks! Just added logging.info()

vasinov (approved):

PR #109, opened 2023-07-14
Added a persistable version of LocalVectorStoreDriver. Added tests fo…

tarbaig:
Implemented a persistable child class of LocalVectorStoreDriver that uses pickle to save the internal ```entries``` dict.
Added test for the serialization and deserialisation of the new class.


vasinov (changes requested):
Thanks for the PR!

Is there a reason to use pickle here? We can just store/load the entries dict in a json file on disk; should eliminate potential code injection vulnerabilities.

tarbaig:
I guess it was just a bad habit. I’ll modify it to use json, and will add an optional argument to save it as zipped json.

tests/unit/drivers/storage/vector/test_persitable_local_vector_store_driver.py
```diff
@@ -0,0 +1,41 @@
...
+            )
+            k, v = old_vector_store_entries.popitem()
+            kn, vn = new_vector_store.entries.popitem()
+            assert v == vn
+
+    def test_save_load_json(self):
+        self.t_persistable_vector_store("json")
+
+    def test_save_load_zip(self):
+        self.t_persistable_vector_store("json")
```
zachgiordano, line 34: I think this should be passing the "zip" suffix

griptape/drivers/vector/peristable_local_vector_store_driver.py
```diff
@@ -0,0 +1,84 @@
...
+        if self.file_is_zip:
+            with ZipFile(self.file_path, write_mode) as zf:
+                json_data = json.dumps(self.entries, cls=EntryAwareJSONEncoder)
+                zf.writestr("data.json", json_data)
+        else:
+            with open(self.file_path, write_mode) as f:
+                json.dump(self.entries, f, cls=EntryAwareJSONEncoder)
+
+    def load(self) -> None:
+        """Read entries that have been serialized using pickle. Will overwrite internal entries with the loaded ones"""
```
zachgiordano, line 69: nit: This comment should be updated since this is no longer serializing with pickle

tarbaig:
I addressed the comments. Not sure why the Quality / lint job complains e.g.

```python
with open(...) as f:
    ....
```
is pretty idiomatic python . " Variable name "f" doesn't conform to snake_case naming style" might be overkill.

I can change these things i.e. "f->file", but I am not sure it increases readability.

tests/unit/drivers/storage/vector/test_persitable_local_vector_store_driver.py
tarbaig: done

griptape/drivers/vector/peristable_local_vector_store_driver.py
tarbaig: done

vasinov:
Thanks for adding the changes! Will review tomorrow. Don't worry about squashing—I squash all commits on merge.

griptape/drivers/vector/peristable_local_vector_store_driver.py
```diff
@@ -0,0 +1,83 @@
...
+                f"file must either be a json or a zip but file name is {self.file_path}"
+            )
+
+    def __attrs_post_init__(self):
+        self.file_is_zip = self.file_path.suffix == ".zip"
+
+    def store(self, overwrite=False):
+        """Serialize the entries of the vectorstore. Will not overwrite an existing file unless overwrite=True"""
+
+        write_mode = cast(Literal["w", "x"], "x" if not overwrite else "w")
```
vasinov, line 56: Can we replace this with `write_mode = "w" if overwrite else "x"`?

griptape/drivers/vector/peristable_local_vector_store_driver.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @file_path.validator
+    def _check_file_path(self, attribute, value):
+        if not value.suffix in (".json", ".zip"):
+            raise ValueError(
+                f"file must either be a json or a zip but file name is {self.file_path}"
+            )
+
+    def __attrs_post_init__(self):
+        self.file_is_zip = self.file_path.suffix == ".zip"
```
vasinov, line 51: Can we move this into a property? [HERE](https://github.com/griptape-ai/griptape/blob/main/griptape/drivers/sql/amazon_redshift_sql_driver.py#L18) is an example of how it can be implemented.

griptape/drivers/vector/peristable_local_vector_store_driver.py
```diff
@@ -0,0 +1,83 @@
...
+        if self.file_is_zip:
+            with ZipFile(self.file_path) as zf:
+                json_data = zf.read("data.json")
+                self.entries = json.loads(json_data, object_hook=as_entry)
+
+        else:
+            with open(self.file_path, "r") as f:
+                self.entries = json.load(f, object_hook=as_entry)
+
+    @classmethod
```
vasinov, line 78: Could you move this to the top, since it's a class method?

griptape/drivers/vector/peristable_local_vector_store_driver.py
```diff
@@ -0,0 +1,83 @@
...
+        return super().default(o)
+
+
+def as_entry(dct: Dict):
+
+    if "__Entry__" in dct:
+        dct.pop("__Entry__")
+        return PersistableLocalVectorStoreDriver.Entry(**dct)
+
+    return dct
```
vasinov, line 27: Could you move those inside of `PersistableLocalVectorStoreDriver`? We try to avoid using functions as well as multiple classes per file.

vasinov (changes requested):

griptape/drivers/vector/peristable_local_vector_store_driver.py
tarbaig: Sure if that is your style. I think the most idiomatic way would be to make them underscore methods / classes and not put them inside `PersistableLocalVectorStoreDriver ` , but in the end it does not matter. 

I am on vacation the next week, so change will arrive 2nd week if august.

collindutter:
Hi @tarbaig we are going to close this PR for now in favor of an implementation that will come through [this issue](https://github.com/griptape-ai/griptape/issues/241). If you have any interest in working on the other issue, we'd be happy to assist. Thank you!

PR #121, opened 2023-07-22
Add ai21 prompt driver

collindutter:


griptape/drivers/prompt/ai21_prompt_driver.py
```diff
@@ -0,0 +1,32 @@
+import ai21
+from attr import define, field, Factory
+from griptape.artifacts import TextArtifact
+from griptape.drivers import BasePromptDriver
+from griptape.tokenizers import Ai21Tokenizer
+
+
+@define
+class Ai21PromptDriver(BasePromptDriver):
```
vasinov, line 9: No need to set custom prefix or suffix for their models?

vasinov (changes requested):

griptape/drivers/prompt/ai21_prompt_driver.py
collindutter: As far as I can tell from the [docs](https://docs.ai21.com/reference/j2-complete-api-ref), no. Though they do recommend certain [prompting techniques](https://docs.ai21.com/docs/prompt-engineering) that we might want to play with once we've got driver-specific prompt templates.

griptape/tokenizers/ai21_tokenizer.py
```diff
@@ -0,0 +1,33 @@
+import ai21
+import binascii
+from attr import define, field
+from griptape.tokenizers import BaseTokenizer
+
+
+@define(frozen=True)
+class Ai21Tokenizer(BaseTokenizer):
```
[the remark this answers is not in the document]
collindutter: This originally came from the fact that their [tokenize API](https://docs.ai21.com/reference/tokenize-ref) does not return tokens as numerical representations. For instance:
```
{
  "text": "Thesearea few of my favorite",
  "tokens": [
    {
      "token": "▁These",
      "textRange": {
        "start": 0,
        "end": 5
      }
    },
    {
      "token": "area",
      "textRange": {
        "start": 5,
        "end": 9
      }
    },
    {
      "token": "▁few",
      "textRange": {
        "start": 9,
        "end": 13
      }
    },
    {
      "token": "▁of▁my▁favorite",
      "textRange": {
        "start": 13,
        "end": 28
      }
    }
  ]
}
```
So I tried encoding the tokens as integers in a way that could be decoded. Unfortunately after some more testing, this encode/decode technique does not work.

From what I can tell, Griptape only uses the `encode` method to count the number of tokens, and `decode` is not used at all. Is it necessary that we implement both?

vasinov (approved):

collindutter:
Initial testing suggests that Jurassic Ultra may not be capable of using `ToolkitTask`s. Will do more testing once https://github.com/griptape-ai/griptape/pull/183 is ready.

Issue #122, opened 2023-07-23
RateLimitError when reproducing simple Pipeline example


joelwiesmann:
**Describe the bug**
When I try to reproduce the pirate-pipeline example (https://docs.griptape.ai/en/latest/griptape-framework/structures/pipelines/) only the first task in the pipeline is executed successfully, for the 2nd it returns a ratelimit error.

**To Reproduce**
1. Go to https://docs.griptape.ai/en/latest/griptape-framework/structures/pipelines/
2. Copy & paste pirate-talk example
3. Run example
4. First task is executed, for the 2nd task an error is received:
```
attempt #1; slept for 0.0; last result: failed (RateLimitError Rate limit reached for 10KTPM-200RPM in organization xxx on tokens per min. Limit: 10000 / min. Please try again in 6ms. Contact us through our help center at help.openai.com if you continue to have issues.)
```

collindutter:
@joelwiesmann can you please try changing your [prompt driver](https://docs.griptape.ai/en/latest/griptape-framework/structures/prompt-drivers/) to `gpt-3.5-turbo` or `gpt-3.5-turbo-16k`? The rate limits are much lower on `gpt-4`. Even so, it seems odd that griptape would be bumping into the limits out of box.

@vasinov thoughts? I've noticed this behavior on my own API key.

vasinov:
@joelwiesmann, thanks for reporting the issue! Are you the only one using the API in your org? Does OpenAI return the rate limit error on the first pipeline run? I hooked up the token count utility to that example and it shows that it only used `467` tokens:

```python
from griptape import utils
from griptape.events import StartPromptEvent, FinishPromptEvent
from griptape.memory.structure import ConversationMemory
from griptape.tasks import PromptTask
from griptape.structures import Pipeline

token_counter = utils.TokenCounter()

pipeline = Pipeline(
    memory=ConversationMemory(),
    event_listeners={
        StartPromptEvent: [
            lambda e: token_counter.add_tokens(e.token_count)
        ],
        FinishPromptEvent: [
            lambda e: token_counter.add_tokens(e.token_count)
        ],
    }

)

pipeline.add_tasks(
    # take the first argument from the pipeline `run` method
    PromptTask("{{ args[0] }}"),
    # take the input from the previous task and insert it into the prompt
    PromptTask("Say the following like a pirate: {{ input }}")
)

pipeline.run("I am Scotty, who are you?")
pipeline.run("Who am I?")

[...]

PR #123, opened 2023-07-23
Prompt Driver Fixes

vasinov:
- Moved openai api param setting to chat and completion api calls
- Added `OpenAiPromptDriver.max_tokens`
- Refactored OpenAiPromptDriver params into `_base_params`
- Removed few shots from the prompt to save on tokens


vasinov:
This PR partially addresses #122

PR #137, opened 2023-07-25
Replace hardcoded prompt with Jinja template

joelwiesmann:
Replace hardcoded vector query engine prompt with Jinja template.

### DIFF
--- a/griptape/engines/query/vector_query_engine.py
+++ b/griptape/engines/query/vector_query_engine.py
@@ -3,6 +3,7 @@ from attr import define, field, Factory
 from griptape.artifacts import TextArtifact, BaseArtifact
 from griptape.drivers import BaseVectorStoreDriver, LocalVectorStoreDriver, BasePromptDriver, OpenAiPromptDriver
 from griptape.engines import BaseQueryEngine
+from griptape.utils.j2 import J2
 
 
 @define
@@ -13,6 +14,10 @@ class VectorQueryEngine(BaseQueryEngine):
     )
     prompt_driver: BasePromptDriver = field(
         default=Factory(lambda: OpenAiPromptDriver()),
+        kw_only=True
+    )
+    template_generator: J2 = field(
+        default=Factory(lambda: J2("engines/vector_query.j2")),
         kw_only=True
     )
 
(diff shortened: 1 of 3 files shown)


griptape/engines/query/vector_query_engine.py
```diff
@@ -28,28 +33,26 @@ def query(
...
-        ]
-
-        metadata = f"Metadata: {metadata if metadata else 'no metadata available'}"
-        prefix = f"{' '.join(prefix_list)}\n\n{metadata}"
-        question = f"\n\nQuestion: {query}"
-        answer = "\n\nAnswer: "
+        prefix = []
 
         for artifact in artifacts:
             next_segment = f'\n\nText segment:\n"""\n{artifact.value}\n"""'
```
vasinov, line 39: We should be able to tuck "Text segment" into a template as well and just use a for loop there:

```
{% for segment in text_segments %}
...
{% endfor %}
```

griptape/engines/query/vector_query_engine.py
```diff
@@ -28,28 +33,26 @@ def query(
...
 
-            if tokenizer.token_count(prefix + next_segment + question + answer) > tokenizer.max_tokens:
+            message = self.template_generator.render(
+                metadata=metadata or "-",
```
vasinov, line 42: Probably can keep it as `metadata` and not include it in the prompt at all if it's none:

```
{% if metadata %}Metadata: {{ metadata }}{% endif %}
```

vasinov (changes requested):
Thanks for the PR! This is a much needed improvement.

vasinov (changes requested):
This is so great! One last thing: could you please add a couple of unit tests with a custom template just to make sure text segments are appended correctly? Thanks!

joelwiesmann:
> This is so great! One last thing: could you please add a couple of unit tests with a custom template just to make sure text segments are appended correctly? Thanks!

Sure! Will have a look on that tomorrow. Will consider this in the future from the beginning :-).

joelwiesmann:
@vasinov I'm not sure what you meant by "custom template", so I added a test case to see, whether the provided metadata / text segments & question end up in the created prompt at all. This case is expected to work with whatever query template is used.

Issue #94, opened 2023-07-10
Vector driver for MongoDB


vasinov:
Developers are interested in using [MongoDB Atlas](https://www.mongodb.com/docs/atlas/getting-started/) as a vector store while using Griptape.

I propose building a `MongoDbAtlasVectorStoreDriver` that implements that functionality.

wwzeng1:
Hi! I’m one of the founders of Sweep, a github app that solves issues(like adding small features) by writing pull requests.
This looks like a good issue for Sweep https://github.com/sweepai/sweep to try. We have onboarding [instructions here](https://github.com/sweepai/sweep/blob/main/README.md#-getting-started), I’m also happy to help you onboard directly :)

PR #139, opened 2023-07-25
MongoDBAtlasVectorDriver

trish11953:
Implemented the MongoDB driver, please have a look and let me know if something seems wrong.
[Issue #94](https://github.com/griptape-ai/griptape/issues/94)


griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -0,0 +1,85 @@
+from typing import Optional, Dict, Any, List
+from griptape.drivers import BaseVectorStoreDriver
+from pymongo import MongoClient
+from attr import define, field, Factory
+import logging
+
+
+@define
+class MongoDbAtlasVectorStoreDriver(BaseVectorStoreDriver):
```
vasinov, line 9: Can you an import for this class to `drivers/__init__.py`?

vasinov (changes requested):
Could you also please add tests with mocked out MongoDB responses? You can use [Pinecone driver tests](https://github.com/griptape-ai/griptape/blob/main/tests/unit/drivers/storage/vector/test_pinecone_vector_storage_driver.py) as a reference.

vasinov:
Could also add a entry for this driver in [this docs section](https://docs.griptape.ai/en/latest/griptape-framework/data/vector-stores/) please?

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -0,0 +1,85 @@
...
+            meta: Optional[dict] = None,
+            **kwargs
+    ) -> str:
+        try:
+            document = {"vector": vector, "meta": meta, "namespace": namespace}
+            if vector_id:
+                document["_id"] = vector_id
+            result = self.collection.replace_one({"_id": vector_id}, document, upsert=True)
+            return result.upserted_id or vector_id
+        except Exception as e:
```
zachgiordano, line 39: Do we want to swallow these exceptions or allow them to raise?

vasinov: Good catch! We should probably not use try-except here (and in other methods) at all since this is a low level driver.

trish11953: removed it !

pyproject.toml
```diff
@@ -48,6 +49,8 @@ snowflake-sqlalchemy = "^1.4.7"
 pytest = "~=7.1"
 pytest-mock = "*"
 pytest-cover = "*"
+pymongo= "*"
```
vasinov, line 52: This should be under `[tool.poetry.dependencies]`

pyproject.toml
```diff
@@ -25,6 +25,7 @@ graphlib = "*"
 tiktoken = ">=0.3"
 rich = ">=13"
 stopit = "*"
+mongomock = "*"
```
vasinov, line 28: This should be under `[tool.poetry.group.test.dependencies]`.

vasinov (approved):

PR #145, opened 2023-07-27
Add pre-commit

collindutter:
Adds [pre-commit](https://pre-commit.com/) to run linting and formatting locally before commit.


vasinov (commented):
So, this runs on `pre-commit install`, correct?

PR #151, opened 2023-07-29
Knnbeta implementation for the MongoDbAltlasVectorStoreDriver

trish11953:
I implemented the query function for the MongoDbAltlasVectorStoreDriver using knnBeta and it works for a real client too!


griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -81,10 +81,38 @@ def load_entries(
...
         count: Optional[int] = None,
         namespace: Optional[str] = None,
         include_vectors: bool = False,
+        skip: Optional[int] = 0,
         **kwargs
     ) -> list[BaseVectorStoreDriver.QueryResult]:
-        raise NotImplementedError("Vector search not supported in MongoDB")
+        collection = self.get_collection()
+
+        knn_k = skip + (count if count else 10)
```
vasinov, line 93: This took me some time to parse. I wonder if we should rename `skip` to `offset` in the method, so it's more in line with the meaning behind it. Shouldn't be an issue since we already use `count` to represent `k`.

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -81,10 +81,38 @@ def load_entries(
...
+                        "path": "vector",
+                        "k": knn_k
+                    }
+                }
+            },
+            {"$skip": skip},
+        ]
+
+        if count is not None:
+            pipeline.append({"$limit": count})
```
vasinov, line 108: Based on `(count if count else 10)` above, I think this should also default to 10, right? I'd just set `knn_k` to `count if count else 10` and then set `"k"` to `knn_k + skip` on line 100 and use `knn_k` as a `$limit` value here, so it's less error-prone.

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -81,10 +81,38 @@ def load_entries(
...
+
+        cursor = collection.aggregate(pipeline)
+
+        for doc in cursor:
+            yield BaseVectorStoreDriver.QueryResult(
```
vasinov, line 113: Why use `yield` here? The method returns a `list` not a `Generator`.

vasinov (changes requested):

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -80,11 +80,58 @@ def load_entries(
...
+            include_vectors: bool = False,
+            offset: Optional[int] = 0,
+            **kwargs
     ) -> list[BaseVectorStoreDriver.QueryResult]:
-        raise NotImplementedError("Vector search not supported in MongoDB")
+        collection = self.get_collection()
+
+        # Using the embedding driver to convert the query string into a vector
+        vector = self.embedding_driver.embed_string(query)
+        print("Vector for query:", vector)
```
vasinov, line 95: Cleanup.

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -80,11 +80,58 @@ def load_entries(
...
+                    "meta": 1
+                }
+            },
+            {"$skip": offset},
+            {"$limit": knn_k}
+        ]
+
+        cursor = collection.aggregate(pipeline)
+        #print("Pipeline:", pipeline)
+        #print("Cursor:", list(cursor))
```
vasinov, line 123: Cleanup.

pyproject.toml
```diff
@@ -44,6 +44,7 @@ sqlalchemy = "~=1.0"
 sqlalchemy-redshift = "*"
 snowflake-sqlalchemy = "^1.4.7"
 pymongo= "*"
+scikit-learn = "^1.3.0"
```
vasinov, line 47: ?

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -80,11 +80,58 @@ def load_entries(
...
+        list_cursor = list(cursor)
+        for doc in list_cursor:
+            print(doc)
```
vasinov, line 126: Cleanup.

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -80,11 +80,58 @@ def load_entries(
...
+        results = [
+            BaseVectorStoreDriver.QueryResult(
+                vector=doc["vector"] if include_vectors else None,
+                score=None,  # Score is not directly provided by knnBeta
+                meta=doc["meta"],
+                namespace=namespace,  # Added namespace to QueryResult
+            )
+            for doc in list_cursor
+        ]
+        #print("In Driver results: ", results)
```
vasinov, line 136: Cleanup.

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -80,11 +80,58 @@ def load_entries(
...
+
+        # Using the embedding driver to convert the query string into a vector
+        vector = self.embedding_driver.embed_string(query)
+        print("Vector for query:", vector)
+
+        knn_k = count if count else 10
+        pipeline = [
+            {
+                "$search": {
+                    "index": "knn",  # Use the name of the k-NN index you created
```
vasinov, line 101: Why is it `knn`? According to the [docs](https://www.mongodb.com/docs/atlas/atlas-search/knn-beta/#syntax) the default index name is `default`. I think we should add another optional method parameter `index: Optional[str] = None` to the method and only set it in this dictionary if it's not `None`.

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -80,11 +80,58 @@ def load_entries(
...
+        cursor = collection.aggregate(pipeline)
+        #print("Pipeline:", pipeline)
+        #print("Cursor:", list(cursor))
+        list_cursor = list(cursor)
+        for doc in list_cursor:
+            print(doc)
+        results = [
+            BaseVectorStoreDriver.QueryResult(
+                vector=doc["vector"] if include_vectors else None,
+                score=None,  # Score is not directly provided by knnBeta
```
vasinov, line 130: I think [it is](https://www.mongodb.com/docs/atlas/atlas-search/knn-beta/#scoring) :)

vasinov (changes requested):

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -80,11 +80,56 @@ def load_entries(
...
+        collection = self.get_collection()
+
+        # Using the embedding driver to convert the query string into a vector
+        vector = self.embedding_driver.embed_string(query)
+
+        knn_k = count if count else 10
+        pipeline = [
+            {
+                "$search": {
+                    "index": index if index else "default",
```
vasinov, line 101: I was thinking of not setting it at all if it's not passed. You can just add a line after `pipeline = [...]` setting the key:

```python
if index:
    pipeline[0]["$search"]["index"] = index
```

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -80,11 +80,56 @@ def load_entries(
...
+                    "_id": 1,
+                    "vector": 1,
+                    "namespace": 1,
+                    "meta": 1,
+                    "score": {"$meta": "searchScore"}  # Include the score in the projection
+                }
+            },
+            {"$skip": offset},
+            {"$limit": knn_k}
+        ]
```
vasinov, line 120: Just realized, that according to the docs, this doesn't have to be a list, since there is only one element. Can just be a `dict` right?

vasinov (changes requested):

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -0,0 +1,76 @@
+from typing import Optional
+from griptape import utils
+from griptape.drivers import BaseVectorStoreDriver
+import redis
+from attr import define, field
+import json
+
+
+@define
+class RedisVectorStoreDriver(BaseVectorStoreDriver):
```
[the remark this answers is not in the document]
trish11953: Yes sorry about that!

griptape/drivers/vector/mongodb_vector_store_driver.py
trish11953: Im using $search and $project here so I do need the list here. It gives me an OperationError without it.
It also gives me - 'A pipeline stage specification object must contain exactly one field.' so using a dictionary would make it multiple items since I'm using $search and $project and that's not allowed.

griptape/drivers/vector/mongodb_vector_store_driver.py
trish11953: done

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -80,11 +80,56 @@ def load_entries(
...
+
+        cursor = collection.aggregate(pipeline)
+        list_cursor = list(cursor)
+        results = [
```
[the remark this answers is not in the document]
trish11953: done

tests/unit/drivers/vector/test_mongodb_vector_store_driver.py
```diff
@@ -33,13 +33,54 @@ def test_upsert_text(self, driver):
...
+        )
+
+        query_str = "some query string"
+        results = driver.query(query_str, include_vectors=True)
+        assert len(results) == len(mock_query_result)
+        for result, expected in zip(results, mock_query_result):
+            assert result.vector == expected.vector
+            assert isinstance(result, BaseVectorStoreDriver.QueryResult)
+
+    # def test_query(self, driver):
```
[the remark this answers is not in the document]
trish11953: done

vasinov (approved):

PR #154, opened 2023-08-01
RedisVectorStoreDriver Implementation

trish11953:
Just created the basic RedisVectorStoreDriver along with its unit tests. 
Still, have to implement query search and index creation.


vasinov:
Will take a look after all changes are in place. Let's submit incomplete PRs as drafts, so it's clear that there is more in-progress work (i.e., in this case, query search and index creation).

trish11953:
> Still have to do tests with a real connection.

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -98,6 +98,9 @@ def query(
         pipeline = [
```
vasinov, line 98: I think those are old changes. Can you remove them from the PR?

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -0,0 +1,113 @@
...
+class RedisVectorStoreDriver(BaseVectorStoreDriver):
+    host: str = field(kw_only=True)
+    port: int = field(kw_only=True)
+    db: int = field(kw_only=True, default=0)
+    password: Optional[str] = field(default=None, kw_only=True)
+    client: redis.StrictRedis = field(init=False)
+    index: str = field(kw_only=True)
+
+    def __attrs_post_init__(self) -> None:
+        self.client = redis.StrictRedis(
```
vasinov, line 21: I think `StrictRedis` is an old alias for `Redis`. So, you can change it to `Redis`.

vasinov: Could you also initialize it in the property factory like we do with other clients?

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -0,0 +1,113 @@
...
+            "vec_string": bytes_vector
+        }
+        if meta:
+            mapping_obj["metadata"] = json.dumps(meta)
+
+        self.client.hset(key, mapping=mapping_obj)
+        return vector_id
+
+    def load_entry(self, vector_id: str, namespace: Optional[str] = None) -> Optional[BaseVectorStoreDriver.Entry]:
+        key = f'{namespace}:{vector_id}' if namespace else vector_id
```
vasinov, line 54: DRY it up by moving into a separate method, since it's used at least twice in this class.

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -0,0 +1,113 @@
...
+        result = self.client.hgetall(key)
+
+        reversed_vector = np.frombuffer(result[b"vector"], dtype=np.float32)
+        reversed_vector_list = reversed_vector.tolist()
+
+        if result:
```
vasinov, line 60: Is `result` ever `None`? If so then code on line 57 will raise an exception.

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -0,0 +1,113 @@
...
+            value = {
+                "vector": reversed_vector_list,
+                "metadata": json.loads(result[b"metadata"]) if b"metadata" in result else None
```
vasinov, line 63: Are keys really returned as byte strings by `hgetall`?

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def load_entries(self, namespace: Optional[str] = None) -> list[BaseVectorStoreDriver.Entry]:
+        pattern = f'{namespace}:*' if namespace else '*'
+        keys = self.client.keys(pattern)
+        entries = []
+        for key in keys:
+            entries.append(self.load_entry(key.decode("utf-8"), namespace=namespace))
+        return entries
+
+    def query(self, vector: list[float], count: Optional[int] = None, namespace: Optional[str] = None, **kwargs) -> \
```
vasinov, line 82: Change it to this instead:

```python
def query(
    self, vector: list[float], count: Optional[int] = None, namespace: Optional[str] = None, **kwargs
) -> List[BaseVectorStoreDriver.QueryResult]:
```

vasinov (changes requested):
I think it looks good overall. Just left a few minor comments.

griptape/drivers/vector/redis_vector_store_driver.py
trish11953: This is already that!

griptape/drivers/vector/redis_vector_store_driver.py
trish11953: It comes back as bytes, yes

griptape/drivers/vector/redis_vector_store_driver.py
trish11953: Done!

griptape/drivers/vector/mongodb_vector_store_driver.py
trish11953: This was just me updating the branch, sorry!

griptape/drivers/vector/redis_vector_store_driver.py
trish11953: Done!

griptape/drivers/vector/redis_vector_store_driver.py
vasinov: I don't think this is resolved :)

With the latest changes, it's line 58 that will fail if `result` is `None`.

griptape/drivers/vector/redis_vector_store_driver.py
vasinov: I see this in your code:

```python
def query(self, vector: list[float], count: Optional[int] = None, namespace: Optional[str] = None, **kwargs) -> \
    List[BaseVectorStoreDriver.QueryResult]:
```

griptape/drivers/vector/mongodb_vector_store_driver.py
vasinov: I can still see those changes.

vasinov (changes requested):

griptape/drivers/vector/redis_vector_store_driver.py
trish11953: Ohh, its the new line character! changed it!

PR #181, opened 2023-08-25
RedisVectorStoreDriver Updates

trish11953:


griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -90,20 +91,47 @@ def query(
...
+        for document, vector in zip(results, vectors):
             query_results.append(
                 BaseVectorStoreDriver.QueryResult(
-                    vector=vector_float_list,
+                    vector=np.frombuffer(vector, dtype=np.float32).tolist(),
                     score=float(document['score']),
-                    meta=metadata,
-                    namespace=namespace
+                    meta=None,
+                    namespace=None
```
vasinov, line 107: Why did you change those to `None`?

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -90,20 +91,47 @@ def query(
 
         results = self.client.ft(self.index).search(query_expression, query_params).docs
 
-        query_results = []
+        pipe = self.client.pipeline()
         for document in results:
-            metadata = getattr(document, "metadata", None)
-            namespace = document.id.split(":")[0] if ":" in document.id else None
-            vector_float_list = json.loads(document["vec_string"])
+            pipe.hget(document.id, "vector")
```
vasinov, line 96: Looks like this line executes an extra request to Redis for every result. Not sure we want that. Previous implementation didn't have that problem.

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -90,20 +91,47 @@ def query(
...
+        except:
+            schema = (
+                TagField("tag"),
+                VectorField("vector",
+                            "FLAT", {
+                                "TYPE": "FLOAT32",
+                                "DIM": self.vector_dimensions,
+                                "DISTANCE_METRIC": "COSINE",
+                            }
+                            ),
```
vasinov, line 125: Formatting should look like this:

```
VectorField(
    "vector",
    "FLAT", {
        "TYPE": "FLOAT32",
        "DIM": self.vector_dimensions,
        "DISTANCE_METRIC": "COSINE",
    }
)
```

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -90,20 +91,47 @@ def query(
...
+                    meta=None,
+                    namespace=None
                 )
             )
         return query_results
 
+    def create_index(self, namespace: Optional[str] = None):
+        try:
+            self.client.ft(self.index).info()
+            print("Index already exists!")
```
vasinov, line 115: Use logging instead.

vasinov (changes requested):

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -15,6 +19,7 @@ class RedisVectorStoreDriver(BaseVectorStoreDriver):
     db: int = field(kw_only=True, default=0)
     password: Optional[str] = field(default=None, kw_only=True)
     index: str = field(kw_only=True)
+    vector_dimensions: int = field(kw_only=True, default=1536)
```
vasinov, line 22: Let's move this into `create_index` parameters, since it's only used there. Let's also not have a default, since embedding dimensions vary from model to model.

griptape/drivers/vector/redis_vector_store_driver.py
```diff
@@ -105,5 +110,29 @@ def query(
             )
         return query_results
 
+    def create_index(self, namespace: Optional[str] = None):
+        try:
+            self.client.ft(self.index).info()
+            logging.info("Index already exists!")
```
vasinov, line 116: Should it be an error?

vasinov (changes requested):

vasinov (commented):
Just realized that `query()` takes a vector instead of a string. Should be a string, so it's consistent with other drivers and the abstract method in `BaseVectorStoreDriver`.

griptape/drivers/vector/redis_vector_store_driver.py
trish11953: It shouldn't be an error since it's okay if the index exists right?

vasinov: Yeah, may be not an error then but a `warning`, since the method technically failed to create a new index.

Also, could you add ` -> None` to the signature, since we are not returning anything here?

trish11953: Done!

vasinov (approved):

PR #183, opened 2023-08-26
Add `SummaryTask` and `BaseInputTask`

vasinov:


griptape/tasks/base_text_input_task.py
collindutter, line 1: @vasinov we should have unit tests for this too.

Issue #148, opened 2023-07-28
If the caller doesn't have access to a model, it should not retry the call on InvalidRequestError.


jgreenemi:
**Describe the bug**
When the agent is run against a model that the user does not have access to, and encounters a InvalidRequestError, it retries the call.

**To Reproduce**
Running the following in a new OpenAI API account (one that has not passed the payment threshold to gain access to the GPT-4 model) will produce an InvalidRequestError:
```
agent = Agent(
            prompt_driver=OpenAiPromptDriver(
                model='gpt-4'
            ),
            tools=[WebScraper(),]
        )

agent.run('tell me if I have access to GPT-4')
```

**Expected behavior**
InvalidRequestError should be bubbled up without retrying.

**Screenshots**
```
WARNING:root:<RetryCallState 5417445168: attempt #1; slept for 0.0; last result: failed (InvalidRequestError The model `gpt-4` does not exist or you do not have access to it. Learn more: [https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.)>](https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.)%3E)
WARNING:root:<RetryCallState 5417445168: attempt #2; slept for 2.0; last result: failed (InvalidRequestError The model `gpt-4` does not exist or you do not have access to it. Learn more: [https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.)>](https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.)%3E)
WARNING:root:<RetryCallState 5417445168: attempt #3; slept for 4.0; last result: failed (InvalidRequestError The model `gpt-4` does not exist or you do not have access to it. Learn more: [https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.)>](https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.)%3E)
WARNING:root:<RetryCallState 5417445168: attempt #4; slept for 8.0; last result: failed (InvalidRequestError The model `gpt-4` does not exist or you do not have access to it. Learn more: [https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.)>](https://help.openai.com/en/articles/7102672-how-can-i-access-gpt-4.)%3E)
```

**Desktop (please complete the following information):**
 - OS: macOS
 - Version Ventura 13.4

PR #187, opened 2023-08-28
Don't retry invalid request to openai

collindutter:
Draft PR for #148


vasinov (commented):
We could add `ignored_exception_types: List[Type[Exception]]` to `ExponentialBackoffMixin` and set that in each subclass instead.

collindutter:
I like that more, will update tomorrow.

griptape/drivers/prompt/openai_completion_prompt_driver.py
```diff
@@ -33,6 +34,7 @@ class OpenAiCompletionPromptDriver(BasePromptDriver):
         kw_only=True
     )
     user: str = field(default="", kw_only=True)
+    ignored_exception_types: Tuple[Type[Exception], ...] = field(default=(openai.InvalidRequestError), kw_only=True)
```
vasinov, line 37: This default needs to be wrapped in a `Factory`.

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -33,6 +34,7 @@ class OpenAiChatPromptDriver(BasePromptDriver):
         kw_only=True
     )
     user: str = field(default="", kw_only=True)
+    ignored_exception_types: Tuple[Type[Exception], ...] = field(default=(openai.InvalidRequestError), kw_only=True)
```
vasinov, line 37: This default needs to be wrapped in a `Factory`.

griptape/mixins/exponential_backoff_mixin.py
```diff
@@ -13,13 +14,15 @@ class ExponentialBackoffMixin(ABC):
         default=lambda s: logging.warning(s),
         kw_only=True
     )
+    ignored_exception_types: Tuple[Type[Exception], ...] = field(default=(), kw_only=True)
```
vasinov, line 17: Should be `factory=tuple`.

vasinov (changes requested):

PR #189, opened 2023-08-29
Docs Integration Tests

collindutter:


griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -52,7 +53,7 @@ def load_entry(
             self, vector_id: str, namespace: Optional[str] = None
     ) -> Optional[BaseVectorStoreDriver.Entry]:
         collection = self.get_collection()
-        doc = collection.find_one({"_id": vector_id})
+        doc = collection.find_one({"_id": ObjectId(vector_id)})
```
collindutter, line 56: @trish11953 I was getting a MongoDb error without this. I imagine there are other areas where we need to wrap this too, but they weren't covered in the examples.

collindutter: Nvm I reverted the change and it still worked fine!

collindutter:
Closing in favor of https://github.com/griptape-ai/griptape/pull/189

PR #190, opened 2023-08-29
Add missing start/finish prompt events

collindutter:
Fixes regression from prompt stack update.


collindutter:
@vasinov is there a reason the `StartPromptEvent` and `FinishPromptEvent` events were removed?

vasinov (approved):

PR #191, opened 2023-08-30
Dev tools

collindutter:


collindutter:
@vasinov we should probably run these tools once on the entire project, especially before they become part of the automated checks. Do you want that to be part of this PR?

vasinov:
@collindutter, could you please resolve the poetry.lock conflict, so I can merge?

PR #196, opened 2023-08-31
Create shopify tool

collindutter:


griptape/tools/shopify_client/tool.py
```diff
@@ -0,0 +1,66 @@
...
+from griptape.tools import BaseTool
+from griptape.utils.decorators import activity
+from griptape.artifacts import BaseArtifact, TextArtifact, ErrorArtifact
+
+
+@define
+class ShopifyClient(BaseTool):
+    storename: str = field(kw_only=True)
+    storefront_access_token: str = field(kw_only=True)
+    shopify_endpoint: str = field(default="myshopify.com/api/2023-07/graphql.json")
```
vasinov, line 16: It says "not found" when I hit the endpoint in the browser.

vasinov: Missing `kw_only=True`.

vasinov: Should we call it something like `schema_endpoint` instead?

griptape/tools/shopify_client/tool.py
```diff
@@ -0,0 +1,66 @@
...
+from attr import define, field
+from griptape.tools import BaseTool
+from griptape.utils.decorators import activity
+from griptape.artifacts import BaseArtifact, TextArtifact, ErrorArtifact
+
+
+@define
+class ShopifyClient(BaseTool):
+    storename: str = field(kw_only=True)
+    storefront_access_token: str = field(kw_only=True)
```
vasinov, line 15: Are there more different types of access tokens that they issue? If not, then this can just be `access_token`.

griptape/tools/shopify_client/tool.py
```diff
@@ -0,0 +1,66 @@
...
+    shopify_endpoint: str = field(default="myshopify.com/api/2023-07/graphql.json")
+
+    @activity(
+        config={
+            "description": "Can be used to search for shopify products",
```
vasinov, line 20: Capitalize "shopify." Might matter for the LLM.

griptape/tools/shopify_client/tool.py
```diff
@@ -0,0 +1,66 @@
...
+            "schema": Schema(
+                {
+                    Literal("first", description="How many products to retreive."): str,
```
vasinov, line 23: Let's rename it to something more descriptive for the LLM. May be `product_count`?

vasinov: *retrieve

griptape/tools/shopify_client/tool.py
```diff
@@ -0,0 +1,66 @@
...
+                }
+            ),
+        }
+    )
+    def search(self, params: dict) -> BaseArtifact:
+        from requests import post, exceptions
+
+        first = params.get("first", 10)
```
vasinov, line 31: I think we can set the default in the schema. But also, we should move this into one of the class properties, so the user can specify this number.

griptape/tools/shopify_client/tool.py
```diff
@@ -0,0 +1,66 @@
...
+
+        url = f"https://{self.storename}.{self.shopify_endpoint}"
```
vasinov, line 33: Can we use something like `RestApiClient._build_url`, so it doesn't break because of extra/missing slashes? May be abstract it into a `util`?

griptape/tools/shopify_client/tool.py
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        try:
+            response = post(
+                url,
+                json=body,
+                headers={
+                    "X-Shopify-Storefront-Access-Token": self.storefront_access_token
+                },
+                timeout=30,
```
vasinov, line 61: Let's move the timeout into a class property as well.

griptape/tools/shopify_client/tool.py
```diff
@@ -0,0 +1,66 @@
...
+            )
+
+            return TextArtifact(response.text)
```
vasinov, line 64: The response is a list, right? We should return it as `list[TextArtifact]` here.

vasinov (changes requested):

griptape/tools/shopify_client/tool.py
collindutter: @vasinov do you have an example of specifying a default value in a tool schema? I can do it via:
```
Optional(
  Literal(
      "product_count",
      description="How many products to retrieve.",
  ),
  default=10,
): str,
```
But the default value is not passed into the activity when the LLM calls it.

Regarding the class property, isn't this field kind of specific to this single API call?

griptape/tools/shopify_client/tool.py
collindutter: I really need to review my I's and E's!

griptape/tools/shopify_client/tool.py
collindutter: Does `urljoin` help here? We don't really have a base url until we join `storename` + `schema_endpoint`.

griptape/tools/shopify_client/tool.py
collindutter: @vasinov it's just a component of the full url. From the [Shopify curl docs](https://shopify.dev/docs/api/storefront#authentication):
`https://{store}.myshopify.com/api/2021-07/graphql.json`

PR #206, opened 2023-09-07
Don't create index on initialization

collindutter:


collindutter:
@vasinov @trish11953 let me know what you think about this one. I ran into some unexpected behavior in my testing because the `MarqoVectorStoreDriver` creates an index on initialization. As far as I can tell, none of the other drivers do this.

vasinov (approved):

PR #207, opened 2023-09-08
Add support for saving content into files from the prompt to FileManager

vasinov:


griptape/tools/file_manager/tool.py
```diff
@@ -45,39 +45,66 @@ def load_files_from_disk(self, params: dict) -> ListArtifact | ErrorArtifact:
...
 
     @activity(config={
-        "description": "Can be used to save an artifact namespace to disk",
+        "description": "Can be used to save data to disk",
         "schema": Schema({
-            "memory_name": str,
-            "artifact_namespace": str,
             Literal(
                 "path",
                 description="Destination path on disk in the POSIX format. For example, ['foo/bar/file.txt']"
```
collindutter, line 52: While we're in here making changes...

The description should probably be `Destination path on disk in the POSIX format. For example, 'foo/bar/file.txt'` since it is only a single output path.

PR #209, opened 2023-09-11
Add loaders and optional save file encoding setting to FileManager

collindutter:
Adds an optional `encoding` field to `FileManager` so that basic file content can be loaded into Tool Memory.


griptape/tools/file_manager/tool.py
```diff
@@ -10,6 +10,7 @@
 @define
 class FileManager(BaseTool):
     dir: str = field(default=os.getcwd(), kw_only=True)
+    encoding: str = field(default=None, kw_only=True)
```
vasinov, line 13: `Optional[str]`

griptape/tools/file_manager/tool.py
```diff
@@ -30,13 +31,18 @@ def load_files_from_disk(self, params: dict) -> ListArtifact | ErrorArtifact:
 
             try:
                 with open(full_path, "rb") as file:
-                    list_artifact.value.append(
-                        BlobArtifact(
+                    if self.encoding is not None:
```
vasinov, line 34: Can just be `if self.encoding:`

vasinov: We should also make it more descriptive. Something like `load_file_encoding`?

Wondering if we should integrate it into `_save_to_disk` as well.... Then the question becomes if we use the same `encoding` field for both or if we want to have `load_file_encoding` and `save_file_encoding`.

vasinov (changes requested):

vasinov (approved):
I resolved merge conflicts and also added the following changes:

- Removed `FileLoader.workdir`
- Added `FileLoader.encoding`
- Added `FileManager.default_loader`
- Added `FileManager.loaders`
- Removed `FileManager.save_file_encoding`

PR #211, opened 2023-09-11
Support parsing subtask output with newlines

collindutter:
Adds support for `ActionSubtask`s to parse output that spans multiple lines i.e. writing paragraphs of text.


griptape/tasks/action_subtask.py
```diff
@@ -162,8 +162,7 @@ def __init_from_prompt(self, value: str) -> None:
         if len(action_matches) > 0:
             try:
                 data = action_matches[-1]
-                data = data.replace('\n', '')
```
vasinov, line 165: If I remember correctly this was a temporary fix to address LLM outputs like 

```
Action:
{
"type": "tool",
...
}
```

Without this fix, will it still work? And if it doesn't, the fix should ideally be moved into the regex extraction expression.

vasinov (changes requested):

griptape/tasks/action_subtask.py
collindutter: @vasinov I added a test that shows that this use-case does work. Though maybe we should still resolve via the regex instead of `strict=False`?

Issue #168, opened 2023-07-24
Relative date support for the `DateTime` tool


vasinov:
LLMs often need to know dates and times for "yesterday," "tomorrow," "3 hours ago," etc.

I propose adding an activity to the `DateTime` tool (based on [dateparser](https://dateparser.readthedocs.io/en/latest/)) that does exactly that.

collindutter:
https://github.com/griptape-ai/griptape/pull/216

PR #216, opened 2023-09-12
Add relative datetime activity

collindutter:
#168 Adds relative date activity, enabling use-cases like this:
```python
from griptape.tools import DateTime

agent = Agent(tools=[DateTime()])
agent.run("What time was it 3 months, 1 week and 1 day ago") # relative functionality
agent.run("What is the current time?") # existing functionality
```

### DIFF
--- a/griptape/tools/date_time/tool.py
+++ b/griptape/tools/date_time/tool.py
@@ -24,8 +24,8 @@ class DateTime(BaseTool):
         "description": "Can be used to return a relative date and time.",
         "schema": Schema({
             Literal(
-                description="A string representing date and/or time in a recognizably valid format."
                 "relative_date_string",
+                description='Relative date in English. For example, "now EST", "20 minutes ago", or "yesterday at 2pm"'
             ): str
         })
     })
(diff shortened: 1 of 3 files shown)


griptape/tools/date_time/tool.py
```diff
@@ -17,3 +18,22 @@ def get_current_datetime(self, _: dict) -> BaseArtifact:
             return TextArtifact(str(current_datetime))
         except Exception as e:
             return ErrorArtifact(f"error getting current datetime: {e}")
+    
+    @activity(config={
+        "uses_default_memory": False,
+        "description": "Can be used to return a relative date and time.",
+        "schema": Schema({
+            Literal(
+                "date_string",
```
vasinov, line 27: Let's make it less ambiguous and call it `relative_date_string`.

griptape/tools/date_time/tool.py
```diff
@@ -17,3 +18,22 @@ def get_current_datetime(self, _: dict) -> BaseArtifact:
...
+                description="A string representing date and/or time in a recognizably valid format."
```
vasinov, line 28: I think this might be too ambiguous for the LLM as well. May be it should be something like:
> Relative date in English. For example, "now EST", "20 minutes ago", or "yesterday at 2pm"

vasinov (changes requested):
Could you add a test for this activity as well?

vasinov (approved):

Issue #171, opened 2023-06-08
All tools should use Griptape loaders when approriate


vasinov:
- [ ] `WebScraper` should use `WebLoader`

PR #217, opened 2023-09-12
Use WebLoader in WebScraper tool

collindutter:
#171

### DIFF
--- a/griptape/tools/web_scraper/tool.py
+++ b/griptape/tools/web_scraper/tool.py
@@ -1,12 +1,11 @@
 from __future__ import annotations
-import logging
-import json
 from attr import define, field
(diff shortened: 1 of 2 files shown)


griptape/tools/web_scraper/tool.py
```diff
@@ -24,7 +23,7 @@ class WebScraper(BaseTool):
     })
     def get_content(self, params: dict) -> ListArtifact | ErrorArtifact:
         url = params["values"]["url"]
-        page = self._load_page(url)
+        page = WebLoader()._load_page(url, self.include_links)
```
collindutter, line 26: @vasinov Not sure if we should be using this protected method like this.

vasinov: Let's just make it public.

vasinov (changes requested):

vasinov (approved):

PR #218, opened 2023-09-12
Google Drive Tool

trish11953:


griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,171 @@
...
+        from googleapiclient.discovery import build
+
+        values = params["values"]
+
+        try:
+            credentials = service_account.Credentials.from_service_account_info(
+                self.service_account_credentials, scopes=self.LIST_FILES_SCOPES
+            )
+            delegated_credentials = credentials.with_subject(values["drive_owner_email"])
+            service = build('drive', 'v3', credentials=delegated_credentials)
```
collindutter, line 41: Let's move this to a separate utility function `_build_client(scopes)`.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,171 @@
...
+        from googleapiclient.http import MediaFileUpload
+
+        values = params['values']
+
+        try:
+            credentials = service_account.Credentials.from_service_account_info(
+                self.service_account_credentials, scopes=self.UPLOAD_FILE_SCOPES
+            )
+            delegated_credentials = credentials.with_subject(values["drive_owner_email"])
+            service = build('drive', 'v3', credentials=delegated_credentials)
```
collindutter, line 84: Move to `_build_scopes`

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,171 @@
...
+        from googleapiclient.discovery import build
+
+        values = params["values"]
+
+        try:
+            credentials = service_account.Credentials.from_service_account_info(
+                self.service_account_credentials, scopes=self.LIST_FILES_SCOPES
+            )
+            delegated_credentials = credentials.with_subject(values["drive_owner_email"])
+            service = build('drive', 'v3', credentials=delegated_credentials)
```
collindutter, line 120: Move to `_build_scopes`

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,171 @@
...
+                "drive_owner_email",
+                description="email of the Google Drive's owner"
+            ): str,
+            Optional(Literal(
+                "max_files",
+                description="maximum number of files to return"
+            )): int
+        })
+    })
+    def list_files(self, params: dict) -> list[TextArtifact] | ErrorArtifact:
```
collindutter, line 30: Use `ListArtifact` instead of `list[TextArtifact]`

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,171 @@
...
+        try:
+            credentials = service_account.Credentials.from_service_account_info(
+                self.service_account_credentials, scopes=self.LIST_FILES_SCOPES
+            )
+            delegated_credentials = credentials.with_subject(values["drive_owner_email"])
+            service = build('drive', 'v3', credentials=delegated_credentials)
+
+            results = service.files().list(pageSize=values.get('max_files', 10)).execute()
+            items = results.get('files', [])
+            return [TextArtifact(str(i)) for i in items]
```
collindutter, line 45: Return `ListArtifact`

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,171 @@
...
+                "drive_owner_email",
+                description="email of the Google Drive's owner"
+            ): str,
+            Literal(
+                "file_name",
+                description="name of the file to search for in Google Drive"
+            ): str
+        })
+    })
+    def search_file(self, params: dict) -> list[TextArtifact] | ErrorArtifact:
```
collindutter, line 147: Use `ListArtifact` instead of `list[TextArtifact]`

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,171 @@
...
+                self.service_account_credentials, scopes=self.LIST_FILES_SCOPES
+            )
+            delegated_credentials = credentials.with_subject(values["drive_owner_email"])
+            service = build('drive', 'v3', credentials=delegated_credentials)
+
+            query = f"name='{values['file_name']}'"
+            results = service.files().list(q=query).execute()
+            items = results.get('files', [])
+
+            return [TextArtifact(str(i)) for i in items]
```
collindutter, line 164: Return `ListArtifact`

collindutter (changes requested):

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def list_files(self, params: dict) -> ListArtifact | ErrorArtifact:
+
+        values = params["values"]
+
+        try:
+            service = self._build_client(self.LIST_FILES_SCOPES, values)
+            results = service.files().list(pageSize=values.get('max_files', 10)).execute()
+            items = results.get('files', [])
+            return items
```
vasinov, line 48: This should return a `ListArtifact`, not a `list`.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+
+        except Exception as e:
+            logging.error(e)
+            return ErrorArtifact(f"error retrieving files from Drive: {e}")
```
vasinov, line 52: *from Google Drive

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @activity(config={
+        "description": "Can be used to upload a file to Google Drive",
+        "schema": Schema({
+            Literal(
+                "drive_owner_email",
```
vasinov, line 58: Same as `list_files`. Do we really need it here?

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+                "file_name",
+                description="name to give the file on Google Drive"
+            ): str,
+            Literal(
+                "mime_type",
+                description="MIME type of the file being uploaded"
+            ): str
+        })
+    })
+    def upload_file(self, params: dict) -> InfoArtifact | ErrorArtifact:
```
vasinov, line 75: Take a look at how we do this in `FileManager.save_memory_artifacts_to_disk` and `FileManager.save_content_to_file`. I think we want the same pattern repeated here.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+            Literal(
+                "file_path",
+                description="path to the file to be uploaded"
+            ): str,
+            Literal(
+                "file_name",
+                description="name to give the file on Google Drive"
+            ): str,
+            Literal(
+                "mime_type",
```
vasinov, line 70: Is this a requirement to have this field defined? Ideally, the LLM shouldn't know about this.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+
+        except Exception as e:
+            logging.error(e)
+            return ErrorArtifact(f"error uploading file to Drive: {e}")
+
+    @activity(config={
+        "description": "Can be used to download a file from Google Drive",
+        "schema": Schema({
+            Literal(
+                "drive_owner_email",
```
vasinov, line 96: Same as list_files. Do we really need it here?

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+                description="email of the Google Drive's owner"
+            ): str,
+            Literal(
+                "file_id",
```
vasinov, line 100: We should support direct file downloading like we do in `FileManager.load_files_from_disk`. There are two options here:

1. Make this an `schema.Or` field and let the LLM download the file by `file_id` or from `paths`.
2. Add a separate activity that closely matches `FileManager.load_files_from_disk`.

I think option 2 makes more sense as it's less ambiguous for the LLM but let me know what you think.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+    def download_file(self, params: dict) -> TextArtifact | ErrorArtifact:
+
+        values = params["values"]
+
+        try:
+            service = self._build_client(self.LIST_FILES_SCOPES, values)
+            request = service.files().get_media(fileId=values["file_id"])
+            downloaded_file = request.execute()
+
+            return TextArtifact(downloaded_file)
```
vasinov, line 114: Should be `BlobArtifact`.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+
+        except HttpError as e:
+            logging.error(e)
+            return ErrorArtifact(f"error downloading file from Drive: {e}")
+        except MalformedError:
+            logging.error("MalformedError occurred")
+            return ErrorArtifact("error downloading file due to malformed credentials")
+
+    @activity(config={
+        "description": "Can be used to search for a file in Google Drive",
```
vasinov, line 124: Same as list_files. Do we really need it here?

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+        })
+    })
+    def search_file(self, params: dict) -> ListArtifact | ErrorArtifact:
+
+        values = params["values"]
+
+        try:
+            service = self._build_client(self.LIST_FILES_SCOPES, values)
+            query = f"name='{values['file_name']}'"
+            results = service.files().list(q=query).execute()
```
vasinov, line 143: What kind of search does this do? Based on file name? Content? Metadata? All of the above? Depending on the answer we should re-structure the schema.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+            items = results.get('files', [])
+
+            return items
```
vasinov, line 146: Should be `ListArtifact`.

vasinov (changes requested):

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @activity(config={
+        "description": "Can be used to list files from Google Drive",
+        "schema": Schema({
+            Literal(
+                "drive_owner_email",
+                description="email of the Google Drive's owner"
+            ): str,
+            Optional(Literal(
+                "max_files",
```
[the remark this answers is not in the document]
trish11953: done

griptape/tools/google_drive/tool.py
trish11953: based on file name

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,153 @@
...
+@define
+class GoogleDriveClient(BaseGoogleClient):
+    LIST_FILES_SCOPES = ['https://www.googleapis.com/auth/drive.readonly']
+    UPLOAD_FILE_SCOPES = ['https://www.googleapis.com/auth/drive.file']
+
+    @activity(config={
+        "description": "Can be used to list files from Google Drive",
+        "schema": Schema({
+            Literal(
+                "drive_owner_email",
```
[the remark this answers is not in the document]
trish11953: I was essentially keeping it as a validation layer to ensure that the function receives the right kind of data. I cant remove it if you want

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+        from google.oauth2 import service_account
+        from googleapiclient.discovery import build
+        credentials = service_account.Credentials.from_service_account_info(
+            self.service_account_credentials, scopes=scopes
+        )
+        delegated_credentials = credentials.with_subject(self.owner_email)
+        service = build('drive', 'v3', credentials=delegated_credentials)
+        return service
+
+    def list_files(self, params: dict) -> ListArtifact | ErrorArtifact:
```
vasinov, line 29: What happened to `@activities` for all of the methods?

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+        values = params["values"]
+
+        try:
+            service = self._build_client(self.LIST_FILES_SCOPES)
+
+            folder_id = None
+            if "folder_path" in values and values["folder_path"]:
```
vasinov, line 36: You don't have to check for the key since we do that during LLM input validation.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+
+            results = service.files().list(q=query, pageSize=max_files).execute()
+            items = results.get('files', [])
+            return ListArtifact(items)
+
+        except Exception as e:
+            logging.error(e)
+            return ErrorArtifact(f"error retrieving files from Drive: {e}")
+
+    def _resolve_path_to_id(self, service, path: str) -> Optional[str]:
```
vasinov, line 55: Let's rename it something more descriptive like `_path_to_file_id`.

vasinov: On the second pass...I'm not actually sure what the method does. Can you explain it and write a comment?

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+            if index == len(parts) - 1:  # If it's the last part of the path
+                query = f"name='{part}' and '{current_id}' in parents"
+            else:
+                query = f"name='{part}' and '{current_id}' in parents and mimeType='application/vnd.google-apps.folder'"
+
+            response = service.files().list(q=query).execute()
+            files = response.get('files', [])
+
+            if not files:
+                return None
```
vasinov, line 69: This will exit the method on any `part` that doesn't return files. Is that what you really want?

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def upload_file(self, params: dict) -> InfoArtifact | ErrorArtifact:
+        from googleapiclient.http import MediaFileUpload
+
+        values = params['values']
+
+        try:
+            service = self._build_client(self.UPLOAD_FILE_SCOPES)
+            file_metadata = {'name': values['file_name']}
+            media = MediaFileUpload(values['file_path'], mimetype=values['mime_type'])
```
vasinov, line 82: What is `file_path`?

griptape/tools/google_drive/tool.py
vasinov: Doesn't look like this comment was addressed.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+        try:
+            service = self._build_client(self.LIST_FILES_SCOPES)
+            file_id = self._resolve_path_to_id(service, values["file_path"])
+            if not file_id:
+                return ErrorArtifact(f"Could not find file: {values['file_path']}")
+
+            file_info = service.files().get(fileId=file_id).execute()
+            mime_type = file_info['mimeType']
+
+            google_export_mime_mapping = {
```
vasinov, line 103: This should be a const at the top of the class.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+                    "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+                "application/vnd.google-apps.presentation":
+                    "application/vnd.openxmlformats-officedocument.presentationml.presentation"
+            }
+
+            if mime_type in google_export_mime_mapping:
+                export_mime = google_export_mime_mapping[mime_type]
+                request = service.files().export_media(fileId=file_id, mimeType=export_mime)
+            else:
+                request = service.files().get_media(fileId=file_id)
```
vasinov, line 116: What's the difference between those two calls? Do we care about download vs export in the context of the LLM downloading it?

griptape/tools/google_drive/tool.py
vasinov: Looks like it [supports](https://developers.google.com/drive/api/guides/search-files) multiple modes. I think it would be very useful to provide search based on the content of the file as well. Imagine asking the LLM "find all files in folder/foo/bar that contain 'vasily' in body."

We can move that to one of the activity schema params, so that the LLM can decide what type of search to perform.

vasinov (changes requested):

griptape/tools/google_drive/tool.py
trish11953: This method is translating the user-friendly file or folder path into the corresponding file ID used by Google Drive to identify files and folders. It recursively searches for each part of the path within the context of the previous part's search results. If any part of the path is not found, it returns None to indicate that the path is invalid or the file/folder does not exist.

griptape/tools/google_drive/tool.py
trish11953: I just removed them as of now, wanted to check with you on how to write them correctly.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def _build_client(self, scopes: list[str]) -> Resource:
+        from google.oauth2 import service_account
+        from googleapiclient.discovery import build
+        credentials = service_account.Credentials.from_service_account_info(
+            self.service_account_credentials, scopes=scopes
+        )
+        delegated_credentials = credentials.with_subject(self.owner_email)
+        service = build('drive', 'v3', credentials=delegated_credentials)
+        return service
```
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+from griptape.tools import BaseGoogleClient
+from googleapiclient.errors import HttpError
+from google.auth.exceptions import MalformedError
+from googleapiclient.discovery import Resource
+
+@define
+class GoogleDriveClient(BaseGoogleClient):
+    owner_email: str = field(kw_only=True)
+    LIST_FILES_SCOPES = ['https://www.googleapis.com/auth/drive.readonly']
+    UPLOAD_FILE_SCOPES = ['https://www.googleapis.com/auth/drive.file']
```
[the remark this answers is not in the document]
trish11953: Like before the class ?

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,145 @@
...
+    max_files: int = field(default=10, kw_only=True)
```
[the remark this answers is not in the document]
trish11953: I have a question for you. So, what will I do if the number of retrieved files is greater than the maximum number of files specified by the user? (I have included the pagination code to handle the case where the retrieved files are greater than the maximum number of files specified by the user)
Also, the pagelimit specified by the user can exceed the maximum pagelimit taken by the Drive API.

griptape/tools/google_drive/tool.py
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: So the if statement exports the Google docs as docx, Google sheets as xlsx and Google Slides as pptx. But the other type of files can just be downloaded as their own format.

griptape/tools/google_drive/tool.py
trish11953: file_path is the local path of the file you want to upload.

griptape/tools/google_drive/tool.py
trish11953: Yes, I think this is desired behavior for this method given its current use case. if the function returns None, I have put an error message for whichever function that's using the _path_to_file_id.

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+        return current_id
+
+    @activity(config={
+        "description": "List files in a specific Google Drive folder or the root if no folder is specified",
+        "schema": Schema({
+            Optional("folder_path",
+                     description="Path of the Google Drive folder (like 'MainFolder/Subfolder1/Subfolder2') "
+                                 "from which files should be listed. If not provided, files from the root "
+                                 "directory will be listed."): str,
+            Optional("max_files",
```
vasinov, line 70: So, answering your previous question, I think this shouldn't be up to the LLM. The goal of the activity is to show all files in the directory, not a subset of files.

We should return all files in the directory, so it's unambiguous. That means that the method that pulls files needs to have pagination logic built-in. In that case, we don't even need to expose `page_size` as a property because it will be just an implementation details specific to your implementation.

griptape/tools/google_drive/tool.py
vasinov: If this is not `Optional` anymore, you don't have to do the check.

As a side note: if it were optional then a much better way to check for `None` would be `values.get("folder_path")` which doesn't throw an exception on `None` unlike `values["folder_path"]`.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+    def list_files(self, params: dict) -> ListArtifact | ErrorArtifact:
+        values = params["values"]
+
+        try:
+            service = self._build_client(self.LIST_FILES_SCOPES)
+
+            folder_id = None
+            if "folder_path" in values and values["folder_path"]:
+                folder_id = self._path_to_file_id(service, values["folder_path"])
+                if not folder_id:
```
vasinov, line 83: Should be:

```python
if folder_id:
    ... # the rest of your code
else:
    return ErrorArtifact(f"Could not find folder: {values['folder_path']}")
```

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+                    return ErrorArtifact(f"Could not find folder: {values['folder_path']}")
+
+            if folder_id:
```
vasinov, line 86: Don't need this check since you already return in the case when `folder_id` is `None`.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+
+                files = results.get('files', [])
+                items.extend(files)
+
+                next_page_token = results.get('nextPageToken')
+
+                if not next_page_token:
+                    break
+
+            return ListArtifact(items)
```
vasinov, line 112: `items` here should be a list of `TextArtifact`s. Take a loot at the `ListArtifact.value` signature.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+        "schema": Schema({
+            Optional("folder_path",
+                     description="Path of the Google Drive folder (like 'MainFolder/Subfolder1/Subfolder2') "
+                                 "from which files should be listed. If not provided, files from the root "
+                                 "directory will be listed."): str,
+            Optional("max_files",
+                     description="Maximum number of files to list. If not provided, the default page size will be used."): int
+        })
+    })
+    def list_files(self, params: dict) -> ListArtifact | ErrorArtifact:
```
vasinov, line 74: This method can be simplified by breaking away the listing logic into a protected method.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+            logging.error(e)
+            return ErrorArtifact(f"error retrieving files from Google Drive: {e}")
+
+    @activity(config={
+        "description": "Can be used to save memory artifacts to Google Drive using folder paths",
+        "schema": Schema({
+            "memory_name": str,
+            "artifact_namespace": str,
+            "file_name": str,
+            Optional("folder_path",
```
vasinov, line 124: Let's make this required, so we avoid awkward logic in the implementation and also make sure the LLM is more intentional in its actions.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+            if not files:
+                return None
+            current_id = files[0]['id']
+
+        return current_id
+
+    @activity(config={
+        "description": "List files in a specific Google Drive folder or the root if no folder is specified",
+        "schema": Schema({
+            Optional("folder_path",
```
vasinov, line 66: Let's make this required, so we avoid awkward logic in the implementation and also make sure the LLM is more intentional in its actions.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+        if memory:
+            artifacts = memory.load_artifacts(values["artifact_namespace"])
+    
+            if not artifacts:
+                return ErrorArtifact("no artifacts found")
+
+            service = self._build_client(self.UPLOAD_FILE_SCOPES)
+
+            folder_id = None
+            if "folder_path" in values and values["folder_path"]:
```
vasinov, line 144: Remove this check after making `folder_path` required.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+        "description": "Can be used to save content to a file",
+        "schema": Schema(
+            {
+                Literal(
+                    "path",
+                    description="Destination file path on disk in the POSIX format. For example, 'foo/bar/baz.txt'"
+                ): str,
+                "content": str
+            }
+        )
```
vasinov, line 173: Doesn't look like you modified those after copying them from `FileManager`.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+        service = self._build_client(self.UPLOAD_FILE_SCOPES)
+    
+        if isinstance(value, str):
+            value = value.encode()
+    
+        file_metadata = {"name": filename}
+        if parent_folder_id:
+            file_metadata["parents"] = [parent_folder_id]
+    
+        media = MediaIoBaseUpload(BytesIO(value), mimetype="application/octet-stream", resumable=True)
```
vasinov, line 196: Nice.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+    
+        try:
+            file = service.files().create(body=file_metadata, media_body=media, fields="id").execute()
+            return InfoArtifact(file.get("id"))
+        except HttpError as error:
+            return ErrorArtifact(f"An error occurred: {error}")
```
vasinov, line 202: Don't catch the exception here. Let the consumers do it.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+        filename = params["values"]["path"]
+
+        try:
+            self._save_to_drive(filename, content)
+
+            return InfoArtifact(f"saved successfully")
+        except Exception as e:
+            return ErrorArtifact(f"error saving file to Google Drive: {e}")
+
+    def _save_to_drive(self, filename: str, value: any, parent_folder_id=None) -> InfoArtifact | ErrorArtifact:
```
vasinov, line 186: Move method to the bottom of the class since it's a protected method.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+
+        except HttpError as e:
+            logging.error(e)
+            return ErrorArtifact(f"error downloading file from Google Drive: {e}")
+        except MalformedError:
+            logging.error("MalformedError occurred")
+            return ErrorArtifact("error downloading file due to malformed credentials")
+
+    @activity(config={
+        "description": "Search for a file on Google Drive based on name or content",
```
vasinov, line 242: Should read as "Can be used to search for files..."

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+        "schema": Schema({
+            Literal("search_mode",
+                    description="Mode of search, either 'name' to search by file name or "
+                                "'content' to search by file content."): str,
+            Literal("file_name",
+                    description="Name of the file to search for, used when search_mode is 'name'"): str,
+            Optional("search_text",
+                     description="Text to search for within files, used when search_mode is 'content'"): str,
```
vasinov, line 250: Combine into one field called `search_query` and interpret based on the value of `search_mode`.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+            logging.error(e)
+            return ErrorArtifact(f"error downloading file from Google Drive: {e}")
+        except MalformedError:
+            logging.error("MalformedError occurred")
+            return ErrorArtifact("error downloading file due to malformed credentials")
+
+    @activity(config={
+        "description": "Search for a file on Google Drive based on name or content",
+        "schema": Schema({
+            Literal("search_mode",
```
vasinov, line 244: Should be an `Or` property:

```python
Literal(
    "search_mode",
    description="File search mode. Use 'name' to search in file name or 'content' to search in file content"
): schema.Or("name", "content")
```

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+                    description="Mode of search, either 'name' to search by file name or "
+                                "'content' to search by file content."): str,
+            Literal("file_name",
+                    description="Name of the file to search for, used when search_mode is 'name'"): str,
+            Optional("search_text",
```
vasinov, line 249: Should be required.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+                     description="Text to search for within files, used when search_mode is 'content'"): str,
+            Optional("folder_path",
+                     description="Path of the Google Drive folder (like 'MainFolder/Subfolder1') "
+                                 "in which to restrict the search. If not provided, the search "
+                                 "will span the entire Drive."): str
+        })
+    })
+    def search_file(self, params: dict) -> ListArtifact | ErrorArtifact:
```
vasinov, line 257: Should be called `search_files`.

vasinov (changes requested):

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+        from google.oauth2 import service_account
+        from googleapiclient.discovery import build
+
+        credentials = service_account.Credentials.from_service_account_info(
+            self.service_account_credentials, scopes=scopes
+        )
+        delegated_credentials = credentials.with_subject(self.owner_email)
+        service = build('drive', 'v3', credentials=delegated_credentials)
+
+        return service
```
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+        "application/vnd.google-apps.spreadsheet":
+            "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+        "application/vnd.google-apps.presentation":
+            "application/vnd.openxmlformats-officedocument.presentationml.presentation"
+    }
+
+    owner_email: str = field(kw_only=True)
+    page_size: int = field(default=100, kw_only=True)
+
+    def _build_client(self, scopes: list[str]) -> Resource:
```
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Sure!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+    
+        file_metadata = {"name": filename}
+        if parent_folder_id:
+            file_metadata["parents"] = [parent_folder_id]
+    
+        media = MediaIoBaseUpload(BytesIO(value), mimetype="application/octet-stream", resumable=True)
+    
+        try:
+            file = service.files().create(body=file_metadata, media_body=media, fields="id").execute()
+            return InfoArtifact(file.get("id"))
```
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,293 @@
...
+
+    @activity(config={
+        "description": "Download a file from Google Drive based on a provided path",
+        "schema": Schema({
+            Literal("file_path",
+                    description="Path of the Google Drive file (like 'MainFolder/Subfolder1/filename.ext') "
+                                "that needs to be downloaded."): str
+        })
+    })
+    def download_file(self, params: dict) -> BlobArtifact | ErrorArtifact:
```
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,325 @@
...
+            return ErrorArtifact(f"error searching for file in Google Drive: {e}")
+        except MalformedError:
+            logging.error("MalformedError occurred")
+            return ErrorArtifact("error searching for file due to malformed credentials")
+
+    def _build_client(self, scopes: list[str]) -> Resource:
+        from google.oauth2 import service_account
+        from googleapiclient.discovery import build
+
+        credentials = service_account.Credentials.from_service_account_info(self.service_account_credentials,
```
vasinov, line 248: Formatting:

```python
credentials = service_account.Credentials.from_service_account_info(
    self.service_account_credentials, scopes=scopes
)
```

vasinov (changes requested):

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,325 @@
...
+                    "artifact_namespace": str,
+                    "file_name": str,
+                    Literal(
+                        "folder_path",
+                        description="Path of the Google Drive folder (like 'MainFolder/Subfolder1/Subfolder2') "
+                        "where the file should be saved. Specify 'root' to save in the "
+                        "root directory.",
+                    ): str,
+                }
+            ),
```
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,325 @@
...
+    )
+    def save_memory_artifacts_to_drive(self, params: dict) -> ErrorArtifact | InfoArtifact:
+        values = params["values"]
+        memory = self.find_input_memory(values["memory_name"])
+        file_name = values["file_name"]
+    
+        if memory:
+            artifacts = memory.load_artifacts(values["artifact_namespace"])
+    
+            if not artifacts:
```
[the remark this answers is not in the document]
trish11953: Ok done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,325 @@
...
+        "description": "Can be used to download multiple files from Google Drive based on a provided list of paths",
+        "schema": Schema({
+            Literal(
+                "paths",
+                description="List of paths to files to be loaded in the POSIX format. "
+                            "For example, ['foo/bar/file1.txt', 'foo/bar/file2.txt']"
+            ): [str]
+        })
+    })
+    def download_files(self, params: dict) -> list[BlobArtifact] | ErrorArtifact:
```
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,325 @@
...
+        values = params["values"]
+        downloaded_files = []
+
+        service = self._build_client(self.LIST_FILES_SCOPES)
+
+        for path in values["paths"]:
+            try:
+                file_id = self._path_to_file_id(service, path)
+                if not file_id:
```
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,325 @@
...
+        search_mode = values["search_mode"]
+    
+        try:
+            service = self._build_client(self.LIST_FILES_SCOPES)
+    
+            if values["folder_path"] == "root":
+                folder_id = "root"
+            else:
+                folder_id = self._path_to_file_id(service, values["folder_path"])
+                if not folder_id:
```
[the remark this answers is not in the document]
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,325 @@
...
+                    return ErrorArtifact(f"Folder path {values['folder_path']} not found")
+
+            if search_mode == "name":
+                query = f"name='{values['search_query']}'"
+            elif search_mode == "content":
+                query = f"fullText contains '{values['search_query']}'"
+            else:
+                return ErrorArtifact(f"Invalid search mode: {search_mode}")
+
+            query += " and trashed=false"
```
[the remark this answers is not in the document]
trish11953: Yes

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,325 @@
...
+                folder_id = self._path_to_file_id(service, values["folder_path"])
+                if not folder_id:
+                    return ErrorArtifact(f"Folder path {values['folder_path']} not found")
+
+            if search_mode == "name":
+                query = f"name='{values['search_query']}'"
+            elif search_mode == "content":
+                query = f"fullText contains '{values['search_query']}'"
+            else:
+                return ErrorArtifact(f"Invalid search mode: {search_mode}")
```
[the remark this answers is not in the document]
trish11953: You're right!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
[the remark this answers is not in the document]
vasinov: Don't see it :)

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,335 @@
...
+        service = self._build_client(self.UPLOAD_FILE_SCOPES)
+    
+        if isinstance(value, str):
+            value = value.encode()
+
+        parts = filename.split("/")
+        if len(parts) > 1:
+            directory = "/".join(parts[:-1])
+            parent_folder_id = self._path_to_file_id(service, directory)
+            if not parent_folder_id:
```
vasinov, line 301: No early returns.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,335 @@
...
+        file_metadata = {"name": filename}
+        if parent_folder_id:
+            file_metadata["parents"] = [parent_folder_id]
+    
+        media = MediaIoBaseUpload(BytesIO(value), mimetype="application/octet-stream", resumable=True)
+
+        file = service.files().create(body=file_metadata, media_body=media, fields="id").execute()
+        return InfoArtifact(file)
+
+    def _list_files(self, service, query: str) -> list:
```
vasinov, line 314: Can you specify a type for the return `list`?

vasinov (changes requested):

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,335 @@
...
+        "application/vnd.google-apps.presentation":
+            "application/vnd.openxmlformats-officedocument.presentationml.presentation"
+    }
+
+    owner_email: str = field(kw_only=True)
+
+    @activity(config={
+        "description": "Can be used to list files in a specific Google Drive folder or the root directory",
+        "schema": Schema({
+            Literal("folder_path",
```
vasinov, line 33: Can be optional too.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,335 @@
...
+                    description="Path of the Google Drive folder (like 'MainFolder/Subfolder1/Subfolder2') "
+                                "from which files should be listed. Specify 'root' to list files from the "
```
vasinov, line 35: Once it's made optional, you don't need to prompt the LLM to provide "root" as a value.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,335 @@
...
+            "description": "Can be used to save memory artifacts to Google Drive using folder paths",
+            "schema": Schema(
+                {
+                    "memory_name": str,
+                    "artifact_namespace": str,
+                    "file_name": str,
+                    Optional(
+                        "folder_path",
+                        description="Path of the Google Drive folder (like 'MainFolder/Subfolder1/Subfolder2') "
+                                    "where the file should be saved. Specify 'root' to save in the "
```
vasinov, line 72: You don't need to prompt the LLM to provide "root" as a value.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,335 @@
...
+                        "search_mode",
+                        description="File search mode. Use 'name' to search in file name or "
+                                    "'content' to search in file content",
+                    ): Or("name", "content"),
+                    Literal(
+                        "search_query",
+                        description="Query to search for. If search_mode is 'name', it's the file name. If 'content', "
+                        "it's the text within files.",
+                    ): str,
+                    Literal(
```
vasinov, line 205: Can be optional as well.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,335 @@
...
+                        "folder_path",
+                        description="Path of the Google Drive folder (like 'MainFolder/Subfolder1/Subfolder2') "
+                        "where the search should be performed. Specify 'root' to search in the "
```
vasinov, line 208: Once it's made optional, you don't need to prompt the LLM to provide "root" as a value.

vasinov (changes requested):

griptape/tools/google_drive/tool.py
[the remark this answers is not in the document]
vasinov: Nevermind, I see the change in one place, but it needs to cover all activities. I added comments in every individual case.

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,342 @@
...
+                    mime_type = file_info["mimeType"]
+    
+                    if mime_type in self.GOOGLE_EXPORT_MIME_MAPPING:
+                        export_mime = self.GOOGLE_EXPORT_MIME_MAPPING[mime_type]
+                        request = service.files().export_media(fileId=file_id, mimeType=export_mime)
+                    else:
+                        request = service.files().get_media(fileId=file_id)
+    
+                    downloaded_file = request.execute()
+                    logging.info(f"File '{path}' successfully downloaded.")
```
vasinov, line 171: Can remove this log.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,342 @@
...
+                    downloaded_files.append(BlobArtifact(downloaded_file))
+                else:
+                    logging.error(f"Could not find file: {path}")
+                    downloaded_files.append(
+                        ErrorArtifact(f"Could not find file: {path}")
```
vasinov, line 176: Let's log but not append, since `ListArtifact` will throw an error if you have objects of different types in `value`.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,342 @@
...
+                    )
+    
+            except HttpError as e:
+                logging.error(e)
+                downloaded_files.append(
```
vasinov, line 181: Same here.

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,342 @@
...
+                    ErrorArtifact(f"error downloading file '{path}' from Google Drive: {e}")
+                )
+            except MalformedError:
+                logging.error("MalformedError occurred")
+                downloaded_files.append(
```
vasinov, line 186: And here.

vasinov (changes requested):

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Sure!

griptape/tools/google_drive/tool.py
trish11953: Done!

griptape/tools/google_drive/tool.py
trish11953: Done!

tests/unit/tools/test_google_drive.py
vasinov, line 1: File should be called `test_google_drive_client.py`

tests/integration/tools/test_google_drive.py
vasinov, line 1: File should be called `test_google_drive_client.py`

griptape/tools/google_drive/tool.py
```diff
@@ -0,0 +1,333 @@
...
+                    mime_type = file_info["mimeType"]
+    
+                    if mime_type in self.GOOGLE_EXPORT_MIME_MAPPING:
+                        export_mime = self.GOOGLE_EXPORT_MIME_MAPPING[mime_type]
+                        request = service.files().export_media(fileId=file_id, mimeType=export_mime)
+                    else:
+                        request = service.files().get_media(fileId=file_id)
+    
+                    downloaded_file = request.execute()
+                    downloaded_files.append(BlobArtifact(downloaded_file))
```
vasinov, line 171: If the variable is not re-used and if the statement on the right side of the assignment is short, we should aim to combine it in the same statement. In this case, it should look like this:

```python
downloaded_files.append(
    BlobArtifact(request.execute())
)
```

It improves readability.

vasinov: Also, just realized that it looks a bit sus. Wanted to make sure `execute()` returns a byte string instead of a JSON object.

vasinov (changes requested):

tests/unit/tools/test_google_drive.py
trish11953: Done!

vasinov (approved):

vasinov (approved):

Issue #135, opened 2023-07-24
Bedrock prompt driver support


vasinov:
We need to add support for the upcoming [AWS Bedrock](https://aws.amazon.com/bedrock/) service in the form of the `AwsBedrockPromptDriver`.

collindutter:
https://github.com/griptape-ai/griptape/pull/220

PR #220, opened 2023-09-13
Add bedrock prompt driver

collindutter:
https://github.com/griptape-ai/griptape/issues/135


collindutter:
@vasinov I think we need to integrate https://github.com/griptape-ai/griptape/pull/222 before merging Bedrock. I'm noticing that even claude-v2 is outputting faulty json.

```python
from griptape.tasks import ToolTask
from griptape.tools import Calculator
from griptape.drivers import AmazonBedrockPromptDriver
from griptape.drivers import (
    BedrockClaudePromptModelDriver
)

anthropic_prompt_driver = AmazonBedrockPromptDriver(
    model="anthropic.claude-v2",
    prompt_model_driver_type=BedrockClaudePromptModelDriver,
)
pipeline = Pipeline(
    prompt_driver=anthropic_prompt_driver,
)
pipeline.add_tasks(
    ToolTask("What is 10^5", tool=Calculator()),
)

pipeline.run()
```

collindutter:
@vasinov We should also merge in https://github.com/griptape-ai/griptape/pull/225 before this.

griptape/drivers/embedding/amazon_bedrock_embedding_driver.py
```diff
@@ -0,0 +1,47 @@
...
+        default=Factory(
+            lambda self: AmazonBedrockTokenizer(model=self.model), takes_self=True
+        ),
+        kw_only=True,
+    )
+    session: boto3.Session = field(
+        default=Factory(lambda: boto3.Session()), kw_only=True
+    )
+    content_type: str = field(default="application/json", kw_only=True)
+    accept: str = field(default="application/json", kw_only=True)
```
vasinov, line 25: Could it ever be anything else?

griptape/drivers/embedding/amazon_bedrock_embedding_driver.py
```diff
@@ -0,0 +1,47 @@
...
+    bedrock_client: boto3.client = field(
+        default=Factory(
+            lambda self: self.session.client("bedrock"),
+            takes_self=True,
+        ),
+        kw_only=True,
+    )
+
+    def try_embed_string(self, string: str) -> list[float]:
+        text = string.replace(os.linesep, " ")
```
vasinov, line 35: Why remove new lines?

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
```diff
@@ -0,0 +1,43 @@
...
+
+if TYPE_CHECKING:
+    from griptape.utils import PromptStack
+
+
+@define
+class AmazonBedrockPromptDriver(BaseMultiModelPromptDriver):
+    session: boto3.Session = field(default=Factory(lambda: boto3.Session()), kw_only=True)
+    content_type: str = field(default="application/json", kw_only=True)
+    accept: str = field(default="application/json", kw_only=True)
```
vasinov, line 17: Same question: would this ever be something non-json?

griptape/drivers/prompt_model/bedrock_claude_prompt_model_driver.py
```diff
@@ -0,0 +1,49 @@
...
+from griptape.utils import PromptStack
+from griptape.drivers import BasePromptModelDriver
+from griptape.tokenizers import AnthropicTokenizer
+
+
+@define
+class BedrockClaudePromptModelDriver(BasePromptModelDriver):
+    tokenizer: AnthropicTokenizer = field(
+        default=Factory(
+            lambda: AnthropicTokenizer(model="anthropic.claude-v2"),
```
vasinov, line 13: Should we move model into a class property since Bedrock supports claude-v1 as well? And will likely support claude-v3 in the future.

griptape/drivers/prompt_model/bedrock_claude_prompt_model_driver.py
```diff
@@ -0,0 +1,49 @@
...
+        prompt_lines.append("Assistant:")
+
+        prompt = "\n".join(prompt_lines)
+        return {"prompt": prompt}
+
+    def prompt_stack_to_model_params(self, prompt_stack: PromptStack) -> dict:
+        prompt = self.prompt_driver.prompt_stack_to_string(prompt_stack)
+
+        return {
+            "anthropic_version": "bedrock-2023-05-31",
```
vasinov, line 38: Same here: should this be a class-level property?

griptape/drivers/prompt_model/bedrock_titan_prompt_model_driver.py
```diff
@@ -0,0 +1,48 @@
...
+from griptape.utils import PromptStack
+from griptape.drivers import BasePromptModelDriver
+from griptape.tokenizers import AmazonBedrockTokenizer
+
+
+@define
+class BedrockTitanPromptModelDriver(BasePromptModelDriver):
+    tokenizer: AmazonBedrockTokenizer = field(
+        default=Factory(
+            lambda: AmazonBedrockTokenizer(model="amazon.titan-tg1-large"),
```
vasinov, line 13: Move to class property?

griptape/mixins/exponential_backoff_mixin.py
```diff
@@ -9,7 +9,7 @@
 class ExponentialBackoffMixin(ABC):
     min_retry_delay: float = field(default=2, kw_only=True)
     max_retry_delay: float = field(default=10, kw_only=True)
-    max_attempts: int = field(default=10, kw_only=True)
+    max_attempts: int = field(default=1, kw_only=True)
```
vasinov, line 12: ?

griptape/tokenizers/anthropic_tokenizer.py
```diff
@@ -6,13 +6,16 @@
 @define(frozen=True)
 class AnthropicTokenizer(BaseTokenizer):
     DEFAULT_MODEL = "claude-2"
-    DEFAULT_MAX_TOKENS = 100000
-
+    MODEL_TO_MAX_TOKENS = {
+        "claude-2": 100000,
+        "anthropic.claude-v2": 8192
```
vasinov, line 11: This feels a bit awkward. Don't really know what to do about it.

vasinov (changes requested):

griptape/drivers/prompt_model/bedrock_claude_prompt_model_driver.py
collindutter: @vasinov I'm actually going to remove it. I think I had captured it from a network request, but it doesn't appear to be a documented parameter.

griptape/mixins/exponential_backoff_mixin.py
collindutter: Oops, didn't mean to commit this  🙃

griptape/tokenizers/amazon_bedrock_tokenizer.py
```diff
@@ -0,0 +1,55 @@
+import os
+import json
+import boto3
+from attr import define, field, Factory
+from griptape.tokenizers import BaseTokenizer
+
+
+@define(frozen=True)
+class AmazonBedrockTokenizer(BaseTokenizer):
```
collindutter, line 9: @vasinov this only works with titan models since they're the only ones that return a token count. Should we rename this to `AmazonBedrockTitanTokenizer`?

griptape/tokenizers/anthropic_tokenizer.py
collindutter: Would something like [this](https://github.com/griptape-ai/griptape/blob/main/griptape/tokenizers/tiktoken_tokenizer.py#L17) be any better?

griptape/tokenizers/amazon_bedrock_tokenizer.py
massi-ang: AI21 models served via Bedrock return the detail of each generated token, and the token count can be derived from that.

griptape/drivers/embedding/amazon_bedrock_embedding_driver.py
vasinov: Yeah, I think removing it is fine. We can always introduce it later if some model requires something custom. And in that case, I imagine, it would live in the model driver.

griptape/drivers/embedding/amazon_bedrock_embedding_driver.py
vasinov: Gotcha. Can we change this to `\n`, so it's consistent across the whole framework?

Separately, should we update all `\n` usage in the framework to `os.linesep`? For now, could just be a separate issue.

vasinov (changes requested):

griptape/tokenizers/amazon_bedrock_tokenizer.py
collindutter: Separated out `AmazonBedrockTokenizer` into `BedrockTitanTokenizer`, `BedrockJurassicTokenizer`, and `BedrockClaudeTokenizer`.

`BedrockTitanTokenizer` and `BedrockJurassicTokenizer` both use the token count from the `invoke_model`, though this is not ideal since it effectively doubles the cost. `BedrockJurassicTokenizer` can probably be swapped out for [ai21's tokenization API](https://docs.ai21.com/reference/tokenize-ref) but this requires an API key. Hopefully Amazon exposes proper tokenization APIs in the future.

`BedrockClaudeTokenizer` subclasses `AnthropicTokenizer` since we can use the same client-side tokenizer.

vasinov (approved):

PR #222, opened 2023-09-13
Improve action matching

collindutter:
Occasionally models will output actions with extra text i.e.
```
[09/13/23 14:28:55] INFO     Subtask e97a869778614e1f96640943d29eb990
                             Action: Here is the action to calculate 10^5 using the Calculator tool:

                             ```json
                             {
                               "type": "tool",
                               "name": "Calculator",
                               "activity": "calculate",
                               "input": {
                                 "values": {
                                   "expression": "10**5"
                                 }
                               }
                             }
                             ```

                             The Calculator tool can evaluate simple arithmetic expressions like this using Python.
```
This change updates the action matching regex to still capture the intended action JSON.


vasinov (changes requested):
This is a really good change. Can we add some tests that cover a couple of those edge cases?

vasinov (approved):

Issue #62, opened 2023-05-23
Add `Structure.rules` in addition to `Structure.rulesets`


vasinov:
Sometimes it makes sense to only add a small set of rules to a structure instead of multiple rulesets.

For that purpose we propose adding an optional `rules: list[Rule]` field to `Structure`s. It should only be possible to use one or the other.

PR #223, opened 2023-09-13
Add ability to use rules

collindutter:
#62 

```python
from griptape.rules import Rule, Ruleset
from griptape.tasks import PromptTask

agent = Pipeline(
    rulesets=[Ruleset("Name", rules=[Rule("Your name is Gustavo")])]
)
agent.add_task(PromptTask(rules=[Rule("Talk like a pirate")]))
agent.add_task(PromptTask(rules=[Rule("Talk in spanish")]))

agent.run("Who are you?")
```

```
[09/13/23 16:21:07] INFO     PromptTask d7c378f29fb84d9f8b2d4343b9e32c02
                             Input: Who are you?
[09/13/23 16:21:09] INFO     PromptTask d7c378f29fb84d9f8b2d4343b9e32c02
                             Output: Arr matey, I be Gustavo, at yer service!
                    INFO     PromptTask d0d6aee29b3949fea711322d0b762c10
                             Input: Who are you?
[09/13/23 16:21:11] INFO     PromptTask d0d6aee29b3949fea711322d0b762c10
                             Output: Soy Gustavo, tu asistente de conversación amigable. ¿Cómo puedo ayudarte hoy?
```

### DIFF
--- a/griptape/tasks/prompt_task.py
+++ b/griptape/tasks/prompt_task.py
@@ -44,15 +44,17 @@ class PromptTask(BaseTextInputTask):
 
     @property
     def all_rulesets(self) -> list[Ruleset]:
+        structure_rulesets = []
         if self.structure.rulesets:
             structure_rulesets = self.structure.rulesets
-        else:
-            structure_rulesets = [Ruleset(name="Structure Ruleset", rules=self.structure.rules)]
+        elif self.structure.rules:
+            structure_rulesets = [Ruleset(name="Default Ruleset", rules=self.structure.rules)]
 
+        task_rulesets = []
         if self.rulesets:
             task_rulesets = self.rulesets
-        else:
-            task_rulesets = [Ruleset(name="Task Ruleset", rules=self.rules)]
+        elif self.rules:
+            task_rulesets = [Ruleset(name="Additional Ruleset", rules=self.rules)]
 
         return structure_rulesets + task_rulesets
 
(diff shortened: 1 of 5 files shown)


griptape/tasks/prompt_task.py
```diff
@@ -24,10 +25,36 @@ class PromptTask(BaseTextInputTask):
...
+        if self.rulesets:
+            raise ValueError("can't have both rules and rulesets specified")
 
     @property
     def all_rulesets(self) -> list[Ruleset]:
-        return self.structure.rulesets + self.rulesets
+        if self.structure.rulesets:
+            structure_rulesets = self.structure.rulesets
+        else:
+            structure_rulesets = [Ruleset(name="Structure Ruleset", rules=self.structure.rules)]
```
vasinov, line 50: We should check here if `rules` are not empty.

griptape/tasks/prompt_task.py
```diff
@@ -24,10 +25,36 @@ class PromptTask(BaseTextInputTask):
...
+
+        if self.rulesets:
+            task_rulesets = self.rulesets
+        else:
+            task_rulesets = [Ruleset(name="Task Ruleset", rules=self.rules)]
```
vasinov, line 55: We should check here if rules are not empty.

griptape/tasks/prompt_task.py
vasinov: Also, not sure about the `name` here.... The problem is that this goes into the prompt and the LLM has no idea what "structure" is. May be call it `Default Ruleset`?

griptape/tasks/prompt_task.py
vasinov: ...and let's rename this to `Additional Ruleset`?

Still doesn't quite feel right. Any other ideas?

vasinov (changes requested):

vasinov (approved):

PR #231, opened 2023-09-15
Add default tool memory embedding driver, cleanup vector store, query engine, and tool memory defaults

collindutter:
https://github.com/orgs/griptape-ai/discussions/230

### DIFF
--- a/griptape/drivers/vector/base_vector_store_driver.py
+++ b/griptape/drivers/vector/base_vector_store_driver.py
@@ -26,10 +26,7 @@ class BaseVectorStoreDriver(ABC):
         meta: Optional[dict] = None
         namespace: Optional[str] = None
 
-    embedding_driver: BaseEmbeddingDriver = field(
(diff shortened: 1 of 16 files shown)


vasinov (commented):
This only works if the prompt driver is OpenAI. Otherwise, this breaks. Let's not make this change but instead document embedding driver API keys better.

Or add a default embedding driver to structure, so it propagates to the underlying components like memory?

vasinov (approved):
In addition to your changes, I cleaned up a bunch of defaults that shouldn't exist anymore. We want to control one happy path at the structure level to avoid dependency injection complexity. If the user wants to something more advanced, they should be able to set underlying dependencies depending on their needs.

We should document this better.

collindutter:
@vasinov what about `VectorStoreClient`? With these changes, users need to initialize it with: `VectorStoreClient` -> `VectorQueryEngine` -> `VectorStoreDriver` -> `EmbeddingDriver`.

Should we set a happy path in `VectorStoreClient` too?

Issue #239, opened 2023-09-18
Unable to complete the following prompt


hyusetiawan:
**Describe the bug**
Unable to complete a prompt that I thought would be simple to complete, the example prompts are:
- `how many days until Sept 23rd?`
- `what is the date 3 days from today?`
Here's an example log of the output:
```
Q: how many days until Sept 23rd?
processing...
[09/18/23 04:26:11] INFO     ToolkitTask 7e2af4d35e944089a8f6c47f21f222ee
                             Input: how many days until Sept 23rd?
[09/18/23 04:26:13] INFO     Subtask 72391817347e4b88a85b82f345770137
                             Thought: I can use the DateTime tool to get the current date and time, and then use the Calculator tool to calculate the difference between the two dates.
                             Action: {"type":"tool","name":"DateTime","activity":"get_current_datetime"}
                    INFO     Subtask 72391817347e4b88a85b82f345770137
                             Observation: 2023-09-18 04:26:13.997353
[09/18/23 04:26:16] INFO     Subtask 1121b303ef924961a5fac6a8755f1c72
                             Assistant: Thought: I can use the Calculator tool to calculate the difference between the current date and Sept 23rd.
                             Action: {"type": "tool", "name": "Calculator", "activity": "calculate", "input": {"values": {"expression": "datetime.datetime(2023, 9, 23) - datetime.datetime(2023, 9, 18)"}}}

                             User:
                    INFO     Subtask 1121b303ef924961a5fac6a8755f1c72
                             Observation: name 'datetime' is not defined
[09/18/23 04:26:19] INFO     Subtask 48557b0de8e44fdeba0b90d0ac751f26
                             Assistant: Thought: I need to use the Computer tool to execute the code to get the difference between the two dates.
                             Action: {"type": "tool", "name": "Computer", "activity": "execute_code", "input": {"values": {"code": "import datetime\ndatetime.datetime(2023, 9, 23) - datetime.datetime(2023, 9, 18)", "filename": "difference.py"}}}

                             User:
                    INFO     Subtask 48557b0de8e44fdeba0b90d0ac751f26
                             Observation: Output of "Computer.execute_code" was stored in memory with memory_name "TextToolMemory" and artifact_namespace "7dd313205b8f4e4cb88477395e9d278f"
[09/18/23 04:26:22] INFO     ToolkitTask 7e2af4d35e944089a8f6c47f21f222ee
[...]

PR #257, opened 2023-09-21
Adds forward looking example, handle parse error, update tests

collindutter:
https://github.com/griptape-ai/griptape/issues/239

### DIFF
--- a/griptape/tools/date_time/tool.py
+++ b/griptape/tools/date_time/tool.py
@@ -34,9 +34,10 @@ class DateTime(BaseTool):
         try:
             date_string = params["values"]["relative_date_string"]
             relative_datetime = parse(date_string)
-            if not relative_datetime:
-                return ErrorArtifact("invalid date string")
 
-            return TextArtifact(str(relative_datetime))
+            if relative_datetime:
+                return TextArtifact(str(relative_datetime))
+            else:
+                return ErrorArtifact("invalid date string")
         except Exception as e:
             return ErrorArtifact(f"error getting current datetime: {e}")
--- a/griptape/tools/date_time/tool.py
+++ b/griptape/tools/date_time/tool.py
@@ -25,7 +25,8 @@ class DateTime(BaseTool):
         "schema": Schema({
             Literal(
                 "relative_date_string",
-                description='Relative date in English. For example, "now EST", "20 minutes ago", or "yesterday at 2pm"'
+                description='Relative date in English. For example, "now EST", "20 minutes ago", '
+                            '"in 2 days", "3 months, 1 week and 1 day ago", or "yesterday at 2pm"'
             ): str
         })
     })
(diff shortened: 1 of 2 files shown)


collindutter:
I tried adding an activity for getting a time delta, but the LLM got confused on when to use `get_datetime_delta` and when to use `get_relative_datetime`. I think this should be a future PR.
```python
    @activity(config={
        "uses_default_memory": False,
        "description": "Can be used to get get a time delta between two dates.",
        "schema": Schema({
            Literal(
                "date_1",
                description='The first date. Must be a valid date string.'
            ): str,
            Literal(
                "date_2",
                description='The second date. Must be a valid date string.'
            ): str
        })
    })
    def get_datetime_delta(self, params: dict) -> BaseArtifact:
        try:
            date_1 = parse(params["values"]["date_1"])
            date_2 = parse(params["values"]["date_2"])
            if not date_1:
                return ErrorArtifact("invalid date string for date_1")
            if not date_2:
                return ErrorArtifact("invalid date string for date_2")

            delta = date_1 - date_2

            return TextArtifact(delta)
        except Exception as e:
            return ErrorArtifact(f"error getting current datetime: {e}")
```

griptape/tools/date_time/tool.py
```diff
@@ -25,14 +25,17 @@ def get_current_datetime(self, _: dict) -> BaseArtifact:
...
         })
     })
     def get_relative_datetime(self, params: dict) -> BaseArtifact:
         try:
             date_string = params["values"]["relative_date_string"]
             relative_datetime = parse(date_string)
+            if not relative_datetime:
+                return ErrorArtifact("invalid date string")
 
             return TextArtifact(str(relative_datetime))
```
vasinov, line 40: Let's always have an `if-else` structure for all returns. General rule should be "no early returns" as they make code harder to follow and can result in subtle bugs.

vasinov (changes requested):

griptape/tools/date_time/tool.py
```diff
@@ -34,6 +35,9 @@ def get_relative_datetime(self, params: dict) -> BaseArtifact:
             date_string = params["values"]["relative_date_string"]
             relative_datetime = parse(date_string)
 
-            return TextArtifact(str(relative_datetime))
+            if not relative_datetime:
```
vasinov, line 38: Sorry to nit more but let's also make sure `if-else` structures read "naturally" with a positive `if` coming first and a negative being naturally tucked under `else`. Makes it a little easier to follow the code when developers don't have to negate a condition in their head while reading the code.

vasinov (changes requested):

griptape/tools/date_time/tool.py
collindutter: Makes sense, updated!

vasinov (approved):

PR #264, opened 2023-09-21
Lazily initialize Prompt Model Driver tokenizer

collindutter:
Making this change so that the Prompt Driver (and its session) is already set by the time the Tokenizer is initialized


vasinov (changes requested):
Could you add a comment somewhere in the code explaining why we can't use a regular `field` here? Makes sense now but I'm sure it won't be as obvious in a couple of months :)

vasinov (approved):

Issue #203, opened 2023-09-07
Add control parameters to TextSummaryTask 


shhlife:
**Is your feature request related to a problem? Please describe.**
TextSummaryTask is a great way to quickly summarize input text without needing to use COT. At the moment there isn't much control provided to the developer on how the given text should be summarized.

**Describe the solution you'd like**

It would be great to have some additional parameters (names TBD) such as:
* length
* target audience
* format

This would let us summarize with a specifc length (3 sentences, 2 paragraphs, etc) or for an audience (child, executive summary) or in a format (bullet points, yaml, json) etc.

collindutter:
Implemented in https://github.com/griptape-ai/griptape/issues/342

PR #277, opened 2023-09-23
Added extra parameters to TextSummaryTask

ashkaaar:
Added the parameters regarding the issue https://github.com/griptape-ai/griptape/issues/203

length
target audience
format

Unit tests: passed
waiting for review and suggestions

### DIFF
--- a/griptape/engines/summary/base_summary_engine.py
+++ b/griptape/engines/summary/base_summary_engine.py
@@ -2,12 +2,11 @@ from abc import ABC, abstractmethod
 from attr import define
 from griptape.artifacts import TextArtifact, BaseArtifact
 
-
 @define
 class BaseSummaryEngine(ABC):
-    def summarize_text(self, text: str) -> str:
-        return self.summarize_artifacts([TextArtifact(text)]).value
+    def summarize_text(self, text: str, length: int = None, target_audience: str = None, format: str = None) -> str:
+        return self.summarize_artifacts([TextArtifact(text)], length, target_audience, format).value
 
     @abstractmethod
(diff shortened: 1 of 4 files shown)


griptape/engines/summary/base_summary_engine.py
```diff
@@ -2,12 +2,11 @@
 from attr import define
 from griptape.artifacts import TextArtifact, BaseArtifact
 
-
 @define
 class BaseSummaryEngine(ABC):
-    def summarize_text(self, text: str) -> str:
-        return self.summarize_artifacts([TextArtifact(text)]).value
+    def summarize_text(self, text: str, length: int = None, target_audience: str = None, format: str = None) -> str:
```
vasinov, line 7: Those should be class properties for:
- DRYness across class methods and
- ease of integration with tasks and `TextMemoryActivitiesMixin.summarize`.

vasinov: Additionally, we always use `Optional`s in type hints if the default value is `None`. In this case it should be `Optional[int]`, `Optional[str]`, and `Optional[str]`.

vasinov (changes requested):
Thanks for the PR!

collindutter:
@ashkaaar I am going to close this PR in favor of the work that will be done on https://github.com/griptape-ai/griptape/issues/342. Implementing this functionality as described in the other issue gives users more control over the output, and it provides a more consistent experience across Tasks.

collindutter:
No problem at all! Thanks for the original PR 🙂

Issue #243, opened 2023-09-18
Add a `PgvectorVectorStoreDriver` driver for pgvector


vasinov:
Griptape should support [pgvector](https://github.com/pgvector/pgvector) one of the most popular vector store solutions on the market.

To accomplish that, we need to add `PgvectorVectorStoreDriver` that implements all abstract`BaseVectorStoreDriver` methods.

PR #286, opened 2023-09-26
Add support for PGVector vector store driver

andrewfrench:
This PR implements a PGVector vector store driver (#243). Additionally, it creates a test dependency on a PGVector-flavor Postgres container (I'm using `ankane/pgvector`) present in the test environment. For CI tests, this has been added to the Github Actions workflow specificaiton.

New tests for the PGVector driver are located in the tests/unit/ directory. These tests rely on a live Postgres instance with the PGVector extension installed. If a Postgres instance is not present, the tests relying on it will be skipped.


.github/workflows/unit-tests.yml
```diff
@@ -13,6 +13,20 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.9", "3.10", "3.11"]
+    services:
+      postgres:
+        image: ankane/pgvector:v0.5.0
+        env:
+            POSTGRES_USER: postgres
+            POSTGRES_PASSWORD: postgres
+            POSTGRES_DB: postgres
```
collindutter, line 22: Let's store these values in Github secrets. I.e. `${{ secrets.UNIT_POSTGRES_USER }}`

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,159 @@
...
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.orm import Session
+from pgvector.sqlalchemy import Vector
+
+
+@define
+class PgVectorVectorStoreDriver(BaseVectorStoreDriver):
+    """
+    A vector store driver to Postgres using the PGVector extension.
```
collindutter, line 16: Put the docstring description on the first line, I think it impacts how reference docs are generated. Same for functions throughout the class. See https://google.github.io/styleguide/pyguide.html#384-classes.

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Attributes:
+        connection_string: The connection string for the Postgres database instance.
+        create_engine_params: Additional configuration params passed when creating the database connection.
+        engine: An optional sqlalchemy Postgres engine to use.
+    """
+
+    connection_string: Optional[str] = field(default=None, kw_only=True)
```
collindutter, line 24: Remove `Optional`, `default`

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,159 @@
...
+    create_engine_params: Optional[dict] = field(factory=dict, kw_only=True)
+    engine: Optional[Engine] = field(
```
collindutter, line 26: Remove `Optional`. We only want to mark `Optional` if  `default=None`.

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,159 @@
...
+        default=Factory(
+            lambda self: create_engine(self.connection_string, **self.create_engine_params), takes_self=True
+        )
+    )
+
+    class VectorModel(declarative_base()):
```
collindutter, line 32: I _think_ add `@dataclass` decorator. Not sure `declarative_base()` changes anything.

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,159 @@
...
+        __tablename__ = "griptape_vectors"
```
collindutter, line 33: Probably want to be able to override this default with class property.

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,159 @@
...
+
+        id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4, unique=True, nullable=False)
+        vector = Column(Vector())
+        namespace = Column(String)
+        meta = Column(JSON)
+
+    def install_postgres_extensions(self) -> None:
```
collindutter, line 40: Several of our prompt drivers implement functionality to "initialize" themselves. i.e. set up indexes/tables/etc. Should we move this to the base class so there is a standard interface?

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Ensures the uuid-ossp and vector extensions are installed in the database.
+        """
+        self.engine.execute('CREATE EXTENSION IF NOT EXISTS "uuid-ossp";')
+        self.engine.execute('CREATE EXTENSION IF NOT EXISTS "vector";')
+
+    def create_schema(self) -> None:
```
collindutter, line 47: Same comment as above. Maybe move to some sort of initialization function?

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,159 @@
...
+        namespace: Optional[str] = None,
+        include_vectors: bool = False,
+        distance_metric: str = "cosine_distance",
+        **kwargs
+    ) -> list[BaseVectorStoreDriver.QueryResult]:
+        """
+        Performs a search on the collection to find vectors similar to the provided input vector,
+        optionally filtering to only those that match the provided namespace.
+        """
+        count = count if count else BaseVectorStoreDriver.DEFAULT_QUERY_COUNT
```
collindutter, line 126: Can we move this to the parameter default? Instead of count defaulting to `None` have it default to `BaseVectorStoreDriver.DEFAULT_QUERY_COUNT`

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,159 @@
...
+
+        distance_metrics = {
+            "cosine_distance": self.VectorModel.vector.cosine_distance,
+            "l2_distance": self.VectorModel.vector.l2_distance,
+            "inner_product": self.VectorModel.vector.max_inner_product,
+        }
+        op = distance_metrics[distance_metric]
+        if op is None:
```
collindutter, line 134: I think this will raise a `KeyError` if an invalid `distance_metric` is provided. Update check to `if distance_metric not in distance_metrics`

.github/workflows/unit-tests.yml
```diff
@@ -13,6 +13,20 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.9", "3.10", "3.11"]
+    services:
+      postgres:
+        image: ankane/pgvector:v0.5.0
```
collindutter, line 18: Should we use a base postgres image and then use the initialization functions to set up pgvector?

tests/unit/drivers/vector/test_pgvector_vector_store_driver.py
collindutter, line 1: I'm not well-versed enough on all testing terminologies. Since this connecting to a "real" instance, does this fall more under integration tests?

collindutter (changes requested):

tests/unit/drivers/vector/test_pgvector_vector_store_driver.py
andrewfrench: I've moved it back and forth at least a couple times and I'm not married to either option. I will say that we currently don't run integration tests on pull requests and I think this is something that should be run against every revision. At the end of the day, this is still testing only the PGVector driver component, albeit with a non-mocked backend. An integration test should probably be more end-to-end, from agent to storage, and I don't think a lower level unit test without the actual Postgres backend would provide much value.

.github/workflows/unit-tests.yml
andrewfrench: That's a reasonable option, but installing PGVector is [non-trivial](https://github.com/pgvector/pgvector#installation) (as in, it doesn't ship out of the box with Postgres and we'd need to orchestrate compiling/installing it). This image handles that step for us, but we still have to enable the extension.

griptape/drivers/vector/pgvector_vector_store_driver.py
andrewfrench: Will this prevent us from supporting bring-your-own-engine here? The approach I aimed for is either provide the connection string (and we'll build the engine) or provide the engine directly.

griptape/drivers/vector/pgvector_vector_store_driver.py
andrewfrench: I believe `declarative_base()` provides the `.metadata.create_all()` method that we use to create the schema.

vasinov (commented):
Great progress! Can you please re-request my review once @collindutter's comments are addressed? Thanks!

tests/unit/drivers/vector/test_pgvector_vector_store_driver.py
collindutter: I agree with your points, though this change makes running unit tests locally a bit trickier.

.github/workflows/unit-tests.yml
collindutter: Ah I didn't realize that this is the [official image](https://github.com/pgvector/pgvector#docker), let's just stick with this then 👍

griptape/drivers/vector/pgvector_vector_store_driver.py
collindutter: Good point, I was looking at the [other drivers](https://github.com/griptape-ai/griptape/blob/main/griptape/drivers/vector/marqo_vector_store_driver.py#L21) as examples but I think it's actually incorrectly implemented on them. If both field are optional, maybe we should add a [validator](https://github.com/griptape-ai/griptape/blob/main/griptape/mixins/activity_mixin.py#L13) that ensures that at least one is set.

griptape/drivers/vector/pgvector_vector_store_driver.py
collindutter: @vasinov what're your thoughts on this?

vasinov: I originally decided against it because, at the time, it seemed like different vector store initialization processes were too unique for an abstract method and it felt like more of a bonus thing that _some_ drivers would provide.

We should re-evaluate that assumption though. It might be nice to have a standard method with `**kwargs` in it that child classes optionally implement.

.github/workflows/unit-tests.yml
andrewfrench: Updated within the unit test spec but I'll need someone with the permissions to do so to create the secrets on the Github side. Let's not worry about making them secure because we're connecting to an ephemeral database and the values should be pretty clear for someone debugging local testing.

griptape/drivers/vector/pgvector_vector_store_driver.py
andrewfrench: I've created an initialization method that we can propagate to other vector store classes if we'd like. We can discuss further.

andrewfrench:
I've addressed @collindutter's comments, but tests will fail until we can add the relevant Github secrets for the Actions Postgres configuration (the tests pass locally with the Postgres container running).

cc @vasinov

andrewfrench:
Closing this PR in favor of #313 as this was created from a fork and is not allowed access to Github secrets.

PR #288, opened 2023-09-27
Add prompt driver streaming

collindutter:


griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
```diff
@@ -29,16 +30,40 @@ def try_run(self, prompt_stack: PromptStack) -> TextArtifact:
...
             payload.update(model_input)
 
-        response = self.bedrock_client.invoke_model(
-            modelId=self.model,
-            contentType='application/json',
-            accept='application/json',
-            body=json.dumps(payload),
-        )
+        if self.stream:
+            response = self.bedrock_client.invoke_model_with_response_stream(
```
massi-ang, line 34: Make sure that this design also works for Cohere models. To enable streaming with Cohere command text model via Bedrock, one need to also pass a `"stream": true` parameter in the body as explained in https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters.html#model-parameters-cohere

collindutter: We actually didn't have access to Cohere until a few days ago so we're missing a Prompt Model Driver for it. I created a separate [issue](https://github.com/griptape-ai/griptape/issues/312) for it. Depending on timing I may implement streaming on that PR.

griptape/drivers/prompt/anthropic_prompt_driver.py
```diff
@@ -24,17 +25,20 @@ class AnthropicPromptDriver(BasePromptDriver):
...
-            temperature=self.temperature,
+            **self._base_params(prompt_stack)
         )
-
         return TextArtifact(value=response.completion)
 
+    def try_stream(self, prompt_stack: PromptStack) -> Iterator[TextArtifact]:
+        response = anthropic.Anthropic(api_key=self.api_key).completions.create(
+            **self._base_params(prompt_stack),
+            stream=True
```
collindutter, line 36: Explicitly passing `True` instead of `self.stream` for 2 reasons:
1. Anthropic's SDK has the type for `stream` as `Literal[True]`. Meaning that passing `self.stream` yields the following pyright error: `Argument of type "bool" cannot be assigned to parameter "stream" of type "Literal[True]" in function "create" "bool" cannot be assigned to type "Literal[True]`. Other LLM SDKs do not have this issue.
2. `self.stream` is really intended for `BasePromptDriver.run()` to determine whether to call `try_run` or `try_stream`. If we use it here too, we introduce the potential for a user setting `stream=False` in the Prompt Driver, but then calling `try_stream`. With this implementation, `try_stream` will always pass the correct streaming parameters to the LLM.

Let me know if I'm thinking about this wrong.

andrewfrench (dismissed):

griptape/drivers/prompt/amazon_sagemaker_prompt_driver.py
```diff
@@ -43,3 +43,6 @@ def try_run(self, prompt_stack: PromptStack) -> TextArtifact:
             return self.prompt_model_driver.process_output(decoded_body)
         else:
             raise Exception("model response is empty")
+
+    def try_stream(self, _: PromptStack) -> Iterator[TextArtifact]:
```
vasinov, line 47: Should we add a validation for the `stream` property instead, so it doesn't fail at runtime?

griptape/drivers/prompt/hugging_face_hub_prompt_driver.py
```diff
@@ -70,6 +71,9 @@ def try_run(self, prompt_stack: PromptStack) -> TextArtifact:
...
                 )
             else:
-                raise Exception("Completion with more than one choice is not supported yet.")
+                raise Exception("completion with more than one choice is not supported yet")
         else:
-            raise Exception(f"Only models with the following tasks are supported: {self.SUPPORTED_TASKS}")
+            raise Exception(f"only models with the following tasks are supported: {self.SUPPORTED_TASKS}")
+
+    def try_stream(self, _: PromptStack) -> Iterator[TextArtifact]:
+        raise NotImplementedError("streaming is not supported")
```
vasinov, line 79: Use validation instead?

griptape/drivers/prompt/hugging_face_pipeline_prompt_driver.py
```diff
@@ -61,6 +62,9 @@ def try_run(self, prompt_stack: PromptStack) -> TextArtifact:
...
                 )
             else:
-                raise Exception("Completion with more than one choice is not supported yet.")
+                raise Exception("completion with more than one choice is not supported yet")
         else:
-            raise Exception(f"Only models with the following tasks are supported: {self.SUPPORTED_TASKS}")
+            raise Exception(f"only models with the following tasks are supported: {self.SUPPORTED_TASKS}")
+
+    def try_stream(self, _: PromptStack) -> Iterator[TextArtifact]:
+        raise NotImplementedError("streaming is not supported")
```
vasinov, line 70: Use validation instead?

griptape/drivers/prompt_model/bedrock_jurassic_prompt_model_driver.py
```diff
@@ -14,6 +14,11 @@ class BedrockJurassicPromptModelDriver(BasePromptModelDriver):
     _tokenizer: BedrockJurassicTokenizer = field(default=None, kw_only=True)
     prompt_driver: Optional[AmazonBedrockPromptDriver] = field(default=None, kw_only=True)
 
+    @prompt_driver.validator
+    def validate_prompt_driver(self, _, prompt_driver):
+        if prompt_driver and prompt_driver.stream:
```
vasinov, line 19: Should we add a `supports_streaming` property to prompt model drivers and validate at the Bedrock prompt driver level? This way this pattern can be easily extended to HF and SM prompt/model drivers in the future.

griptape/events/finish_structure_run_event.py
```diff
@@ -0,0 +1,7 @@
+from attrs import define
+from griptape.events.base_event import BaseEvent
+
+
+@define
+class FinishStructureRunEvent(BaseEvent):
+    pass
```
vasinov, line 7: Change to `...` for consistency with other empty classes/methods.

griptape/events/start_structure_run_event.py
```diff
@@ -0,0 +1,7 @@
+from attrs import define
+from griptape.events.base_event import BaseEvent
+
+
+@define
+class StartStructureRunEvent(BaseEvent):
+    pass
```
vasinov, line 7: Change to `...` for consistency with other empty classes/methods.

griptape/drivers/prompt/amazon_sagemaker_prompt_driver.py
```diff
@@ -13,7 +13,7 @@
 @define
 class AmazonSageMakerPromptDriver(BaseMultiModelPromptDriver):
     session: boto3.Session = field(default=Factory(lambda: boto3.Session()), kw_only=True)
-    sagemaker_client: boto3.client = field(
+    sagemaker_client: Any = field(
```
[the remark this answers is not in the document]
collindutter: `boto3.client` is not a valid type. From pyright:

`Pyright: Expected type expression but received "(*args: Unknown, **kwargs: Unknown) -> (Unknown | _)" [reportGeneralTypeIssues]`

[boto3 stubs](https://pypi.org/project/boto3-stubs/) seems to be the proper solution. Maybe a separate PR?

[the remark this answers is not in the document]
vasinov: Makes sense.

vasinov (approved):

PR #289, opened 2023-09-28
duckduckgo_search Tool Implementation ✅

agniiva:
Implementation of duckduckgo_search 
Usage:
`from griptape.tools.duckduckgo_search import DDGSearch`

`tools=[DDGSearch()]`


collindutter:
Thanks for the PR @agniiva and apologies for the slow response! DDG is definitely a search engine that we'd like to support, though we'd like to implement it, and other search backends, in a more generalized way. I am going to close this PR in favor of a [new issue](https://github.com/griptape-ai/griptape/issues/383) to track this development.

PR #290, opened 2023-09-28
Fix google drive tests

collindutter:


griptape/tools/google_drive/tool.py
```diff
@@ -149,37 +145,44 @@ def save_content_to_drive(self, params: dict) -> ErrorArtifact | InfoArtifact:
...
+                        logging.error(f"Could not find file: {path}")
+        
+                except HttpError as e:
+                    logging.error(e)
+        
+                except MalformedError:
+                    logging.error(f"MalformedError occurred while downloading file '{path}' from Google Drive")
+        
+            return ListArtifact(downloaded_files)
+        except MalformedError:
```
vasinov, line 183: Let's catch all errors here.

griptape/tools/google_drive/tool.py
```diff
@@ -244,7 +250,7 @@ def search_files(self, params: dict) -> ListArtifact | ErrorArtifact:
             logging.error("MalformedError occurred")
             return ErrorArtifact("error searching for file due to malformed credentials")
 
-    def _build_client(self, scopes: list[str]) -> Resource:
+    def _build_client(self, scopes: list[str]) -> Any:
```
vasinov, line 253: `Any`?

vasinov (changes requested):

griptape/tools/google_drive/tool.py
```diff
@@ -173,13 +176,16 @@ def download_files(self, params: dict) -> ListArtifact | ErrorArtifact:
...
-            except HttpError as e:
-                logging.error(e)
-    
-            except MalformedError:
-                logging.error(f"MalformedError occurred while downloading file '{path}' from Google Drive")
-    
-        return ListArtifact(downloaded_files)
+            return ListArtifact(downloaded_files)
+        except HttpError as e:
+            logging.error(e)
```
collindutter, line 181: @vasinov should we remove the `logging.error`s? None of the other tools really have it, and seems redundant since we're already returning an `ErrorArtifact` which could take care of the logging.

griptape/tools/google_drive/tool.py
collindutter: We can't use the `Resource` type because it comes from google. And the `Resource` type is dynamically generated anyways so there aren't really useful type hints in it.

griptape/tools/google_drive/tool.py
vasinov: Agreed.

vasinov (approved):

PR #298, opened 2023-09-29
Split dependencies into extras

### LINKED ISSUE #134: How should we handle optional dependencies?
As the size of the framework grows, how should we think about handling non-core, aka driver, dependencies?

collindutter:
Resolves #134

This PR does the following:
- Moves certain dependencies to the [extras](https://python-poetry.org/docs/pyproject/#extras) section to reduce overall package size. The core dependencies are targeted towards OpenAi.
- Marks the test and dev dependency groups as optional. 
- Removes un-used packages `python-dotenv`, `stopit`, `fastapi`, `uvicorn`, `python-decouple`.

**This reduces the installation size from 447MB to 77MB**

There are a few affected workflows:

User wants to install griptape with everything it has to offer:
Poetry: `poetry install griptape --all-extras` or `poetry install griptape --extras "all"`
Pip: `pip install griptape[all]`

User wants to install griptape with only the core dependencies:
Poetry: `poetry install griptape`
Pip: `pip install griptape`

User wants to install dev or test dependencies:
Poetry: `poetry install griptape --with test,dev`
Pip: Pip does not support installing poetry's [optional dependency groups](https://python-poetry.org/docs/managing-dependencies/#dependency-groups)

Some questions:
1. Is it confusing for end users that they need to opt out of dependencies? [Moto](https://docs.getmoto.org/en/latest/docs/getting_started.html) does this, maybe we just need to explain it in documentation?
2. Are the groupings too granular? And for something like `trafilatura`, it is used by the `WebLoader` which is used by the `WebScraper` tool. Should the group be `tool-web-scraper` or `loader-web`? Both?
3. How do we convey in documentation which extras users need to install?


vasinov (commented):
A general comment: what does customer experience look like with this change when they run `pip install griptape`?

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
```diff
@@ -1,18 +1,21 @@
...
 from typing import TYPE_CHECKING, Any
-import boto3
 from attr import define, field, Factory
 from griptape.artifacts import TextArtifact
+from griptape.utils import import_optional_dependency
 from .base_multi_model_prompt_driver import BaseMultiModelPromptDriver
 
 if TYPE_CHECKING:
     from griptape.utils import PromptStack
+    import boto3
```
andrewfrench, line 11: I'm not familiar with this pattern, so just want to air out my understanding of what's happening here. `boto3` is an expensive import so we don't want to actually import it in the case where this file gets imported but isn't actually used. Then, on [line 17](https://github.com/griptape-ai/griptape/pull/298/files#diff-9464a93e53508f56e9787cb4e4a6192711a13ea4a00573bdd151298242bb31e9R17), we import it regardless because we're now actually using code that depends on the import. Is that right?

griptape/drivers/sql/amazon_redshift_sql_driver.py
```diff
@@ -1,14 +1,17 @@
+from __future__ import annotations
```
andrewfrench, line 1: Is this imported for side effects, unused, or is there some other magic happening here?

griptape/tokenizers/cohere_tokenizer.py
```diff
@@ -1,15 +1,18 @@
...
+    from cohere import Client
 
 @define(frozen=True)
 class CohereTokenizer(BaseTokenizer):
     DEFAULT_MODEL = "command"
     MAX_TOKENS = 2048
 
     model: str = field(default=DEFAULT_MODEL, kw_only=True)
-    client: cohere.Client = field(kw_only=True)
+    client: Client = field(kw_only=True)
```
andrewfrench, line 15: If this is conditionally imported above, do we need `import_optional_dependency()` here?

andrewfrench (commented):
A couple comments, mostly to check my own understanding.

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
collindutter: Code guarded by `TYPE_CHECKING` will only get run by a static type checker, not during actual runtime.

So we can import `boto3` for its  _type_ `boto3.Session` without actually importing the module.

Then, the actual module import is done at runtime lazily using `import_optional_dependency("boto3")`

griptape/drivers/sql/amazon_redshift_sql_driver.py
collindutter: Yeah it's some python magic that lets us use the `TYPE_CHECKING` functionality even thought it's not technically released in our current python version. You can read more about it [here](https://stackoverflow.com/questions/61544854/from-future-import-annotations).

griptape/tokenizers/cohere_tokenizer.py
collindutter: We're just using the type here, not the actual module, so no need for `import_optional_dependency()`

andrewfrench: Intersting, thank you! Lots to learn about the Python import model.

andrewfrench (approved):

griptape/utils/import_utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+INSTALL_MAPPING = {
+    "huggingface_hub": "huggingface-hub",
+    "pinecone": "pinecone-client",
+    "opensearchpy": "opensearch-py",
+    "requests_aws4auth": "requests-aws4auth",
+}
+
+
+def import_optional_dependency(name: str) -> Optional[ModuleType]:
```
vasinov, line 14: Should we extend this method with a custom mapping in addition to `INSTALL_MAPPING` (that we can may be rename to `DEFAULT_INSTALL_MAPPING`), so it's more flexible?

Can add an optional param like `install_mapping: Optional[dict] = None` and merge it with:

```python
mapping = DEFAULT_INSTALL_MAPPING

if install_mapping:
    mapping.update(install_mapping)

package_name = mapping.get(name)
```

vasinov (changes requested):

vasinov (approved):

griptape/utils/import_utils.py
vasinov: Ignore this comment :)

vasinov (approved):

PR #313, opened 2023-10-02
Add support for PGVector vector store driver

andrewfrench:
This PR implements a PGVector vector store driver (closes https://github.com/griptape-ai/griptape/issues/243). Additionally, it creates a test dependency on a PGVector-flavor Postgres container (I'm using ankane/pgvector) present in the test environment. For CI tests, this has been added to the Github Actions workflow specification.

New tests for the PGVector driver are located in the tests/unit/ directory. These tests rely on a live Postgres instance with the PGVector extension installed. If a Postgres instance is not present, the tests relying on it will be skipped.

This is a copy of the closed #286 which was created from a fork and not allowed access to Github secrets.


griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,170 @@
...
+        create_engine_params: Additional configuration params passed when creating the database connection.
+        engine: An optional sqlalchemy Postgres engine to use.
+    """
+
+    connection_string: str = field(default=None, kw_only=True)
+    create_engine_params: Optional[dict] = field(factory=dict, kw_only=True)
+    _Model: any = field(default=vector_model_factory, init=False)
+    engine: Optional[Engine] = field(default=None)
+
+    def initialize(
```
vasinov, line 44: Can we rename this (may be to `setup`?), so it doesn't sound too similar to `__init__`?

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,170 @@
...
+from griptape.drivers import BaseVectorStoreDriver
+from sqlalchemy.engine import Engine
+from sqlalchemy import create_engine, Column, String, JSON
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.orm import Session
+from pgvector.sqlalchemy import Vector
+
+
+def vector_model_factory(table_name="griptape_vectors"):
```
vasinov, line 14: Can you move this inside `PgVectorVectorStoreDriver`? Griptape convention is to keep all static methods as class methods.

vasinov: Can you also add type hints for `table_name` and the return type?

vasinov: Do we need a default here?

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,170 @@
...
+
+    Attributes:
+        connection_string: The connection string for the Postgres database instance.
+        create_engine_params: Additional configuration params passed when creating the database connection.
+        engine: An optional sqlalchemy Postgres engine to use.
+    """
+
+    connection_string: str = field(default=None, kw_only=True)
+    create_engine_params: Optional[dict] = field(factory=dict, kw_only=True)
+    _Model: any = field(default=vector_model_factory, init=False)
```
vasinov, line 41: Should be `_model`

vasinov: `default` should be `Factory(lambda: vector_model_factory)`. Read more on why [here](https://stackoverflow.com/questions/44141104/when-and-why-should-i-use-attr-factory).

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,170 @@
...
+    engine: Optional[Engine] = field(default=None)
```
vasinov, line 42: Add `kw_only=True`.

vasinov: What happens when the engine is `None? Should it be automatically instantiated based on the `connection_string`?

griptape/drivers/vector/pgvector_vector_store_driver.py
vasinov: Add type hints to all props and the return type.

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def initialize(
+        self,
+        table_name="griptape_vectors",
```
vasinov, line 46: Does this need a default?

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,170 @@
...
+        count: Optional[int] = BaseVectorStoreDriver.DEFAULT_QUERY_COUNT,
+        namespace: Optional[str] = None,
+        include_vectors: bool = False,
+        distance_metric: str = "cosine_distance",
+        **kwargs
+    ) -> list[BaseVectorStoreDriver.QueryResult]:
+        """Performs a search on the collection to find vectors similar to the provided input vector,
+        optionally filtering to only those that match the provided namespace.
+        """
+        distance_metrics = {
```
vasinov, line 137: Any way to move those to a constant at the top of the class? Should be possible if those be accessed as class properties of `VectorModel` directly.

griptape/drivers/vector/pgvector_vector_store_driver.py
vasinov: But we can't expect `setup` to be called every time the driver is used.

griptape/drivers/vector/pgvector_vector_store_driver.py
andrewfrench: Thanks for the link, important subtlety there.

griptape/drivers/vector/pgvector_vector_store_driver.py
andrewfrench: Not sure if this is doable without lots of magic as the dict values depend on the dynamic `_model` being built. The comparison methods themselves are returned by a factory inside `sqlalchemy.Vector` and must be instantiated for this to work according to my experimentation.

griptape/drivers/vector/pgvector_vector_store_driver.py
andrewfrench: Gotcha. I've moved the absolutely required validation/initialization stuff to  `__attrs_post_init__()` in the pattern of many of our other drivers.

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,175 @@
...
+        connection_string: An optional string describing the target Postgres database instance.
+        create_engine_params: Additional configuration params passed when creating the database connection.
+        engine: An optional sqlalchemy Postgres engine to use.
+        table_name: Optionally specify the name of the table to used to store vectors.
+    """
+
+    connection_string: str = field(default=None, kw_only=True)
+    create_engine_params: Optional[dict] = field(factory=dict, kw_only=True)
+    engine: Optional[Engine] = field(default=None, kw_only=True)
+    table_name: str = field(default="griptape_vectors", kw_only=True)
```
vasinov, line 28: Do we need a default name here? I'd argue against it.

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Attributes:
+        connection_string: An optional string describing the target Postgres database instance.
+        create_engine_params: Additional configuration params passed when creating the database connection.
+        engine: An optional sqlalchemy Postgres engine to use.
+        table_name: Optionally specify the name of the table to used to store vectors.
+    """
+
+    connection_string: str = field(default=None, kw_only=True)
+    create_engine_params: Optional[dict] = field(factory=dict, kw_only=True)
```
vasinov, line 26: Shouldn't be `Optional` since we are providing a concrete empty default here.

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,175 @@
...
+    """A vector store driver to Postgres using the PGVector extension.
+
+    Attributes:
+        connection_string: An optional string describing the target Postgres database instance.
+        create_engine_params: Additional configuration params passed when creating the database connection.
+        engine: An optional sqlalchemy Postgres engine to use.
+        table_name: Optionally specify the name of the table to used to store vectors.
+    """
+
+    connection_string: str = field(default=None, kw_only=True)
```
vasinov, line 25: This should be `Optional[str]` since we are providing `None` as the default.

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,175 @@
...
+                    vector=result[0].vector if include_vectors else None,
+                    score=result[1],
+                    meta=result[0].meta,
+                    namespace=result[0].namespace,
+                )
+                for result in results
+            ]
+
+    @classmethod
+    def vector_model_factory(cls, table_name: str) -> any:
```
vasinov, line 163: Can we rename it to `default_vector_model` to match our naming convention?

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,175 @@
...
+        create_engine_params: Additional configuration params passed when creating the database connection.
+        engine: An optional sqlalchemy Postgres engine to use.
+        table_name: Optionally specify the name of the table to used to store vectors.
+    """
+
+    connection_string: str = field(default=None, kw_only=True)
+    create_engine_params: Optional[dict] = field(factory=dict, kw_only=True)
+    engine: Optional[Engine] = field(default=None, kw_only=True)
+    table_name: str = field(default="griptape_vectors", kw_only=True)
+    _model: any = field(default=Factory(lambda self: self.vector_model_factory, takes_self=True), init=False)
```
vasinov, line 29: Doesn't need `takes_self` here.

vasinov: Why `init=False`?

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __attrs_post_init__(self) -> None:
+        """Validates engine and connection string inputs.
+        If a an engine is provided, it will be used to connect to the database. If not, a connection string
+        is used to create a new database connection.
+        """
+        if self.engine is None and self.connection_string is None:
```
vasinov, line 36: Can we move this into two validators for `engine` and `connection_string` to make it more explicit?

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,175 @@
...
+            raise ValueError("Either an engine or connection string must be provided")
+
+        if self.engine is None:
+            self.engine = create_engine(self.connection_string, **self.create_engine_params)
+
+        self._model = self.vector_model_factory(self.table_name)
```
vasinov, line 42: Don't need this line since you are setting things in the `Factory` now.

vasinov (changes requested):

griptape/drivers/vector/pgvector_vector_store_driver.py
andrewfrench: I've removed `init=False`, but I think we do need `takes_self` as this is a class method and we lose the reference to `self` when it's removed.

collindutter (dismissed):

trish11953 (dismissed):

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -0,0 +1,192 @@
...
+
+        # If a connection string is not provided, an engine is required.
+        if engine is None:
+            raise ValueError("An engine or connection string is required")
+
+    def __attrs_post_init__(self) -> None:
+        """Validates engine and connection string inputs.
+        If a an engine is provided, it will be used to connect to the database. If not, a connection string
+        is used to create a new database connection.
+        """
```
vasinov, line 58: Update doc.

vasinov (changes requested):

griptape/drivers/vector/pgvector_vector_store_driver.py
vasinov: Yep, you are right. Sorry for the confusion!

vasinov (approved):

PR #316, opened 2023-10-03
BaseGoogleClient: Updated with reused functions

trish11953:


griptape/tools/base_google_client.py
```diff
@@ -1,7 +1,52 @@
...
+            self.service_account_credentials, scopes=scopes
+        )
+
+        return build(
+            serviceName=service_name,
+            version=version,
+            credentials=credentials.with_subject(owner_email)
+        )
+
+    def _path_to_file_id(self, service, path: str) -> Optional[str]:
```
vasinov, line 25: Missing type hint for `service`.

vasinov: Can we rename it to `_convert_path_to_file_id`? Otherwise, it feels like this method returns a "path" to some "file ID," which doesn't make sense :)

griptape/tools/google_drive/tool.py
```diff
@@ -45,7 +45,7 @@ def list_files(self, params: dict) -> ListArtifact | ErrorArtifact:
         folder_path = values.get("folder_path", self.DEFAULT_FOLDER_PATH)
 
         try:
-            service = self._build_client(self.LIST_FILES_SCOPES)
+            service = self._build_client(self.LIST_FILES_SCOPES, "drive", "v3", self.owner_email)
```
vasinov, line 48: Let's move those into `SERVICE_NAME`, and `SERVICE_VERSION` since they are strings repeated in multiple places in this class.

vasinov (changes requested):

griptape/tools/base_google_client.py
trish11953: Makes sense, changed it !

griptape/tools/google_drive/tool.py
trish11953: Done!

vasinov (approved):

Issue #255, opened 2023-09-20
Agents always require setting prompt_driver with very low max_tokens


nulldiver:
**Describe the bug**
Without setting an explicit prompt_driver with very low tokens per minute, nearly all requests immediately hit a rate limit of 1000 tokens / minute. This is with no other activity beyond a simple agent request occurring on the account.

With a prompt_driver set, with a low max_tokens, eg:

    prompt_driver=OpenAiChatPromptDriver(
        model="gpt-3.5-turbo",
        max_tokens=200
    )
    
Things work as expected. With a larger max_tokens (eg. the defaults set for the models in Griptape):

    prompt_driver=OpenAiChatPromptDriver(
        model="gpt-3.5-turbo",
        max_tokens=4096
    ),

I get:
InvalidRequestError This model's maximum context length is 4097 tokens. However, you requested 4208 tokens.

**To Reproduce**
This has been the only behavior that I have seen while using Griptape.
This is not a behavior that I have observed directly working with OpenAI's python APIs.

**Desktop (please complete the following information):**
 - macOS 13.5
 - python 3.11
 - griptape 0.17.1

collindutter:
Yep, I'm able to reproduce with this:
```python
from griptape.drivers import OpenAiChatPromptDriver
from griptape.structures import Agent


agent = Agent(
    prompt_driver=OpenAiChatPromptDriver(model="gpt-3.5-turbo", max_tokens=4096),
)

agent.run("Hi")
```
```
openai.error.InvalidRequestError: This model's maximum context length is 4097 tokens. However, you requested 4116 tokens (20 in the messages, 4096 in the completion). Please reduce the length of the messages or completion.
```

@vasinov why do we override [max_output_tokens](https://github.com/griptape-ai/griptape/blob/main/griptape/drivers/prompt/openai_chat_prompt_driver.py#L54)? Seems like we'd want the `min` behavior from the [parent implementation](https://github.com/griptape-ai/griptape/blob/main/griptape/drivers/prompt/base_prompt_driver.py#L31).

nulldiver:
This didn't actually totally fix the issue. It removes the maximum context length error. But to clarify the overall problem that I'm seeing:

This works:

```
agent = Agent(
    prompt_driver=OpenAiChatPromptDriver(model="gpt-3.5-turbo"),
)
```

But this:

```
agent = Agent(
    prompt_driver=OpenAiChatPromptDriver(model="gpt-4"),
)
```

Or this:

```
[...]

PR #333, opened 2023-10-05
Remove max_tokens from OpenAiPromptDriver

andrewfrench:
#255

### DIFF
--- a/griptape/drivers/prompt/openai_chat_prompt_driver.py
+++ b/griptape/drivers/prompt/openai_chat_prompt_driver.py
@@ -7,6 +7,7 @@ from griptape.utils import PromptStack
 from griptape.drivers import BasePromptDriver
 from griptape.tokenizers import OpenAiTokenizer
 from typing import Tuple, Type
+import requests
 
 
 @define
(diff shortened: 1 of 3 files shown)


collindutter:
This is awesome, hopefully this fixes the gpt-4 throttling issues for good. We probably still want to give users the ability to set `max_tokens` if they choose though. Maybe only call `self.max_output_tokens(messages)` and insert `max_tokens` into the payload if the user explicit provides a value for the Prompt Driver `max_tokens`?

andrewfrench:
Yeah, the user should still be able to override. Converted this back to draft until I can confirm that functionality.

collindutter:
Maybe we should investigate this change for the other LLM providers? If they don't require sending `max_tokens`, maybe it's not worth sending by default? It would simplify scenarios where the LLM provider doesn't provide a tokenizer at all.

vasinov:
> Maybe we should investigate this change for the other LLM providers? If they don't require sending `max_tokens`, maybe it's not worth sending by default? It would simplify scenarios where the LLM provider doesn't provide a tokenizer at all.

I agree. Let's create a separate issue for that and try to merge this one quickly.

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -76,6 +76,13 @@ def _base_params(self, prompt_stack: PromptStack) -> dict:
             "messages": messages
         }
 
+        # A max_tokens parameter is not required, but if it is specified by the caller, bound it to
+        # the maximum value as determined by the tokenizer and pass it to the API.
+        if self.max_tokens:
+            params["max_tokens"] = self.max_output_tokens(messages)
```
vasinov, line 82: I think this is fine but could potentially be confusing when the user sets tokens explicitly and we still sometimes override it with fewer tokens. Let's keep it this way, as this feels like a better UX, but may be document it elsewhere.

vasinov (approved):

PR #334, opened 2023-10-05
Add support for Ollama, Palm, Claude-2, Cohere, Replicate Llama2, CodeLlama, Hugging Face (100+LLMs) - using LiteLLM

ishaan-jaff:
This PR adds support for the above mentioned LLMs using LiteLLM https://github.com/BerriAI/litellm/ 
LiteLLM is a lightweight package to simplify LLM API calls - use any llm as a drop in replacement for gpt-3.5-turbo. 

Example
```python
from litellm import completion

## set ENV variables
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion(model="command-nightly", messages)

# anthropic call
response = completion(model="claude-instant-1", messages=messages)
```


ishaan-jaff:
I believe we add Palm, Vertex, Ollama, VLLM, DeepInfra, Ai21, TogetherAI, Petals, OpenRouter, Replicate, NLP Cloud to griptape 

I looked at https://docs.griptape.ai/griptape-framework/structures/prompt-drivers/

See LiteLLM providers here: https://docs.litellm.ai/docs/providers

ishaan-jaff:
@vasinov @collindutter @trish11953 can I get a review on this PR ?

vasinov:
This change doesn't make sense in the context of the Griptape framework: `OpenAiPromptDriver` should use the OpenAI SDK. There is also more nuance to supporting different models in Griptape due to how we integrate tokenizers and prompt generation logic.

Also, looks like this PR is part of a focused (automated?) effort to promote LiteLLM.

ishaan-jaff:
@vasinov I'm doing it manually and thought this would create value for griptape


> OpenAiPromptDriver should use the OpenAI SDK.

Why not have one class for all LLM I/O, do you want to maintain all your LLM integrations on griptape ? 

> There is also more nuance to supporting different models in Griptape due to how we integrate tokenizers and prompt generation logic.

Was litellm missing something ? any feedback here would be useful

PR #339, opened 2023-10-06
Update bedrock default models, remove model from prompt model

collindutter:


griptape/drivers/prompt_model/bedrock_claude_prompt_model_driver.py
```diff
@@ -1,23 +1,41 @@
...
+
+        See this thread more more information: https://github.com/griptape-ai/griptape/issues/244
+
+        Returns:
+            BedrockClaudeTokenizer: The tokenizer for this driver.
+        """
+        if self._tokenizer:
+            return self._tokenizer
+        else:
+            if isinstance(self.prompt_driver, AmazonBedrockPromptDriver):
```
vasinov, line 34: To make it more explicit, let's override the prompt driver in this class with:

```python
```

And let's drop the type check here. We are not doing it anywhere else in the framework, as we are relying on type hints for that.

griptape/drivers/prompt_model/bedrock_jurassic_prompt_model_driver.py
```diff
@@ -33,7 +32,7 @@ def tokenizer(self) -> BedrockJurassicTokenizer:
             return self._tokenizer
         else:
             if isinstance(self.prompt_driver, AmazonBedrockPromptDriver):
```
vasinov, line 34: Same here.

griptape/drivers/prompt_model/bedrock_titan_prompt_model_driver.py
```diff
@@ -33,7 +32,7 @@ def tokenizer(self) -> BedrockTitanTokenizer:
             return self._tokenizer
         else:
             if isinstance(self.prompt_driver, AmazonBedrockPromptDriver):
```
vasinov, line 34: And here.

vasinov (changes requested):

vasinov (approved):

PR #343, opened 2023-10-09
Prune conversation memory

### LINKED ISSUE #172: Add optional `PromptStack` conversation pruning to prompt drivers
Before `PromptStack` was introduced in `v0.15.0`, conversation memory supported automatic pruning. Since then, prompt construction logic was moved to prompt drivers and optional conversation pruning was removed.

I propose adding back support for optional conversation pruning to prompt drivers.

collindutter:
Fixes #172

### DIFF
--- a/griptape/memory/structure/conversation_memory.py
+++ b/griptape/memory/structure/conversation_memory.py
@@ -3,11 +3,12 @@ import json
 from typing import TYPE_CHECKING, Optional
 from attr import define, field, Factory
 from griptape.memory.structure import Run
+from griptape.utils import PromptStack
+from griptape.utils import Conversation
 
 if TYPE_CHECKING:
     from griptape.drivers import BaseConversationMemoryDriver
     from griptape.structures import Structure
-    from griptape.utils import PromptStack
 
 
 @define
@@ -17,6 +18,7 @@ class ConversationMemory:
     runs: list[Run] = field(factory=list, kw_only=True)
     structure: Structure = field(init=False)
     autoload: bool = field(default=True, kw_only=True)
+    autoprune: bool = field(default=True, kw_only=True)
 
     def __attrs_post_init__(self) -> None:
         if self.driver and self.autoload:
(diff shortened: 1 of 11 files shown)


griptape/tasks/toolkit_task.py
collindutter, line 1: I don't love the non-DRYness here, but this new implementation assumes that the stack has been fully built before adding in memory. So we can't use the parent class's `prompt_stack` because it adds memory before the `subtasks` are added.

griptape/memory/structure/conversation_memory.py
```diff
@@ -17,17 +18,60 @@ class ConversationMemory:
...
-            stack.add_assistant_input(r.output)
+    def add_to_prompt_stack(self, stack: PromptStack, index: Optional[int]=None) -> None:
+        """Add the Conversation Memory runs to the Prompt Stack.
+
+        If autoprune is enabled, this will fit as many Conversation Memory runs into the Prompt Stack
+        as possible without exceeding the token limit.
+
+        Args:
+            stack: The Prompt Stack to add the Conversation Memory runs to.
+            index: Optional index to insert the Monversation Memory runs at.
```
vasinov, line 37: *Conversation

griptape/tasks/prompt_task.py
```diff
@@ -67,14 +67,14 @@ def prompt_stack(self) -> PromptStack:
...
-        if memory:
-            memory.add_to_prompt_stack(stack)
-
         stack.add_user_input(self.input.to_text())
 
         if self.output:
             stack.add_assistant_input(self.output.to_text())
 
+        if memory:
+            memory.add_to_prompt_stack(stack, 1)
```
vasinov, line 76: May be add a comment explaining that this injects memory after the system input (`stack.inputs[index:index]`), hence the index of `1` here.

griptape/tasks/toolkit_task.py
```diff
@@ -63,13 +63,25 @@ def memory(self) -> list[BaseToolMemory]:
 
     @property
     def prompt_stack(self) -> PromptStack:
```
vasinov, line 65: Implementation totally makes sense but made me think about how we are at some point bound to "prune" during CoT execution but then, once it's done, some memory will get "unpruned." Nothing we can do about it but can potentially affect LLM reasoning.

griptape/memory/structure/conversation_memory.py
```diff
@@ -17,17 +18,60 @@ class ConversationMemory:
...
         if self.driver and self.autoload:
             memory = self.driver.load()
             if memory is not None:
                 [self.add_run(r) for r in memory.runs]
 
-    def add_to_prompt_stack(self, stack: PromptStack) -> None:
-        for r in self.runs:
-            stack.add_user_input(r.input)
-            stack.add_assistant_input(r.output)
+    def add_to_prompt_stack(self, stack: PromptStack, index: Optional[int]=None) -> None:
```
vasinov, line 29: I'm wondering if this should live in `PromptStack.add_memory(memory, index)` instead. This way the object (prompt stack) is modifying itself instead of another entity (memory) modifying the object. The former is generally less bug-prone. Let me know if that breaks anything though.

vasinov (changes requested):

griptape/tasks/toolkit_task.py
collindutter: I think this is where some sort of limitless vector based memory could come in handy.

griptape/memory/structure/conversation_memory.py
collindutter: I agree, and it does make for a nicer interface. Though it gets kind of ugly with other memory types like `SummaryConversationMemory` which adds `summary` and `unsummarized_runs` to the stack.

And typing that comment made me realize I did not implement pruning logic for `SummaryConversationMemory` 😅

collindutter: Made this refactor, it's much cleaner.

vasinov (approved):

PR #349, opened 2023-10-09
Add buffering to ConversationMemory

### LINKED ISSUE #344: Move functionality from BufferConversationMemory into ConversationMemory
The buffering functionality from `BufferConversationMemory` should be moved to regular `ConversationMemory` so that subclasses like `SummaryConversationMemory` can also leverage it.

collindutter:
Fixes #344


griptape/memory/structure/conversation_memory.py
```diff
@@ -17,6 +17,7 @@ class ConversationMemory:
     runs: list[Run] = field(factory=list, kw_only=True)
     structure: Structure = field(init=False)
     autoload: bool = field(default=True, kw_only=True)
+    buffer_size: Optional[int] = field(default=None, kw_only=True)
```
vasinov, line 20: Should be also added to `ConversationMemorySchema`.

vasinov: You can also remove it from:
1. `BufferConversationMemory`.
2. `BufferConversationMemorySchema`.

vasinov: And, finally, `SummaryConversationMemory.offset` can be swapped out as well.

vasinov (changes requested):

griptape/memory/structure/conversation_memory.py
```diff
@@ -17,6 +17,7 @@ class ConversationMemory:
     runs: list[Run] = field(factory=list, kw_only=True)
     structure: Structure = field(init=False)
     autoload: bool = field(default=True, kw_only=True)
+    buffer_size: int = field(default=0, kw_only=True)
```
vasinov, line 20: I think buffer size of zero should imply that buffer is literally zero, so memory is empty. Let's make it an optional and set the default to `None` instead.

vasinov: Does it make sense to swap out `SummaryConversationMemory.offset` for `buffer_size` to minimize the number of variables we have for memory?

vasinov (changes requested):

griptape/memory/structure/conversation_memory.py
collindutter: @vasinov I set it to `0` because of serialization issues with `None`. But I agree with your point, let me see if I can work around it.

collindutter: Added marshmallow hook to filter out `None`.

griptape/memory/structure/conversation_memory.py
vasinov: 3. Makes sense. Let's rename `buffer_size` to `max_memory_runs` and leave `offset`. Unless we want to rename `offset` into something more meaningful.

griptape/schemas/memory/conversation_memory_schema.py
```diff
@@ -8,6 +8,7 @@ class Meta:
 
     type = fields.Str()
     runs = fields.List(fields.Nested(RunSchema()))
+    buffer_size = fields.Int()
```
vasinov, line 11: You can just use `fields.Int(allow_none=True)` with Marshmallow. No need for manual `None` handling.

vasinov (changes requested):

vasinov (changes requested):

vasinov (approved):

PR #353, opened 2023-10-10
GoogleGmailClient and EmailClient: Optional attachment param

### LINKED ISSUE #352: GoogleGmailClient and EmailClient: "attachment support"
The tools should be able to attach a file to emails. Both `GoogleGmailClient` and `EmailClient` need it. 

**Current scenario**
 It's currently impossible to implement until memory is normalized and we can pipe list artifact tool outputs into other tools as one object instead of a list of objects.

trish11953:
Resolves #352


griptape/tools/email_client/tool.py
```diff
@@ -160,27 +167,43 @@ def send(self, params: dict) -> InfoArtifact | ErrorArtifact:
...
+                    part = MIMEBase("application", "octet-stream")
+                    part.set_payload(attachment.read())
+                    encoders.encode_base64(part)
+                    part.add_header(
+                        "Content-Disposition",
+                        f"attachment; filename= {file_path.split('/')[-1]}"
+                    )
+                    msg.attach(part)
+            except Exception as e:
+                logging.error(f"Error attaching {file_path}: {e}")
```
vasinov, line 193: Should we fail the whole activity if any of the files can't be attached? Not actually sure what the right behavior should be but leaning towards returning an `ErrorArtifact` here.

griptape/tools/email_client/tool.py
```diff
@@ -160,27 +167,43 @@ def send(self, params: dict) -> InfoArtifact | ErrorArtifact:
         smtp_host = self.smtp_host
         smtp_port = int(self.smtp_port)
 
-        to_email = input_values["to"]
-        subject = input_values["subject"]
```
vasinov, line 164: Why remove those? You have them in `GoogleGmailClient`.

griptape/tools/google_gmail/tool.py
vasinov, line 1: Can you rename `google_gmail` to `google_gmail_client` for consistency?

griptape/tools/email_client/tool.py
```diff
@@ -140,7 +141,13 @@ def retrieve(self, params: dict) -> ListArtifact | ErrorArtifact:
             Literal(
                 "body",
                 description="Email body"
-            ): str
+            ): str,
+            Optional(
+                Literal(
```
vasinov, line 146: Can collapse into a single `Optional`.

griptape/tools/email_client/tool.py
```diff
@@ -140,7 +141,13 @@ def retrieve(self, params: dict) -> ListArtifact | ErrorArtifact:
...
+                    "attachments",
+                    description="List of file paths to be attached to the email"
+                )
+            ): list[str]
```
vasinov, line 150: Do we keep this as a list paths? Would we want to load attachments from memory or should we recommend using this tool in conjunction with the `FileManager` tool that can store attachments in a separate action? Just something to think about for now.

vasinov (changes requested):

griptape/tools/email_client/tool.py
trish11953: Done!

griptape/tools/email_client/tool.py
trish11953: That's true, I just thought it's easier if this is done using one tool. And since it's just the EmailClient and GoogleGmailClient which we know needs attachments most of the time, we don't really need an extra tool (FileManager) to do this. Kind of easier for the user to have the functionality by calling the same tool.

griptape/tools/email_client/tool.py
trish11953: just thought its cleaner since im using it here 

```
msg["To"] = input_values["to"]
msg["Subject"] = input_values["subject"]
```

griptape/tools/email_client/tool.py
trish11953: Done!

griptape/tools/google_gmail/tool.py
trish11953: Sure!

griptape/tools/email_client/tool.py
trish11953: But for consistency, will put it back!

griptape/tools/email_client/tool.py
vasinov: How would you, for example, attach a summary of a website to an email without `FileManager`?

griptape/tools/email_client/tool.py
vasinov: Ha, I wasn't sure if that's the right thing to do. What do you think?

@collindutter, @andrewfrench?

vasinov (changes requested):

griptape/tools/email_client/tool.py
trish11953: I would do that using the WebScraper tool to give a summary of a website and then send that as the content of the email. Why would you want that as an attachment though? Attaching a summary of a website to an email, as an actual attachment doesn't seem like a common use case of the attachment support according to me. If I was using `GoogleGmailClient` or `EmailClient`, I would simply want it to attach files for me from my local. I could be wrong though about the use case. You would know better.

vasinov: The website summary was just an example :)

The broader point is that this activity should be able to pull things from memory. If you have `FileManager` available, then the LLM can always load a file into memory and then send it as an attachment. And it can use any other tool to load up data before attaching it (including a website summary! :) ).

vasinov (changes requested):

griptape/tools/email_client/tool.py
trish11953: Ok got it! Let me try to code it in a way we can use FileManager to do this.

griptape/tools/email_client/tool.py
```diff
@@ -139,51 +140,83 @@ def retrieve(self, params: dict) -> ListArtifact | ErrorArtifact:
...
+                if list_artifact:
+                    for artifact, attachment_name in zip(list_artifact.value, attachment_names):
+                        file_data = artifact.value.encode()
+                        part = MIMEBase("application", "octet-stream")
+                        part.set_payload(file_data)
+                        encoders.encode_base64(part)
+                        part.add_header("Content-Disposition", f"attachment; filename={attachment_name}")
+                        msg.attach(part)
+                else:
+                    return ErrorArtifact(f"Artifact with namespace {artifact_namespace} not found.")
```
collindutter, line 200: Nesting gets pretty deep here, can we refactor this? Perhaps split out into some separate functions?

griptape/tools/google_gmail_client/tool.py
```diff
@@ -44,10 +60,27 @@ def create_draft_email(self, params: dict) -> InfoArtifact | ErrorArtifact:
...
+                    if list_artifacts:
+                        for artifact, attachment_name in zip(list_artifacts.value, attachment_names):
+                            file_data = artifact.value.encode()
+                            message.add_attachment(file_data, maintype='application',
+                                                   subtype='octet-stream', filename=attachment_name)
+                    else:
+                        logging.error(f"Artifact with namespace {artifact_namespace} not found.")
+                        return ErrorArtifact(f"Artifact with namespace {artifact_namespace} not found.")
+                else:
+                    return ErrorArtifact(f"memory not found.")
```
collindutter, line 83: Similar logic/nesting here. What do you think about generalizing the logic and moving to a util function? Maybe in `BaseTool`?

collindutter (changes requested):

griptape/tools/email_client/tool.py
collindutter: Sorry very late to this comment 🙃. I don't know if this should always be the case, but I think here it makes sense to return an Error Artifact if any fail to attach. If I wanted to send an email with important documents and only some of them got attached, I would consider that a failure.

griptape/tools/email_client/tool.py
trish11953: made a few protected methods

griptape/tools/google_gmail_client/tool.py
trish11953: I don't think it needs to be in the BaseTool since the logic doesn't get used by any other tool. But protected methods seems better to me.

griptape/tools/email_client/tool.py
vasinov: I agree. Let's fail the whole activity in this case.

griptape/tools/google_gmail_client/tool.py
```diff
@@ -0,0 +1,97 @@
...
+                description="subject of the email"
+            ): str,
+            Literal(
+                "body",
+                description="body of the email"
+            ): str,
+            Optional(
+                "attachment_names",
+                description="Names of the attachments"
+            ): list[str],
```
vasinov, line 36: I don't think this works. Needs to be a list of attachment objects with the following properties:
- attachment_name
- memory_name
- artifact_namespace

griptape/tools/google_gmail_client/tool.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _attach_files_to_message(self, message: EmailMessage, values: dict) -> None:
+        attachment_names = values.get("attachment_names")
+        memory_name = values.get("memory_name")
+        artifact_namespace = values.get("artifact_namespace")
+    
+        if memory_name and artifact_namespace:
+            memory = self.find_input_memory(memory_name)
+            if memory:
+                list_artifacts = memory.load_artifacts(artifact_namespace)
```
vasinov, line 87: This returns a single `ListArtifact`.

griptape/tools/google_gmail_client/tool.py
```diff
@@ -0,0 +1,97 @@
...
+                if list_artifacts:
+                    for artifact, attachment_name in zip(list_artifacts.value, attachment_names):
```
vasinov, line 89: This is the reason your approach doesn't work: you are making an assumption here that attachment names somehow map to values from the `ListArtifact`, which will almost never be true.

vasinov (changes requested):

griptape/tools/google_gmail_client/tool.py
trish11953: Yes correct, the variable name should be changed

griptape/tools/google_gmail_client/tool.py
trish11953: It works actually

griptape/tools/google_gmail_client/tool.py
trish11953: It works!!

Issue #309, opened 2023-10-02
Evaluate solution for token rate limiting (e.g. query remaining tokens)


derekpai:

derekpai:
refer to https://platform.openai.com/docs/guides/rate-limits/overview

andrewfrench:
Closing as completed in #333

PR #354, opened 2023-10-11
Extract rate limit metadata from OpenAI API response headers

andrewfrench:
Adds dormant functionality to parse OpenAI API response headers for ratelimit-related information. In the future, we can use this information to more intelligently implement backoff or calculate token requests.

New tests ensure functionality of header parsing functionality.

Related to #309.

### DIFF
--- a/griptape/drivers/prompt/openai_chat_prompt_driver.py
+++ b/griptape/drivers/prompt/openai_chat_prompt_driver.py
@@ -140,9 +140,9 @@ class OpenAiChatPromptDriver(BasePromptDriver):
             return "user"
 
     def _extract_ratelimit_metadata(self, response, *args, **kwargs):
-        # The timeparse utility doesn't handle sub-second durations as are sometimes returned by OpenAI's API.
-        # If the API returns, for example, "13ms", timeparse returns None. In this case, we will set the time value
-        # to the current time plus a one second buffer.
+        # The dateparser utility doesn't handle sub-second durations as are sometimes returned by OpenAI's API.
+        # If the API returns, for example, "13ms", dateparser.parse() returns None. In this case, we will set
+        # the time value to the current time plus a one second buffer.
         self.__ratelimit_requests_reset_at = dateparser.parse(
             response.headers["x-ratelimit-reset-requests"],
             settings={"PREFER_DATES_FROM": "future"},
--- a/griptape/drivers/prompt/openai_chat_prompt_driver.py
+++ b/griptape/drivers/prompt/openai_chat_prompt_driver.py
@@ -26,6 +26,12 @@ class OpenAiChatPromptDriver(BasePromptDriver):
         organization: OpenAI organization.
         tokenizer: Custom `OpenAiTokenizer`.
         user: OpenAI user.
+        _ratelimit_request_limit: The maximum number of requests allowed in the current rate limit window.
+        _ratelimit_requests_remaining: The number of requests remaining in the current rate limit window.
+        _ratelimit_requests_reset_at: The time at which the current rate limit window resets.
+        _ratelimit_token_limit: The maximum number of tokens allowed in the current rate limit window.
+        _ratelimit_tokens_remaining: The number of tokens remaining in the current rate limit window.
+        _ratelimit_tokens_reset_at: The time at which the current rate limit window resets.
     """
 
     api_type: str = field(default=openai.api_type, kw_only=True)
--- a/griptape/drivers/prompt/openai_chat_prompt_driver.py
+++ b/griptape/drivers/prompt/openai_chat_prompt_driver.py
@@ -41,12 +41,12 @@ class OpenAiChatPromptDriver(BasePromptDriver):
     ignored_exception_types: Tuple[Type[Exception], ...] = field(
         default=Factory(lambda: openai.InvalidRequestError), kw_only=True
     )
-    _ratelimit_request_limit: Optional[int] = field(init=False, default=None)
-    _ratelimit_requests_remaining: Optional[int] = field(init=False, default=None)
-    _ratelimit_requests_reset_at: Optional[datetime] = field(init=False, default=None)
-    _ratelimit_token_limit: Optional[int] = field(init=False, default=None)
-    _ratelimit_tokens_remaining: Optional[int] = field(init=False, default=None)
-    _ratelimit_tokens_reset_at: Optional[datetime] = field(init=False, default=None)
+    __ratelimit_request_limit: Optional[int] = field(init=False, default=None)
+    __ratelimit_requests_remaining: Optional[int] = field(init=False, default=None)
+    __ratelimit_requests_reset_at: Optional[datetime] = field(init=False, default=None)
+    __ratelimit_token_limit: Optional[int] = field(init=False, default=None)
+    __ratelimit_tokens_remaining: Optional[int] = field(init=False, default=None)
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -52,6 +52,7 @@ pandas = "^2.1.0"
 pgvector = "^0.2.3"
 psycopg2-binary = "^2.9.9"
 snowflake-connector-python = "^3.2"
+pytimeparse = "^1.1.8"
```
collindutter, line 55: Let's instead use [dateparser](https://dateparser.readthedocs.io/en/latest/) instead as we're already using it one of the tools.

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -90,3 +138,22 @@ def __to_openai_role(self, prompt_input: PromptStack.Input) -> str:
             return "assistant"
         else:
             return "user"
+
+    def _extract_ratelimit_metadata(self, response, *args, **kwargs):
+        # The timeparse utility doesn't handle sub-second durations as are sometimes returned by OpenAI's API.
+        # If the API returns, for example, "13ms", timeparse returns None. In this case, we will set the time value
+        # to the current time plus a one second buffer.
+        reset_requests_duration_sec = timeparse(response.headers["x-ratelimit-reset-requests"])
```
collindutter, line 146: Replace with:
`dateparser.parse(response.headers["x-ratelimit-reset-requests"], settings={'PREFER_DATES_FROM': 'future'})` 

This has the similar problem in that it will return None for something small like `13ms`.

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -7,58 +7,106 @@
...
-    ignored_exception_types: Tuple[Type[Exception], ...] = field(default=Factory(lambda: (openai.InvalidRequestError)), kw_only=True)
+    ignored_exception_types: Tuple[Type[Exception], ...] = field(
+        default=Factory(lambda: openai.InvalidRequestError), kw_only=True
+    )
+    __ratelimit_request_limit: Optional[int] = field(default=None)
+    __ratelimit_requests_remaining: Optional[int] = field(default=None)
+    __ratelimit_requests_reset_at: Optional[datetime] = field(default=None)
+    __ratelimit_token_limit: Optional[int] = field(default=None)
+    __ratelimit_tokens_remaining: Optional[int] = field(default=None)
+    __ratelimit_tokens_reset_at: Optional[datetime] = field(default=None)
```
collindutter, line 49: I think we can accomplish this private pattern by using `init=False` as described [here](https://www.attrs.org/en/stable/examples.html#basics).

i.e. `_ratelimit_request_limit: Optional[int] = field(init=False, default=None)`

collindutter (changes requested):

pyproject.toml
```diff
@@ -52,7 +52,6 @@ pandas = "^2.1.0"
 pgvector = "^0.2.3"
 psycopg2-binary = "^2.9.9"
 snowflake-connector-python = "^3.2"
-pytimeparse = "^1.1.8"
```
collindutter: Sorry, should've clarified in the other comment. `dateparser` is being used by a tool which means the dependencies are being loaded at runtime from the tool's `requirements.txt`. To use it in the framework we _also_ need to install it here.

collindutter: If the unit tests pass, it is because the tests for the `DateTime` tool happened to run first, installing the package.

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -7,58 +7,106 @@
...
             raise Exception("Completion with more than one choice is not supported yet.")
 
     def token_count(self, prompt_stack: PromptStack) -> int:
-        return self.tokenizer.token_count(
-            self._prompt_stack_to_messages(prompt_stack)
-        )
+        return self.tokenizer.token_count(self._prompt_stack_to_messages(prompt_stack))
+
+    @property
+    def ratelimit_request_limit(self) -> Optional[int]:
```
collindutter, line 67: I think we can get rid of these properties now that the fields are not initialized up top. The purpose of this was to expose readonly public properties of private fields, right?

pyproject.toml
andrewfrench: Gets me every time. Do you recommend a way to test that dependencies are correctly defined or is this muscle memory at this point?

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -7,58 +7,106 @@
...
-        default=Factory(lambda self: OpenAiTokenizer(model=self.model), takes_self=True),
-        kw_only=True
+        default=Factory(lambda self: OpenAiTokenizer(model=self.model), takes_self=True), kw_only=True
     )
     user: str = field(default="", kw_only=True)
-    ignored_exception_types: Tuple[Type[Exception], ...] = field(default=Factory(lambda: (openai.InvalidRequestError)), kw_only=True)
+    ignored_exception_types: Tuple[Type[Exception], ...] = field(
+        default=Factory(lambda: openai.InvalidRequestError), kw_only=True
+    )
+    __ratelimit_request_limit: Optional[int] = field(init=False, default=None)
```
collindutter: Do we use `__` to denote private anywhere else in the library? Is it convention that `_` is protected, and `__` is private?

andrewfrench: We use it in the computer tool in a similar fashion: `__tempdir: Optional[tempfile.TemporaryDirectory] = field(default=None, kw_only=True)`

I'll defer on convention, but the motivation here is to create a read-only field otherwise inaccessible to the user, as someone setting these values locally doesn't make sense.

griptape/drivers/prompt/openai_chat_prompt_driver.py
collindutter: Never-mind, we do want `@property`s if we want to expose these fields to the users. Though they add quite a bit to the class size, so _is_ this something we want to expose to the users? Especially since even we don't know how we're going to use them currently.

pyproject.toml
collindutter: Muscle memory hahah. Maybe in the future we can run our tool tests after everything else.

griptape/drivers/prompt/openai_chat_prompt_driver.py
vasinov: Can't we just have it as `ratelimit_request_limit: Optional[int] = field(default=None)`?

andrewfrench: @vasinov That would allow it to be settable by the user, right? The approach here is to ensure it's a read-only property because, conceptually, it doesn't make sense to allow the user to write this value.

andrewfrench: Updated to protect the fields (`_fieldname`) and remove explicit getters.

collindutter (approved):

pyproject.toml
```diff
@@ -53,6 +53,7 @@ pgvector = "^0.2.3"
 psycopg2-binary = "^2.9.9"
 snowflake-connector-python = "^3.2"
 cryptography = "^41.0.4"
+dateparser = "^1.1.8"
```
collindutter, line 56: Do we still need this?

vasinov (approved):

PR #375, opened 2023-10-20
Update Event Listener Format

collindutter:


griptape/events/__init__.py
```diff
@@ -20,5 +21,7 @@
     "FinishPromptEvent",
     "StartStructureRunEvent",
     "FinishStructureRunEvent",
-    "CompletionChunkEvent"
+    "CompletionChunkEvent",
+
```
vachillo, line 25: extra space?

vachillo (approved):

griptape/events/__init__.py
collindutter: I intentionally added it for organization since Event Listener is a different thing than the Events. We do it in [other places](https://github.com/griptape-ai/griptape/blob/main/griptape/drivers/__init__.py) too.

Though once we merge in the autoformatter these intentional whitespaces are going to get annihilated lol.

griptape/structures/structure.py
```diff
@@ -143,25 +143,23 @@ def find_task(self, task_id: str) -> Optional[BaseTask]:
...
+    def publish_event(self, event: BaseEvent) -> None:
+        for event_listener in self.event_listeners:
+            handler = event_listener.handler
+            event_types = event_listener.event_types
+
+            if event_types is None:
+                handler(event)
+            else:
+                if type(event) in event_types:
+                    handler(event)
```
vasinov, line 161: Can be replaced with

```python
if event_types is None or type(event) in event_types:
    handler(event)
```

griptape/utils/stream.py
```diff
@@ -47,11 +47,10 @@ def run(self, *args) -> Iterator[TextArtifact]:
...
 
     def _run_structure(self, *args):
-        def event_handler(event):
+        def event_handler(event: BaseEvent):
             self._event_queue.put(event, True)
             self._event_queue.join()
 
-        self.structure.add_event_listener(CompletionChunkEvent, event_handler)
-        self.structure.add_event_listener(FinishStructureRunEvent, event_handler)
+        self.structure.add_event_listener(event_handler, [CompletionChunkEvent, FinishStructureRunEvent])
```
vasinov, line 54: So much cleaner!

vasinov (changes requested):

vasinov (approved):

PR #376, opened 2023-10-20
Add an activity for downloading objects to `AwsS3Client`

### LINKED ISSUE #237: Add an activity for downloading objects to `AwsS3Client`
`AwsS3Client` should enable LLMs to download files from S3. I propose adding a `download_objects_from_s3` activity that downloads files into tool memory.

andrewfrench:
Adds an activity to support downloading an object from S3. Tested locally:

```
...
[10/24/23 15:17:39] INFO     Subtask 28f4f3687edb4b499852636a9f122b5a
                             Thought: To provide a summary of the text, I first need to download the object from the S3 bucket. I
                             will use the AwsS3Client tool with the download_object activity to do this.

                             Action:
                             {
                               "type": "tool",
                               "name": "AwsS3Client",
                               "activity": "download_object",
                               "input": {
                                 "values": {
                                   "bucket_name": "www.afren.ch",
                                   "object_key": "2022-04-25.html"
                                 }
                               }
                             }
[10/24/23 15:17:40] INFO     Subtask 28f4f3687edb4b499852636a9f122b5a
                             Observation: Output of "AwsS3Client.download_object" was stored in memory with memory_name "ToolMemory" and
                             artifact_namespace "468be5c234e744b1ba9908900aae9190"
...
```

Resolves #237


griptape/tools/aws_s3_client/tool.py
```diff
@@ -174,6 +174,40 @@ def upload_content_to_s3(self, params: dict) -> ErrorArtifact | InfoArtifact:
...
+            obj = self.s3_client.get_object(
+                Bucket=bucket_name,
+                Key=object_key
+            )
+
+            # Return a TextArtifact if the object appears to be text.
+            if "text" in obj["ContentType"]:
+                return TextArtifact(obj["Body"].read())
+
+            return BlobArtifact(obj["Body"].read())
```
andrewfrench, line 206: Here, we attempt to interpret the object as text if the ContentType implies it's text. Otherwise, we handle it as a BlobArtifact.

collindutter (approved):

collindutter (approved):

griptape/tools/aws_s3_client/tool.py
```diff
@@ -211,6 +212,44 @@ def upload_content_to_s3(
...
+                    ): str,
+                    Literal(
+                        "object_key",
+                        description="The object key name to download.",
+                    ): str,
+                }
+            ),
+        }
+    )
+    def download_object(
```
vasinov, line 232: I think this should function more like `FileManager.load_files_from_disk` to provide the LLM with a way to download multiple files.

Additionally, I wonder if we should add a way to download all objects and load all files from a specific directory (and in `GoogleDriveClient.download_files`). Shouldn't be part of this PR but we should have an issue for that.

vasinov (changes requested):

collindutter (dismissed):

griptape/tools/aws_s3_client/tool.py
```diff
@@ -211,6 +212,42 @@ def upload_content_to_s3(
...
+            "schema": Schema(
+                {
+                    Literal(
+                        "bucket_name",
+                        description="The name of the S3 bucket to download from.",
+                    ): str,
+                    Literal(
+                        "object_keys", description="Object keys to download."
+                    ): [],
+                }
```
vasinov, line 227: I was thinking it could support multiple buckets:

```python
Literal(
    "objects",
    ...
): [
    {
        Literal("bucket_name", ...): str,
        Literal("object_key"): str
    }
]
```

vasinov (changes requested):
Since it's new functionality, could you change the PR to point at `dev` instead of `main`?

vasinov (dismissed):

collindutter (dismissed):

collindutter (approved):

vasinov (approved):

Issue #357, opened 2023-10-11
Events and tasks should be serializable


vasinov:
Event listeners often need to pass events to downstream systems in a serialized way. I propose adding Marshmallow serialization to events and tasks (since a lot of events contain tasks) similar to how memory and artifacts are serialized.

PR #378, opened 2023-10-23
Event Serialization

collindutter:
Closes https://github.com/griptape-ai/griptape/issues/357


griptape/events/base_event.py
```diff
@@ -8,3 +12,47 @@ class BaseEvent(ABC):
     timestamp: float = field(
         default=Factory(lambda: time.time()), kw_only=True
     )
+    type: str = field(default=Factory(lambda self: self.__class__.__name__, takes_self=True), kw_only=True)
+
+    @classmethod
+    def from_dict(cls, event_dict: dict) -> BaseEvent:
```
vasinov, line 18: If the use wants to implement a new event will they have to override this method? There is a way to abstract this but I don't know if it makes sense at this stage.

For now, I think, we can just add a section in the docs describing how to add and use custom events.

griptape/tasks/action_subtask.py
```diff
@@ -31,13 +31,18 @@ class ActionSubtask(PromptTask):
...
     action_input: Optional[dict] = field(default=None, kw_only=True)
+    _input: Optional[TextArtifact] = field(default=None, init=False)
 
     _tool: Optional[BaseTool] = None
     _memory: Optional[ToolMemory] = None
 
     @property
     def input(self) -> TextArtifact:
-        return TextArtifact(self.input_template)
+        if self._input is not None:
```
vasinov, line 41: What's the purpose if this?

griptape/tasks/base_task.py
```diff
@@ -111,7 +110,7 @@ def execute(self) -> BaseArtifact:
         try:
             self.state = BaseTask.State.EXECUTING
 
-            self.structure.publish_event(StartTaskEvent(task=self))
+            self.structure.publish_event(StartTaskEvent.from_task(self))
```
vasinov, line 113: Let's move it to `before_run`.

griptape/tasks/base_task.py
```diff
@@ -123,7 +122,7 @@ def execute(self) -> BaseArtifact:
             self.output = ErrorArtifact(str(e))
         finally:
             self.state = BaseTask.State.FINISHED
-            self.structure.publish_event(FinishTaskEvent(task=self))
+            self.structure.publish_event(FinishTaskEvent.from_task(self))
```
vasinov, line 125: Let's move it to `after_run`. And call it after line 116 (we are not currently calling it for some reason).

griptape/tasks/action_subtask.py
vasinov: Also, not sure it's ever a good idea to swap out values in `TextArtifact`s since they might have other stuff (like metadata and embeddings data) not match. If we want to "update" the value, it needs to be a method that generates a new object.

vasinov (changes requested):

griptape/tasks/action_subtask.py
collindutter: Comments were pending 🙃.  Multiple calls to `input` generate Artifacts with new `id`s. Is this intentional?

griptape/tasks/base_task.py
collindutter: Done. We are already calling `self.after_run()` on line 118, did you mean something else?

griptape/events/base_event.py
collindutter: They would. I had copied the implementation from [BaseArtifact](https://github.com/griptape-ai/griptape/blob/main/griptape/artifacts/base_artifact.py#L34), do we need to update docs for that as well? What did you have in mind for generalizing it? The current implementation of using `class_registry` feels wrong as it is described as an [private API](https://marshmallow.readthedocs.io/en/stable/_modules/marshmallow/class_registry.html).

griptape/tasks/action_subtask.py
vasinov: Ah, I see! I think there is a better way to fix it (e.g., move this into a real property) but I think for now it's better to rebuild an artifact every time than potentially have mismatched "internals." Could result in one of those silent bug scenarios that are hard to trace down.

griptape/tasks/base_task.py
vasinov: Oops, nevermind—line 118 was folded in the PR code view.

griptape/events/base_task_event.py
```diff
@@ -0,0 +1,27 @@
...
+from attrs import define, field
+from typing import Optional, TYPE_CHECKING
+from griptape.artifacts import BaseArtifact
+from .base_event import BaseEvent
+
+if TYPE_CHECKING:
+    from griptape.tasks import BaseTask
+
+@define
+class BaseTaskEvent(BaseEvent):
```
vasinov, line 11: Should be `BaseTaskEvent(BaseEvent, ABC)`.

griptape/events/base_action_subtask_event.py
```diff
@@ -0,0 +1,32 @@
...
+from attrs import define, field
+from typing import Optional, TYPE_CHECKING
+from .base_task_event import BaseTaskEvent
+
+
+if TYPE_CHECKING:
+    from griptape.tasks import ActionSubtask
+
+@define
+class BaseActionSubtaskEvent(BaseTaskEvent):
```
vasinov, line 11: Should be `BaseActionSubtaskEvent(BaseTaskEvent, ABC)`.

griptape/events/base_action_subtask_event.py
```diff
@@ -0,0 +1,32 @@
...
+    subtask_action_input: Optional[dict] = field(kw_only=True)
+
+    @classmethod
+    def from_task(cls, task: ActionSubtask) -> BaseActionSubtaskEvent:
+        return cls(
+            task_id=task.id,
+            task_parent_ids=task.parent_ids,
+            task_child_ids=task.child_ids,
+            task_input=task.input,
+            task_output=task.output,
```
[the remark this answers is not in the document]
vasinov: Yeah, doesn't seem ideal...let's keep it the way it is for now and think about how to improve it later.

griptape/schemas/events/base_action_subtask_event_schema.py
```diff
@@ -0,0 +1,10 @@
+from marshmallow import fields
+from griptape.schemas import BaseTaskEventSchema
+
+
+class BaseActionSubtaskEventSchema(BaseTaskEventSchema):
+    subtask_parent_task_id = fields.Str()
+    subtask_thought = fields.Str()
+    subtask_action_type = fields.Str()
+    subtask_action_name = fields.Str()
+    subtask_action_input = fields.Dict()
```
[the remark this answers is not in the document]
vasinov: May be no prefix for brevity?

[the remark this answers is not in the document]
collindutter: After thinking on it, I think we should keep them prefixed with `subtask`. This clearly differentiates them from the `task` prefixed fields in `BaseTaskEventSchema`, and is consistent with the naming of `subtasks` in `ToolkitTask`.

griptape/tasks/base_text_input_task.py
```diff
@@ -14,14 +14,18 @@ class BaseTextInputTask(BaseTask, ABC):
...
-            )
-        )
+        input_str = J2().render_from_string(self.input_template, **self.full_context)
+            
+        if self._input is not None:
+            self._input.value = input_str
+        else:
+            self._input = TextArtifact(input_str)
+            
+        return self._input
```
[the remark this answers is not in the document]
collindutter: Reverted

griptape/tasks/action_subtask.py
```diff
@@ -31,13 +31,18 @@ class ActionSubtask(PromptTask):
...
+            self._input.value = self.input_template
+        else:
+            self._input = TextArtifact(self.input_template)
+        return self._input
```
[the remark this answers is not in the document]
collindutter: Reverted

griptape/tasks/action_subtask.py
collindutter: Reverted.

griptape/events/base_event.py
```diff
@@ -1,8 +1,71 @@
...
+        )
+        class_registry.register(
+            "StartStructureRunEvent", StartStructureRunEventSchema
+        )
+        class_registry.register(
+            "FinishStructureRunEvent", FinishStructureRunEventSchema
+        )
+        class_registry.register(
+            "CompletionChunkEvent", CompletionChunkEventSchema
+        )
```
vasinov, line 50: This is goofy auto-formatting :)

Can't really do anything about it though, right?

griptape/events/base_event.py
vasinov: I just realized that what I had in mind won't work. Probably not worth worrying about for now.

vasinov (approved):

griptape/events/base_event.py
collindutter: Nothing outside of increasing the line length to more than 80 🤷‍♂️

PR #388, opened 2023-10-25
Refactor Embeddings Drivers To Use Chunkers

collindutter:
This PR contains several key changes:

1.  Moved the chunking logic from `OpenAiEmbeddingDriver` into `BaseEmbeddingDriver`. 
2. Refactored Embedding Drivers to use Chunkers when embedding long text. Previously they relied on Tokenizer's `encode` method which not every Tokenizer supported.
3. Fixed Chunkers not chunking contiguous text with no separators. I.e. `foobarfoobar` would not previously be chunked into `foobar` and `foobar`.
4. Removed Tokenizer's `encode` and `decode` methods. Most APIs did not support such functionality, and we no longer need them for embedding long text.


griptape/tokenizers/base_tokenizer.py
```diff
@@ -27,19 +24,6 @@ def tokens_left(self, text: str) -> int:
...
-        while batch := tuple(islice(it, self.max_tokens)):
-            yield batch
-
     @abstractmethod
-    def encode(self, text: str) -> list[int]:
-        ...
-
-    @abstractmethod
-    def decode(self, tokens: list[int]) -> str:
+    def token_count(self, text: str) -> int:
```
collindutter, line 28: While we're here, should this be renamed to `count_tokens`? `token_count` feels more like a property.

tests/unit/drivers/embedding/test_azure_openai_embedding_driver.py
```diff
@@ -18,12 +18,5 @@ def driver(self):
     def test_init(self, driver):
         assert driver
 
-    def test_try_embed_string(self, driver):
-        assert driver.try_embed_string("foobar") == [0, 1, 0]
-
```
collindutter, line 23: Moved to test for parent class.

tests/unit/drivers/embedding/test_azure_openai_embedding_driver.py
```diff
@@ -18,12 +18,5 @@ def driver(self):
...
     def test_embed_chunk(self, driver):
-        assert driver.embed_chunk("foobar") == [0, 1, 0]
-        assert driver.embed_chunk([1, 2, 3]) == [0, 1, 0]
-
-    def test_embed_long_string(self, driver):
-        assert driver.embed_long_string("foobar" * 5000) == [0, 1, 0]
```
collindutter, line 29: Moved to test for parent class.

tests/unit/drivers/embedding/test_openai_embedding_driver.py
```diff
@@ -15,30 +15,16 @@ def mock_openai(self, mocker):
     def test_init(self):
         assert OpenAiEmbeddingDriver()
 
-    def test_try_embed_string(self):
-        assert OpenAiEmbeddingDriver().try_embed_string("foobar") == [0, 1, 0]
-
-    def test_try_embed_string_with_long_string(self):
-        assert OpenAiEmbeddingDriver().try_embed_string(
-            " ".join(["foobar"] * 5000)
-        ) == [0, 1, 0]
```
collindutter, line 24: Moved to test for parent class.

tests/unit/loaders/test_pdf_loader.py
```diff
@@ -22,7 +22,7 @@ def test_load(self, loader):
 
         artifacts = loader.load(path)
 
-        assert len(artifacts) == 149
+        assert len(artifacts) == 151
```
collindutter, line 25: Have not dug into why this changed. I assume it's from some edge-case in the PDF chunking that is now handled by the non-separator fallback method.

vasinov: Just thought of something...let's add tests for the last chunk that tests the last sentence with `endswith`, so are 100% sure that we are not dropping any chunks in the process.

griptape/chunkers/base_chunker.py
```diff
@@ -48,62 +48,78 @@ def _chunk_recursively(
             tokens_count = 0
             half_token_count = token_count // 2
 
+            # If a separator is provided, only use separators after it.
```
vasinov, line 51: Thanks for adding the comments!

griptape/drivers/embedding/base_embedding_driver.py
```diff
@@ -1,21 +1,62 @@
...
 
 
 @define
 class BaseEmbeddingDriver(ExponentialBackoffMixin, ABC):
     dimensions: int = field(kw_only=True)
+    tokenizer: BaseTokenizer = field(kw_only=True)
+    chunker: BaseChunker = field(init=False)
+
+    def __attrs_post_init__(self) -> None:
+        self.chunker = TextChunker(tokenizer=self.tokenizer)
```
vasinov, line 18: Let's move this to a `Factory`.

vasinov (changes requested):

griptape/drivers/embedding/base_embedding_driver.py
collindutter: Had it as a `Factory` originally but was running into errors about `tokenizer` not being initialized. Let me try again.

vasinov: I think it might have to be `kw_only=True`.

griptape/tokenizers/base_tokenizer.py
collindutter: Done. Also renamed `tokens_left` to `count_tokens_left`

tests/unit/loaders/test_pdf_loader.py
collindutter: Done

griptape/drivers/embedding/base_embedding_driver.py
collindutter: Removed default, child classes must initialize the Chunker.

collindutter: Kept in `__attrs_post_init` based on Slack conversation.

vasinov (approved):

PR #397, opened 2023-10-31
refactor: remove unused default constant

iloveitaly:
Working on another doc change that adds model to the OpenAiTokenizer() example.
These defaults make it feel like they are used in some way when creating a tokenizer.
Removing this makes it more clear that they aren't used by the griptape system at all.


vasinov (approved):

vasinov:
Could you please update destination branch to `dev` in this PR?

vasinov (changes requested):
Could you please update destination branch to `dev` in this PR?

collindutter:
Does the `DEFAULT_` prefix for any of the fields make sense anymore?

collindutter:
@vasinov @iloveitaly what do you think about this? Does it makes sense to have `DEFAULT_` fields in any of the Tokenizers? Should we just remove these fields entirely so that we're not on the hook for maintaining model names?

vasinov:
> @vasinov @iloveitaly what do you think about this? Does it makes sense to have `DEFAULT_` fields in any of the Tokenizers? Should we just remove these fields entirely so that we're not on the hook for maintaining model names?

Would we just rename constants? Because we still need friendly defaults for those things, right?

collindutter:
> Because we still need friendly defaults for those things, right?

These defaults aren't actually used in the Tokenizer though. I think the very least we should remove the `DEFAULT_` prefix since they're more of a reference of available models that can be used. And while we're here, does it really make sense for the Tokenizer to be the source of truth on available models?

collindutter:
Implemented in https://github.com/griptape-ai/griptape/pull/690

PR #404, opened 2023-11-01
Add the requests library as a hard dependency

vasinov:
Add the requests library as a hard dependency, remove it from individual tool requirements.txt files.


collindutter:
Should we move [this import](https://github.com/griptape-ai/griptape/blob/main/griptape/tools/web_search/tool.py#L45) to top level?

collindutter (approved):

Issue #226, opened 2023-09-14
Replace `Union[]` with `|`


vasinov:
Since we already support the [pipe syntax](https://peps.python.org/pep-0604/) for union types in tools, we should update the whole code base to do it.

This task requires:
1. Replacing `Union[]` with `|`.
2. Removing `typing.Union` imports.
3. Adding `from __future__ import annotations` to files that use `|`.

The-Keeper:
Hello. I can work on this.

The-Keeper:
@vasinov 
The relevant PR is here: https://github.com/griptape-ai/griptape/pull/235

PR #407, opened 2023-11-02
Add image generation functionality

### LINKED ISSUE #332: Add image generation functionality
We want to extend Griptape with image generation capabilities, connecting it to image generation back-ends such as DALL-E and Stable Diffusion.

This integration should follow a typical Griptape layout:

* Add image generation drivers for different image generation back-ends.
* Add a generic `BaseImageGenEngine`
* Add a `PromptImageGenEngine(BaseImageGenEngine)` that uses image generation drivers.
* Add an `ImageGenTask` that uses an image generation engine.
* Add an `ImageGenerator` tool that uses an image generation engine.

andrewfrench:
Closes #332


griptape/artifacts/image_artifact.py
```diff
@@ -0,0 +1,11 @@
+from typing import Optional
+from attr import define, field
+from griptape.artifacts import BlobArtifact
+
+
+@define(frozen=True)
+class ImageArtifact(BlobArtifact):
+    mime_type: str = field(default="image/png", kw_only=True)
```
vasinov, line 8: Not sure this should be the default. Can we make it more generic? Seems like we [might have options](https://superuser.com/questions/979135/is-there-a-generic-mime-type-for-all-image-files).

griptape/artifacts/image_artifact.py
```diff
@@ -0,0 +1,11 @@
+from typing import Optional
+from attr import define, field
+from griptape.artifacts import BlobArtifact
+
+
+@define(frozen=True)
+class ImageArtifact(BlobArtifact):
```
vasinov, line 7: Should also implement `to_dict`.

griptape/drivers/image_generation/base_image_generation_driver.py
```diff
@@ -0,0 +1,8 @@
+from abc import abstractmethod
+from griptape.artifacts import ImageArtifact
+
+
+class BaseImageGenerationDriver:
```
vasinov, line 5: Add `(ABC)`.

griptape/drivers/image_generation/dalle2_image_generation_driver.py
```diff
@@ -0,0 +1,57 @@
...
+import requests
+from attr import field, Factory, define
+from griptape.artifacts import ImageArtifact
+from griptape.drivers.image_generation.base_image_generation_driver import (
+    BaseImageGenerationDriver,
+)
+
+
+@define
+class Dalle2ImageGenerationDriver(BaseImageGenerationDriver):
```
vasinov, line 15: Let's make it a generic `DalleImageGenerationDriver` and manage version at the class property level.

griptape/drivers/image_generation/dalle2_image_generation_driver.py
```diff
@@ -0,0 +1,57 @@
+import json
+import os
+import openai
+from typing import Optional
+
+import requests
+from attr import field, Factory, define
+from griptape.artifacts import ImageArtifact
+from griptape.drivers.image_generation.base_image_generation_driver import (
```
vasinov, line 9: Drop `base_image_generation_driver`.

griptape/drivers/image_generation/dalle2_image_generation_driver.py
vasinov: Should we prepend with `OpenAi` to match other driver naming schemas?

griptape/engines/image_generation/prompt_image_generation_engine.py
```diff
@@ -0,0 +1,28 @@
...
+    BaseImageGenerationDriver,
+)
+from griptape.engines.image_generation.base_image_generation_engine import (
+    BaseImageGenerationEngine,
+)
+from griptape.utils import J2
+
+
+@define
+class PromptImageGenerationEngine(BaseImageGenerationEngine):
```
vasinov, line 14: Should we make it a generic `ImageGenerationEngine` engine and add multiple methods for text-to-image and image-to-image modes? Or should we have different engines that encapsulate text-to-image (i.e., `PromptImageGenerationEngine`) and image-to-image?

griptape/tasks/image_generation_task.py
```diff
@@ -0,0 +1,36 @@
...
+    PromptImageGenerationEngine,
+)
+from griptape.tasks import BaseTextInputTask
+
+
+@define
+class ImageGenerationTask(BaseTextInputTask):
+    image_generation_driver: BaseImageGenerationDriver = field(
+        default=Factory(lambda: Dalle2ImageGenerationDriver()), kw_only=True
+    )
```
vasinov, line 20: Should we remove this to simplify the interface? The user can always pass a custom driver into `image_generation_engine`.

griptape/templates/engines/image_generation/prompt_image_generation.j2
```diff
@@ -0,0 +1,3 @@
+Description: """{{ description }}"""
+
+Generate an image based on the description provided.
```
vasinov, line 3: Let's sync up with @shhlife, @Amaru-Zeas, or @averyroche about what the best base prompt is.

griptape/tools/image_generator/tool.py
```diff
@@ -0,0 +1,57 @@
...
+)
+from griptape.tools import BaseTool
+from griptape.utils.decorators import activity
+
+
+@define
+class ImageGenerator(BaseTool):
+    image_generation_driver: BaseImageGenerationDriver = field(
+        default=Factory(lambda: Dalle2ImageGenerationDriver()), kw_only=True
+    )
```
vasinov, line 22: Same comment as in the task: could probably remove this.

griptape/engines/image_generation/prompt_image_generation_engine.py
andrewfrench: I've gone back and forth on it. This implementation represents the assumption that the interface and mechanics for text-to-image and image-to-image will be substantially different or that providers (and their drivers) might only provide a subset of that functionality. We could overcome each of those obstacles, just chose a path here.

griptape/templates/engines/image_generation/prompt_image_generation.j2
andrewfrench: +1. I think we should leave the flowery specifics to the description field, but I can imagine reinforcing instructions like 'stick to the description' or optionally 'be creative', etc.

griptape/artifacts/image_artifact.py
andrewfrench: Agree that a default doesn't quite make sense here. Updated to require the mime_type value.

griptape/schemas/artifacts/image_artifact_schema.py
```diff
@@ -0,0 +1,17 @@
+from marshmallow import post_load, fields
+from griptape.schemas import BaseArtifactSchema
+
+
+class ImageArtifactSchema(BaseArtifactSchema):
+    value = fields.Str()
+    mime_type = fields.Str()
+    width = fields.Int()
```
collindutter, line 8: This might have problems serializing if width/height/etc are optional.

griptape/tools/image_generator/tool.py
```diff
@@ -0,0 +1,39 @@
...
+                }
+            ),
+        }
+    )
+    def generate_image(self, params: dict) -> ImageArtifact | ErrorArtifact:
+        prompts = params["values"]["prompts"]
+        negative_prompts = params["values"]["negative_prompts"]
+
+        try:
+            return self.image_generation_engine.generate_image(prompts=prompts, negative_prompts=negative_prompts)
```
collindutter, line 36: Can `FileManager` work with `ImageArtifact`s?

griptape/tasks/image_generation_task.py
```diff
@@ -0,0 +1,21 @@
...
+class ImageGenerationTask(BaseTextInputTask):
+    image_generation_engine: ImageGenerationEngine = field(kw_only=True)
+
+    def run(self) -> ImageArtifact | ErrorArtifact:
+        try:
+            image_artifact = self.image_generation_engine.generate_image(
+                prompts=[self.input.to_text()], negative_prompts=[]
+            )
+
+            return image_artifact
```
collindutter, line 18: What happens when this is returned? What can the user do with it?

collindutter (changes requested):
Great work, this is going to be awesome!

griptape/tasks/image_generation_task.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+from attr import define, field
+from griptape.artifacts import ImageArtifact, ErrorArtifact
+from griptape.engines import ImageGenerationEngine
+from griptape.tasks import BaseTextInputTask
+
+
+@define
+class ImageGenerationTask(BaseTextInputTask):
```
collindutter, line 9: All Tasks support Rules now, can we integrate that into here too?

griptape/engines/image_generation/image_generation_engine.py
```diff
@@ -0,0 +1,10 @@
+from attr import field, define
+from griptape.drivers import BaseImageGenerationDriver
+
+
+@define
+class ImageGenerationEngine:
+    image_generation_driver: BaseImageGenerationDriver = field(kw_only=True)
+
+    def generate_image(self, prompts: list[str], negative_prompts: list[str], **kwargs):
+        return self.image_generation_driver.generate_image(prompts, negative_prompts, **kwargs)
```
collindutter, line 10: I think the Engine should use `J2` to render out a single string containing prompts, negative prompts, rules, etc. Then this could be passed to the driver.

collindutter (changes requested):

griptape/drivers/image_generation/openai_dalle_image_generation_driver.py
```diff
@@ -0,0 +1,72 @@
+import os
+import openai
```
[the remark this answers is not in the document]
collindutter: #422 was just merged into `dev` btw.

griptape/engines/image_generation/image_generation_engine.py
andrewfrench: Curious what this would look like, maybe you can expand on that idea? The drivers and their underlying models, as a rule, provide very heterogenous interfaces (see ChatGPT vs. Stable Diffusion prompt discussion [on Slack](https://griptape-ai.slack.com/archives/C05T3HLBH52/p1698955113291039)). Would this mean the engine renders a prompt string which then has to be parsed by the driver and passed to the model as API parameters?

More concretely: DALLE2/3 accepts a single prompt string (not sure if it can support negative weighting in prompts, could be experimented with), Leonardo expects a single prompt and a single negative prompt string, and Stable Diffusion accepts an arbitrary list of prompts with user-defined weights [-1, 1].

griptape/tools/image_generator/tool.py
andrewfrench: Absolutely, it's pretty seamless. I have some previous Friday demos and test scripts we can go over to verify that.

griptape/artifacts/image_artifact.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+from typing import Optional
+from attr import define, field
+from griptape.artifacts import BlobArtifact
+
+
+@define(frozen=True)
+class ImageArtifact(BlobArtifact):
+    mime_type: str = field(kw_only=True)
+    width: Optional[int] = field(default=None, kw_only=True)
```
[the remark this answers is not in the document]
andrewfrench: There's really no great argument to make these optional, updated to require width and height.

griptape/drivers/image_generation/openai_dalle_image_generation_driver.py
```diff
@@ -0,0 +1,74 @@
...
+        default=Factory(
+            lambda self: openai.OpenAI(api_key=self.api_key, base_url=self.base_url, organization=self.organization),
+            takes_self=True,
+        )
+    )
+    style: Optional[str] = field(default=None, kw_only=True)
+    quality: Union[Literal["standard"], Literal["hd"]] = field(default=Literal["standard"], kw_only=True)
+    image_size: Union[
+        Literal["256x256"], Literal["512x512"], Literal["1024x1024"], Literal["1024x1792"], Literal["1792x1024"]
+    ] = field(default=Literal["512x512"], kw_only=True)
```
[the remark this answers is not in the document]
collindutter: I think it's somewhat unavoidable to expose SDK patterns in Drivers since their primary purpose is to sit right on top of the SDK/API. Even in the case of `image_size`, a seemingly universal field, it seems non-straightforward to implement it at the `BaseImageGenerationDriver` level. I think providing helpful type hints is a good enough solution.

griptape/drivers/image_generation/openai_dalle_image_generation_driver.py
```diff
@@ -39,30 +38,37 @@ class OpenAiDalleImageGenerationDriver(BaseImageGenerationDriver):
         )
     )
     style: Optional[str] = field(default=None, kw_only=True)
-    quality: Optional[str] = field(default=None, kw_only=True)
-    image_width: int = field(default=512, kw_only=True)
-    image_height: int = field(default=512, kw_only=True)
+    quality: Union[Literal["standard"], Literal["hd"]] = field(default=Literal["standard"], kw_only=True)
+    image_size: Union[
```
collindutter, line 42: Replace `Union` with `|`. See https://github.com/griptape-ai/griptape/issues/226.

collindutter (changes requested):

griptape/drivers/image_generation/openai_dalle_image_generation_driver.py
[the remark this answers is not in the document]
andrewfrench: It's certainly less maintenance to do it this way. If we define and enforce a standard Griptape interface style, the intermediate layer would be a constant source of frustration as we'd have to re-define (or somehow handle) all possible options to some degree. On the other hand, we require the user do some research and have knowledge of the underlying dependency (OpenAI SDK, Leonardo API) before use. Not at all unreasonable, but a bit annoying.

griptape/drivers/image_generation/base_image_generation_driver.py
```diff
@@ -10,5 +10,5 @@ class BaseImageGenerationDriver(ABC):
     model: str = field(kw_only=True)
 
     @abstractmethod
-    def generate_image(self, prompts: list[str], negative_prompts: list[str], **kwargs) -> ImageArtifact:
+    def generate_image(self, prompts: list[str], negative_prompts: list[str] = list, **kwargs) -> ImageArtifact:
```
collindutter, line 13: Python gotcha https://docs.python-guide.org/writing/gotchas/

collindutter (changes requested):

griptape/drivers/image_generation/base_image_generation_driver.py
```diff
@@ -10,5 +10,5 @@ class BaseImageGenerationDriver(ABC):
     model: str = field(kw_only=True)
 
     @abstractmethod
-    def generate_image(self, prompts: list[str], negative_prompts: list[str] = list, **kwargs) -> ImageArtifact:
+    def generate_image(self, prompts: list[str], negative_prompts: list[str] = None, **kwargs) -> ImageArtifact:
```
collindutter, line 13: `list[str]` should be `Optional[list[str]]`

griptape/drivers/image_generation/leonardo_image_generation_driver.py
```diff
@@ -31,7 +31,10 @@ class LeonardoImageGenerationDriver(BaseImageGenerationDriver):
     image_width: int = field(default=512, kw_only=True)
     image_height: int = field(default=512, kw_only=True)
 
-    def generate_image(self, prompts: list[str], negative_prompts: list[str], **kwargs) -> ImageArtifact:
+    def generate_image(self, prompts: list[str], negative_prompts: list[str] = None, **kwargs) -> ImageArtifact:
```
collindutter, line 34: `Optional[list[str]]`

griptape/engines/image_generation/image_generation_engine.py
```diff
@@ -10,11 +10,14 @@ class ImageGenerationEngine:
     def generate_image(
         self,
         prompts: list[str],
-        negative_prompts: list[str] = list,
-        rulesets: list[Ruleset] = list,
-        negative_rulesets: list[Ruleset] = list,
+        negative_prompts: list[str] = None,
```
collindutter, line 13: `Optional`

griptape/engines/image_generation/image_generation_engine.py
```diff
@@ -10,11 +10,14 @@ class ImageGenerationEngine:
...
+        rulesets: list[Ruleset] = None,
+        negative_rulesets: list[Ruleset] = None,
         **kwargs
     ):
+        if not negative_prompts:
+            negative_prompts = []
+
         for ruleset in rulesets:
```
collindutter, line 21: Need to initialize `rulesets` and `negative_rulesets` to empty lists of `None`

collindutter (changes requested):

griptape/engines/image_generation/image_generation_engine.py
```diff
@@ -0,0 +1,31 @@
...
+class ImageGenerationEngine:
+    image_generation_driver: BaseImageGenerationDriver = field(kw_only=True)
+
+    def generate_image(
+        self,
+        prompts: list[str],
+        negative_prompts: Optional[list[str]] = None,
+        rulesets: Optional[list[Ruleset]] = None,
+        negative_rulesets: Optional[list[Ruleset]] = None,
+        **kwargs
```
collindutter, line 18: Do we need `kwargs`? As far as I can tell it's not being used anywhere.

griptape/drivers/image_generation/leonardo_image_generation_driver.py
```diff
@@ -0,0 +1,85 @@
...
+        max_attempts: The maximum number of times to poll the Leonardo API for a completed image.
+        model_id: The ID of the model to use when generating images. If None, the default model will be used.
+        image_width: The width of the generated image in the range [32, 1024] and divisible by 8.
+        image_height: The height of the generated image in the range [32, 1024] and divisible by 8.
+
+    Details on Leonardo image generation parameters can be found here:
+    https://docs.leonardo.ai/reference/creategeneration
+    """
+
+    api_key: str = field(default=Factory(lambda: os.environ.get("LEONARDO_API_KEY")), kw_only=True)
```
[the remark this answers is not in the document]
collindutter: Looks like this change didn't get implemented

griptape/drivers/image_generation/leonardo_image_generation_driver.py
```diff
@@ -0,0 +1,89 @@
...
+                "height": self.image_height,
+                "num_images": 1,
+                "modelId": self.model,
+            },
+        ).json()
+
+        return response["sdGenerationJob"]["generationId"]
+
+    def _get_image_url(self, generation_id: str):
+        for attempt in range(self.max_attempts):
```
collindutter, line 73: This retry logic should probably be moved to `BaseImageGenerationDriver` like we have in `BasePromptDriver`.

griptape/drivers/image_generation/leonardo_image_generation_driver.py
```diff
@@ -0,0 +1,89 @@
...
+            response = self.requests_session.get(
+                url=f"{self.api_base}/generations/{generation_id}", headers={"authorization": f"Bearer {self.api_key}"}
```
collindutter, line 75: Capitalize `authorization`

griptape/drivers/image_generation/leonardo_image_generation_driver.py
```diff
@@ -0,0 +1,85 @@
...
+            if response["generations_by_pk"]["status"] == "PENDING":
+                time.sleep(attempt + 1)
+                continue
+
+            return response["generations_by_pk"]["generated_images"][0]["url"]
+        else:
+            raise Exception("image generation failed to complete")
+
+    def _download_image(self, url: str):
+        response = self.requests_session.get(url=url, headers={"authorization": f"Bearer {self.api_key}"})
```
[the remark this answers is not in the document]
collindutter: Looks like this change didn't get implemented.

griptape/drivers/image_generation/openai_dalle_image_generation_driver.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+    )
+    style: Optional[str] = field(default=None, kw_only=True)
+    quality: Literal["standard"] | Literal["hd"] = field(default="standard", kw_only=True)
+    image_size: (
+        Literal["256x256"] | Literal["512x512"] | Literal["1024x1024"] | Literal["1024x1792"] | Literal["1792x1024"]
+    ) = field(default="1024x1024", kw_only=True)
+    response_format: Literal["b64_json"] = field(default="b64_json", kw_only=True)
+
+    def generate_image(self, prompts: list[str], **kwargs) -> ImageArtifact:
```
collindutter, line 48: Missing `negative_prompts`?

griptape/drivers/image_generation/openai_dalle_image_generation_driver.py
```diff
@@ -0,0 +1,82 @@
...
+        return ImageArtifact(
+            value=image_data,
+            mime_type="image/png",
+            width=image_dimensions[0],
+            height=image_dimensions[1],
+            model=self.model,
+            prompt=prompt,
+        )
+
+    @staticmethod
```
collindutter, line 80: Should this be a static method? I don't think we really use `@staticmethod` anywhere else.

collindutter (changes requested):

griptape/drivers/image_generation/openai_dalle_image_generation_driver.py
andrewfrench: OpenAI's API doesn't support negative prompts. The prompt is rewritten, so the specifics of any 'do not' instructions have an opportunity to be warped by rewriting or summarization. From my experimentation negative prompts in a unified prompt seem to be counterproductive, like `negative prompts: text, clouds` results in more text and clouds in the image than not including the negative prompts at all.

griptape/engines/image_generation/image_generation_engine.py
andrewfrench: That was needed when additional parameters came through the method call and not the driver setup. Removed.

griptape/drivers/image_generation/leonardo_image_generation_driver.py
andrewfrench: That was my initial approach, but I found the log output from that to be overwhelming. We'll naturally expect a few request retries here to poll until the image generation is complete (or has failed), depending on service load and the request parameters and we shouldn't emit warnings under the expected happy path. 

Other providers expose fully-featured SDKs that manage polling and completion events for us and just return the image data directly, so Leonardo is (at the moment) the only driver that forces us into an async model. Maybe we want to implement retries on failed generations at the base driver level that would retry the entire generation request?

collindutter: Ahh I see, I didn't realize this is just for polling. Yeah lets retry failed ones similar to how we do in `BasePromptDriver`, but we can keep this polling logic here.

griptape/drivers/image_generation/openai_dalle_image_generation_driver.py
collindutter: I think the method signature should still include it though, right? Maybe we throw an error if it's provided.

andrewfrench: I realize now that you might just mean in the signature (i.e. instead of `**kwargs`). Updated to make that change. If I'm mistake, let me know.

collindutter (approved):

vasinov (approved):
Great work!

PR #409, opened 2023-11-02
Streaming chat utility

mattma1970:
An implementation of streaming chat utility.

### DIFF
--- a/griptape/utils/__init__.py
+++ b/griptape/utils/__init__.py
@@ -7,13 +7,12 @@ from .conversation import Conversation
 from .manifest_validator import ManifestValidator
 from .python_runner import PythonRunner
 from .command_runner import CommandRunner
-from .chat import Chat
 from .futures import execute_futures_dict
 from .token_counter import TokenCounter
 from .prompt_stack import PromptStack
 from .dict_utils import remove_null_values_in_dict_recursively
 from .stream import Stream
-
+from .chat import Chat
 
 def minify_json(value: str) -> str:
     return json.dumps(json.loads(value), separators=(",", ":"))
@@ -26,7 +25,6 @@ __all__ = [
     "CommandRunner",
     "minify_json",
     "J2",
-    "Chat",
     "str_to_hash",
     "dataframe_to_hash",
     "execute_futures_dict",
@@ -34,4 +32,5 @@ __all__ = [
     "PromptStack",
     "remove_null_values_in_dict_recursively",
     "Stream",
+    "Chat",
 ]
--- a/griptape/utils/chat.py
+++ b/griptape/utils/chat.py
@@ -4,9 +4,9 @@ from attr import define, field, Factory
 
 if TYPE_CHECKING:
     from griptape.structures import Structure
+from griptape.utils import Stream
 
-
-@define(frozen=True)
+@define
 class Chat:
     structure: Structure = field()
     exit_keywords: list[str] = field(default=["exit"], kw_only=True)
@@ -16,22 +16,37 @@ class Chat:
     prompt_prefix: str = field(default="Q: ", kw_only=True)
     response_prefix: str = field(default="A: ", kw_only=True)
     output_fn: Callable[[str], None] = field(
-        default=Factory(lambda: print), kw_only=True
+        default=lambda x: print (x, end=""), kw_only=True
     )
+    stream: bool = field(default=False,kw_only=True)
 
-    def start(self) -> None:
+    def start(self,initial_turn:str=None) -> None:
+        if initial_turn:
+            self.output_fn(initial_turn+'\n')
+
+        self.structure.stream = self.stream
+        self.structure.prompt_driver.stream = self.stream
+        if self.stream:
+            self.structure=Stream(self.structure)
+
         if self.intro_text:
-            self.output_fn(self.intro_text)
+            self.output_fn(self.intro_text+'\n')
         while True:
             question = input(self.prompt_prefix)
 
             if question.lower() in self.exit_keywords:
-                self.output_fn(self.exiting_text)
-
+                self.output_fn(self.exiting_text+'\n')
                 break
             else:
-                self.output_fn(self.processing_text)
+                self.output_fn(self.processing_text+'\n')
 
-            self.output_fn(
-                f"{self.response_prefix}{self.structure.run(question).output.to_text()}"
+            if self.stream:
+                self.output_fn(f"{self.response_prefix}")
+                self.structure.structure.event_listeners=[] # To address https://github.com/griptape-ai/griptape/issues/408
+                for chunk in self.structure.run(question):
+                    self.output_fn(chunk.value)
+                self.output_fn("\n")
+            else:
+                self.output_fn(
+                    f"{self.response_prefix}{self.structure.run(question).output.to_text()}"
             )


griptape/utils/__init__.py
collindutter, line 1: Was this change intentional? If not let's revert so that we reduce overall changes.

griptape/utils/chat.py
```diff
@@ -4,9 +4,10 @@
 
 if TYPE_CHECKING:
     from griptape.structures import Structure
+from griptape.utils import Stream
 
 
-@define(frozen=True)
+@define
```
collindutter, line 10: What is the reason for removing frozen?

griptape/utils/chat.py
```diff
@@ -16,22 +17,39 @@ class Chat:
     prompt_prefix: str = field(default="Q: ", kw_only=True)
     response_prefix: str = field(default="A: ", kw_only=True)
     output_fn: Callable[[str], None] = field(
-        default=Factory(lambda: print), kw_only=True
+        default=lambda x: print(x, end=""), kw_only=True
```
collindutter, line 20: What is the reason for removing `Factory`?

griptape/utils/chat.py
```diff
@@ -16,22 +17,39 @@ class Chat:
...
     )
+    stream: bool = field(default=False, kw_only=True)
+
+    def start(self, initial_turn: str = None) -> None:
+        if initial_turn:
+            self.output_fn(initial_turn + "\n")
```
collindutter, line 26: What is the reason for adding a newline? And if we're going to keep it, let's move it to line 20 inside of the default.

griptape/utils/chat.py
```diff
@@ -16,22 +17,39 @@ class Chat:
...
+
+        self.structure.stream = self.stream
+        self.structure.prompt_driver.stream = self.stream
```
collindutter, line 29: Is this necessary? The `Stream` utility should take care of adding this already.

griptape/utils/chat.py
```diff
@@ -16,22 +17,39 @@ class Chat:
...
+        if self.stream:
+            self.structure = Stream(self.structure)
```
collindutter, line 31: Let's not re-assign `self.structure`. This will change its' type.

griptape/utils/chat.py
```diff
@@ -16,22 +17,39 @@ class Chat:
...
+                self.output_fn(self.processing_text + "\n")
 
-            self.output_fn(
-                f"{self.response_prefix}{self.structure.run(question).output.to_text()}"
-            )
+            if self.stream:
+                self.output_fn(f"{self.response_prefix}")
+                self.structure.structure.event_listeners = (
+                    []
+                )  # To address https://github.com/griptape-ai/griptape/issues/408
```
collindutter, line 48: Let's make sure to resolve 408 first. Don't want to include workarounds.

griptape/utils/chat.py
```diff
@@ -16,22 +17,39 @@ class Chat:
...
                 break
             else:
-                self.output_fn(self.processing_text)
+                self.output_fn(self.processing_text + "\n")
 
-            self.output_fn(
-                f"{self.response_prefix}{self.structure.run(question).output.to_text()}"
-            )
+            if self.stream:
+                self.output_fn(f"{self.response_prefix}")
```
collindutter, line 45: Move outside of `self.stream` and remove from line 54.

griptape/utils/chat.py
```diff
@@ -16,22 +17,39 @@ class Chat:
...
+                self.structure.structure.event_listeners = (
+                    []
+                )  # To address https://github.com/griptape-ai/griptape/issues/408
+                for chunk in self.structure.run(question):
+                    self.output_fn(chunk.value)
```
collindutter, line 50: I think the `output_fn` implementation should be different for `stream=True` vs `stream=False`. For `stream=False` we can keep the old implementation. For `stream=True` we can use the new implementation.

collindutter (changes requested):
Thanks for the PR!

griptape/utils/__init__.py
mattma1970: Because the new streaming chat imports Stream, I got a circularity issue which was resolved by placing Chat import after Stream import. 

`ImportError: cannot import name 'Stream' from partially initialized module 'griptape.utils' (most likely due to a circular import) (/home/mtman/Documents/Repos/griptape_experiments/upstream-griptape/griptape/utils/__init__.py)`

griptape/utils/chat.py
mattma1970: In order to be able to wrap structure in Stream once rather than during each conversation turn. I saw your comments below regarding not reassigning so I'll revert this.

griptape/utils/chat.py
mattma1970: I changed the default output_fn to print(x, end="") to suppress the newline after each chunk returned during streaming. Utility needs both streamed and non-streamed output_fn. Having only one output_fn and adding \n or composing it to suppress the \n when looping over the generator feels very 'print' centric. 

Would having output_fn and stream_output_fn be the best solution?

griptape/utils/chat.py
mattma1970: I had thought about that. If we remove this then we'll require that the user add stream=True to the prompt_driver and the Agent definition. The request had asked for stream to be a parameter to be passed to Chat. 
To keep it as a parameter in Chat, should I create a deepcopy of the Agent and mutate the local copy or remove the stream parameter from Chat?

griptape/utils/chat.py
mattma1970: Will do.

griptape/utils/__init__.py
collindutter: Can you try importing Stream like this? `from .utils import Stream`

griptape/utils/chat.py
mattma1970: As discussed above, in some places we need \n and when streaming we need to suppress it. I think any modifications to the output_fn on the fly end up presuming the output_fn is 'print'. Would it be better to let users have the flexibility to specify 2 output_fn so we don't have to assume what they always be printing to console?

griptape/utils/chat.py
collindutter: Yeah we had gone back and forth on this but ultimately decided not to mutate.  Let's keep it consistent with `Stream` for now. The docs are actually incorrect on this point, fix incoming soon.

griptape/utils/chat.py
collindutter: Yeah I think two functions may be our best best here. Let's try it out and see how it looks.

griptape/utils/__init__.py
mattma1970: using from .stream import Stream works with the original order in __init__. I'll revert the __init__ ordering.

griptape/utils/chat.py
mattma1970: I'm new to attrs. Factory only seems to accept lambda functions that have no parameter or self. 
`default=Factory(lambda x: print(x))` raises output_fn.<lambda>() missing 1 required positional argument: 'x' when generating __init__

using `default = Factory(lambda: lambda x: print(x,end=""), kw_only=True)` works but is ugly. Is there a better way?

griptape/utils/chat.py
mattma1970: If the prefixes are printed prior to calling run on the agents either mode, you end up with output 
Q: What is the chance martians will invade earth? 
Processing...
A:
and then a delay until it starts getting tokens back. I found this a bit confusing from a UX perspective, not knowing if it had hung. With the two output functions now I think we could leave it as is for non-streaming output and render the response_prefix after the first chunk is returned. I think this is a better UX. What do you think?

griptape/utils/chat.py
collindutter: Ahh I see, thank you for clarifying. Ok in that case I think we can remove the `Factory`, it looks like we use this pattern [elsewhere](https://github.com/griptape-ai/griptape/blob/main/griptape/drivers/vector/local_vector_store_driver.py#L15). @vasinov any issues with this?

griptape/utils/chat.py
```diff
@@ -15,23 +16,37 @@ class Chat:
...
-
                 break
             else:
                 self.output_fn(self.processing_text)
 
-            self.output_fn(
-                f"{self.response_prefix}{self.structure.run(question).output.to_text()}"
-            )
+            if self.structure.stream:
+                first_chunk: bool = True
```
collindutter, line 42: @mattma1970 can you please explain this to me again? Why can't we just place `self.streaming_output_fn(f"{self.response_prefix}")` before the loop?

griptape/utils/chat.py
```diff
@@ -15,23 +16,37 @@ class Chat:
...
+        default=lambda x: print(x, end=""), kw_only=True
+    )
     output_fn: Callable[[str], None] = field(
         default=Factory(lambda: print), kw_only=True
     )
 
-    def start(self) -> None:
+    def start(self, initial_turn: str = None) -> None:
+        if initial_turn:
+            self.output_fn(initial_turn)
```
collindutter, line 28: What does `initial_turn` do? I think ideally we don't want to introduce another parameter.

collindutter (changes requested):
Getting really close, thank you for all the updates!

griptape/utils/chat.py
mattma1970: Thanks for persisting! In the request it was suggested that the streaming conversation would be started with

`Chat(agent, stream=True).run("Howdy!")`

I took this to mean that they'd like to be able to pass in an initial greeting to kick the conversation off.  That's what this is for.

griptape/utils/chat.py
mattma1970: It's a user experience issue. If you place it before the loop then the user will see the 'A:' prefix in the console and then have to wait until the first chunk is returned. That time to first token [chunk] can be uncomfortably long. It left me wondering if something had hung. If we include it in the loop then user will see 'processing..' which lets them believe something is happening and then the 'A: [first chunk']' appear together. 

Happy to change it if you have a different view about the UX.

collindutter: Got it, thank you for explaining! I wonder if we can use [this](https://rich.readthedocs.io/en/stable/reference/spinner.html) to display a spinner until the first chunk? We use `rich` in other places of the framework but we underutilize it.

griptape/utils/chat.py
collindutter: I just clarified with @shhlife offline, we don't need that functionality for now. Let's just focus on the streaming side of things.

collindutter:
Heads up @mattma1970 we changed some code style and dependency things on `dev`. When you merge `dev` into your branch, do the following:
1. Resolve the conflicts.
2. Run `poetry install --with dev --with test --all-extras`
3. Run `poetry run pre-commit install`
4. Run the formatter `poetry run black .`
5. Commit as usual.

griptape/utils/chat.py
```diff
@@ -15,23 +16,37 @@ class Chat:
...
+                first_chunk: bool = True
+                for chunk in Stream(self.structure).run(question):
+                    if first_chunk:
+                        self.streaming_output_fn(f"{self.response_prefix}")
+                        first_chunk = False
+                    self.streaming_output_fn(chunk.value)
+                self.streaming_output_fn("\n")
+            else:
+                self.output_fn(
+                    f"{self.response_prefix}{self.structure.run(question).output.to_text()}"
```
collindutter, line 51: I just changed your branch to be based off `dev`. And in this branch we changed the return value of `run` to be `Structure`. Can you please update `output.to_text()` to `output_task.output.to_text()`?

griptape/utils/chat.py
mattma1970: @shhlife's request still has that Chat should take a Stream parameter. Stream was removed from the utility after we decided not to mutate the structure [here (https://github.com/griptape-ai/griptape/pull/409#discussion_r1380741513)], instead we're leaving it to user to set Stream in the prompt driver and agent.

Should Chat still have Stream parameter? If so it could be used to validate that Stream is set in the prompt driver and the agent?

griptape/utils/chat.py
mattma1970: What do you have in mind there? As an alternative to the 'processing..' text or render the 'processing.. \n A:' and put a spinner after the A: ? 
@shhlife has a video in the Trade School where he shows that you can use rich for better command line experience. It feels like its personal taste to make use of it and what spinner to use. What if we had rich wired up in the chat utility but by default had it turned off. Users could then just pass a spinner name if they want it?

collindutter: Ok after thinking about it some more I don't think we need to overcomplicate things with `rich` right now. I was just trying to get away from the `first_chunk` pattern. Let's do this instead:

```python
stream = Stream(self.structure).run(question)
next(stream)
self.streaming_output_fn(self.response_prefix)
for chunk in stream:
    self.streaming_output_fn(chunk.value)
self.streaming_output_fn("\n")
```

griptape/utils/chat.py
collindutter: Let's also stick with our original implementation. The only place the user should set `Stream` is on the Structure.

collindutter:
Hey @mattma1970 just checking in on this to see if you'll be able to get the last few changes implemented soon. Thanks!

griptape/utils/chat.py
mattma1970: Hi. This will drop the first chunk unless we do
`self.streaming_output_fn(self.response_prefix+next(chunk.value)`

Does this still address your concerns about a 'first-token' pattern?

collindutter: Ah good catch. Let's do:

```python
stream = Stream(self.structure).run(question)
first_chunk = next(stream)
self.streaming_output_fn(self.response_prefix + first_chunk)
for chunk in stream:
    self.streaming_output_fn(chunk.value)
self.streaming_output_fn("\n")
```

griptape/utils/chat.py
```diff
@@ -25,9 +27,16 @@ def start(self) -> None:
 
             if question.lower() in self.exit_keywords:
                 self.output_fn(self.exiting_text)
-
                 break
             else:
                 self.output_fn(self.processing_text)
 
-            self.output_fn(f"{self.response_prefix}{self.structure.run(question).output_task.output.to_text()}")
+            if self.structure.stream:
```
collindutter, line 34: Let's update this to `self.structure.prompt_driver.stream` so that it works with all Prompt Drivers.

griptape/utils/chat.py
collindutter, line 1: Can you please update `test_chat.py`'s `test_init` with these new fields?

collindutter (changes requested):
Last changes I promise 😅

shhlife:
One note.. if you change line 19 to include `flush=True`:

`streaming_output_fn: Callable[[str], None] = field(default=lambda x: print(x, end=""), kw_only=True)`

to 

`streaming_output_fn: Callable[[str], None] = field(default=lambda x: print(x, end="", flush=True), kw_only=True)`


the user will see things updating faster.

PR #411, opened 2023-11-02
Hotfix: Remove Event Listeners after streaming

### LINKED ISSUE #408: Stream(agent).run() can cause hanging due to adding event_handlers multiple times.
[ x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Describe the bug**

This issue related to the griptape.utils.stream class, used to stream responses from an LLM api that can lead to  the CompletionChunkEvent and FinishStructureEvent defined into Stream._run_structure being multiple times and causing the threads to get blocked.

This arose for me when I creating a streaming fastAPI endpoint that uses a griptape agent and where the scope of the agent and/or the dictionary of event_listeners is greater than the endpoint function. In this case the event_handlers accumulated between invocations and this caused the code to hang (orphaned event_handlers or multiple .joins?)

**To Reproduce**

This code hangs on the second invocation of the endpoint. 
```
from fastapi import FastAPI
from fastapi.responses import StreamingResponse
from typing import Any, Dict

from griptape.events import CompletionChunkEvent, FinishStructureRunEvent
from griptape.utils import Stream
import uvicorn
import os

load_dotenv()

app=FastAPI()
            
@app.post('/chat_with_agent')
def chat_using_agent(message: Dict[str,str]) -> Any:
    resp =Stream(agent).run(message['user_input'])
    return StreamingResponse(resp)

if __name__=='__main__':     
    callbacks = {CompletionChunkEvent:[lambda x: print (x.token, end="")]}
    agent = Agent(
                    prompt_driver=OpenAiChatPromptDriver(model='gpt-3.5-turbo',stream=True),
                    event_listeners= callbacks,
                    stream=True,
            )
    uvicorn.run(app, host='0.0.0.0', port=8888)
```
In this case, each time the endpoint is invoked, the CompletionChunkEvent and FinishStructureEvent event handlers, defined inside Stream._run_structure(), are added to the agent event_listener attribute. If the endpoint is invoked twice then the handlers are added twice because the function griptape.structure.add_event_listener doesn't appear to prevent the duplication:
``` 
    def add_event_listener(
        self, event_type: Type[BaseEvent], event_listener: Callable
    ) -> None:
        if isinstance(self.event_listeners, list):
            if event_listener not in self.event_listeners:
                self.event_listeners.append(event_listener)
        elif isinstance(self.event_listeners, dict):
            if event_type not in self.event_listeners:
                self.event_listeners[event_type] = [event_listener]
            elif event_listener not in self.event_listeners[event_type]:
                self.event_listeners[event_type].append(event_listener)
```
Perhaps because of the threading Stream._run_structure the event_listener's passed in have different id(event_listener) and so are not found with the 'in' function(?)  and duplicate (r maybe orphaned) callables are added. This causes the code to hang.

Resetting the agent.event_listeners prior to Stream.run() removes the problem. 

```
@app.post('/chat_with_agent')
def chat_using_agent(message: Dict[str,str]) -> Any:
    agent.event_listeners = {CompletionChunkEvent:[lambda x: print (x.token+'-',end="")]}
    resp =Stream(agent).run(message['user_input'])
    return StreamingResponse(resp)
```

**Expected behavior**

Expected no side effects for agent.event_listeners when calling Stream(agent).run(). Maybe they could be an automatically removed prior to returning from _run_structure ?

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04

**Additional context**

collindutter:
Fixes #408.

- Clean up Event Listeners after Streaming
- Don't block on streaming thread.

### DIFF
--- a/griptape/structures/structure.py
+++ b/griptape/structures/structure.py
@@ -157,14 +157,18 @@ class Structure(ABC):
         return [self.add_task(s) for s in tasks]
 
     def add_event_listener(
-        self,
-        handler: Callable[[BaseEvent], Any],
-        event_types: Optional[list[Type[BaseEvent]]] = None,
-    ) -> None:
-        event_listener = EventListener(handler, event_types=event_types)
-
+        self, event_listener: EventListener
+    ) -> EventListener:
         if event_listener not in self.event_listeners:
             self.event_listeners.append(event_listener)
+
+        return event_listener
+
+    def remove_event_listener(self, event_listener: EventListener) -> None:
+        if event_listener in self.event_listeners:
+            self.event_listeners.remove(event_listener)
+        else:
+            raise ValueError(f"Event Listener not found.")
 
     def publish_event(self, event: BaseEvent) -> None:
         for event_listener in self.event_listeners:
--- a/griptape/utils/stream.py
+++ b/griptape/utils/stream.py
@@ -3,7 +3,7 @@ from typing import TYPE_CHECKING, Iterator
 from threading import Thread
 from queue import Queue
 from griptape.artifacts.text_artifact import TextArtifact
-from griptape.events import CompletionChunkEvent
+from griptape.events import CompletionChunkEvent, EventListener
 from attrs import field, define, Factory
 from griptape.events import BaseEvent, FinishStructureRunEvent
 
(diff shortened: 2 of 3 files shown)


griptape/events/event_listener.py
```diff
@@ -1,11 +1,20 @@
...
 
 @define
 class EventListener:
+    id: str = field(default=Factory(lambda: uuid.uuid4().hex), kw_only=True)
     handler: Callable[[BaseEvent], Any] = field()
     event_types: Optional[list[Type[BaseEvent]]] = field(
         default=None, kw_only=True
     )
+
+    def __eq__(self, other: EventListener) -> bool:
```
vasinov, line 16: I think you can just use the built-in `id()` method like this `id(self) == id(other)`, so no need to add `EventListener.id`.

vasinov (changes requested):

Issue #420, opened 2023-11-06
Support OpenAi SDK ^1.1.0


collindutter:

zepef:
Hello,

Supporting v1.1.1 would be much appreciated. I am getting now:

 in OpenAiChatPromptDriver
    api_base: str = field(default=openai.api_base, kw_only=True)
AttributeError: module 'openai' has no attribute 'api_base'. Did you mean: 'api_type'?

I suspect evolution of SDK being the cause for this error.

Best regards,

Pierre-Emmanuel

collindutter:
Hi @zepef apologies for this issue, we will be making a release today that locks `openai` at `^0.28` until we finish this issue (soon).

zepef:
Hello Collin,

I will try it right away. I am really impressed by griptape which I am using to create a medical chatbot. I will keep you updated.

All the best !

Pierre-Emmanuel
________________________________
De : Collin Dutter ***@***.***>
Envoyé : mardi 7 novembre 2023 19:53
À : griptape-ai/griptape ***@***.***>
Cc : Pierre-Emmanuel FEGA ***@***.***>; Mention ***@***.***>
Objet : Re: [griptape-ai/griptape] Support OpenAi SDK ^1.1.0 (Issue #420)


@zepef<https://github.com/zepef> v0.19.4<https://github.com/griptape-ai/griptape/releases/tag/v0.19.4> has been released which should address this issue.

zepef:
Collin,

In fact it is an issue with OpenAI 1.1.1 because I get:

ImportError: cannot import name 'OpenAI' from 'openai' (E:\Users\clt\anaconda3\envs\w2\lib\site-packages\openai\__init__.py)

I guess I'll wait until it is supported by griptape because I coded  STT/TTS for my chatbot using new OpenAI SDK.

Regards,

Pierre-Emmanuel
________________________________
De : Collin Dutter ***@***.***>
Envoyé : mardi 7 novembre 2023 19:53
À : griptape-ai/griptape ***@***.***>
Cc : Pierre-Emmanuel FEGA ***@***.***>; Mention ***@***.***>
Objet : Re: [griptape-ai/griptape] Support OpenAi SDK ^1.1.0 (Issue #420)


@zepef<https://github.com/zepef> v0.19.4<https://github.com/griptape-ai/griptape/releases/tag/v0.19.4> has been released which should address this issue.

collindutter:
[...]

PR #421, opened 2023-11-07
fix: api_base => base_url

iloveitaly:
version lock on the openai package probably needs to be updated as well


collindutter:
Thanks for the PR @iloveitaly! We actually just released `v0.19.4` which locks `openai` at the proper version (`^0.28`) until https://github.com/griptape-ai/griptape/issues/420 is implemented.

PR #434, opened 2023-11-10
Remove ToolMemoryClient.off_prompt default value

vasinov:


README.md
```diff
@@ -102,7 +102,8 @@ And here is the output:
 ```
 
 During the run, the Griptape Agent loaded a webpage with a **Tool**, stored its full content in **Tool Memory**, queried it to answer the original question, and finally saved the answer to a file.
-The important thing to note here is that no matter how big the webpage is it can never blow up the prompt token limit because the full content never goes back to the main prompt.
+
+The important thing to note here is that no matter how big the webpage is it can never blow up the prompt token limit because the full content of the page never goes back to the main prompt. Additionally, no data from the subsequent subtasks were returned back to the prompt (all tools have the `off_prompt` property enabled be default; `ToolMemoryClient` requires the user to set it explicitly).
```
collindutter, line 106: Let's add a line on when the user would want to set `off_prompt=False`.

collindutter (commented):
Any docs changes required?

vasinov:
> Any docs changes required?

Yep, just submitted https://github.com/griptape-ai/griptape-docs/pull/150 :)

collindutter (approved):

PR #437, opened 2023-11-13
data-masking-hooks 

ashkaaar:
example usage:
```
from griptape.processors import PromptDriverPiiProcessor

def generate_random_name():
    return "John Doe"

def mask_pii_func(text):
    original_name = "Sam Altman"
    placeholder_name = generate_random_name()
    text = text.replace(original_name, placeholder_name)
    return text, {original_name: placeholder_name}

def unmask_pii_func(text, pii_replacements):
    for original_pii, masked_pii in pii_replacements.items():
        text = text.replace(masked_pii, original_pii)
    return text

processor = PromptDriverPiiProcessor(
    prompt_driver=OpenAiChatPromptDriver(
        api_key="", model="gpt-3.5-turbo"
    ),
    mask_pii_func=mask_pii_func,
    unmask_pii_func=unmask_pii_func,
)

prompt_stack = PromptStack()
prompt_stack.add_user_input("Summarize Sam Altman's user profile")

prompt_stack = processor.before_run(prompt_stack)

response = processor.prompt_driver.run(prompt_stack)

processed_response = processor.after_run(response)

print(f"Processed response: {processed_response}")
```


collindutter:
Awesome work @ashkaaar! I just updated the base branch to `dev`, can you please resolve the conflict that arose?

griptape/processors/amazon_comprehend_processor.py
```diff
@@ -0,0 +1,31 @@
+from .base_processors import BasePromptStackProcessor
+import boto3
+
+
+class AmazonComprehendPiiProcessor(BasePromptStackProcessor):
+    def __init__(self, comprehend_client=None, custom_filter_func=None):
```
collindutter, line 6: Should be using `attrs` `field`. Take at something like `BedrockTitanTokenizer` for how we integrate `boto3.Session` and other fields.

griptape/processors/base_processors.py
```diff
@@ -0,0 +1,23 @@
+class BasePromptStackProcessor:
```
collindutter, line 1: Needs to implement `ABC`.

griptape/processors/base_processors.py
```diff
@@ -0,0 +1,23 @@
...
+    def before_run(self, prompt):
```
collindutter, line 2: Add `@abstractmethod`

griptape/processors/base_processors.py
```diff
@@ -0,0 +1,23 @@
...
+        # Process the prompt before it is run
+        pass
```
collindutter, line 4: Use `...` syntax

griptape/processors/base_processors.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def after_run(self, result):
+        # Process the result after the prompt is run
+        pass
```
collindutter, line 8: Same here, `@abstractmethod` and `...`

griptape/processors/base_processors.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class BasePromptDriver:
```
collindutter, line 11: We've already got a [BasePromptDriver](https://github.com/griptape-ai/griptape/blob/main/griptape/drivers/prompt/base_prompt_driver.py), what is this?

griptape/processors/prompt_driver_processor.py
```diff
@@ -0,0 +1,24 @@
+from .base_processors import BasePromptStackProcessor
+
+
+class PromptDriverPiiProcessor(BasePromptStackProcessor):
+    def __init__(self, prompt_driver, mask_pii_func, unmask_pii_func):
+        self.prompt_driver = prompt_driver
+        self.mask_pii_func = mask_pii_func
+        self.unmask_pii_func = unmask_pii_func
```
collindutter, line 8: Use `attrs`

griptape/utils/prompt_stack.py
```diff
@@ -5,17 +5,22 @@
 
 if TYPE_CHECKING:
     from griptape.memory.structure import ConversationMemory
+    from griptape.processors.base_processors import BasePromptStackProcessor
 
 
 @define
 class PromptStack:
+    """A class representing a stack of prompts for a conversation."""
```
collindutter, line 13: Please rephrase with [this styleguide](https://google.github.io/styleguide/pyguide.html#384-classes) in mind.
> The class docstring should not repeat unnecessary information, such as that the class is a class.

griptape/utils/prompt_stack.py
```diff
@@ -5,17 +5,22 @@
...
+
     GENERIC_ROLE = "generic"
     USER_ROLE = "user"
     ASSISTANT_ROLE = "assistant"
     SYSTEM_ROLE = "system"
 
     @dataclass
     class Input:
+        """A class representing an input item in the prompt stack."""
```
collindutter, line 22: Same here.

tests/unit/processors/test_amazon_comprehend_processor.py
```diff
@@ -0,0 +1,34 @@
...
+        "Entities": [{"Text": "John Doe"}]
+    }
+    return AmazonComprehendPiiProcessor(
+        comprehend_client=comprehend_client,
+        custom_filter_func=custom_filter_func,
+    )
+
+
+def test_before_run(processor):
+    prompt_stack = [{"content": "This is John Doe"}]
```
collindutter, line 26: Can you please add some tests that have more than one PII being filtered out?

griptape/processors/amazon_comprehend_processor.py
```diff
@@ -0,0 +1,31 @@
...
+
+        # Default implementation using Amazon Comprehend
+        response = self.comprehend_client.detect_pii_entities(
+            Text=text, LanguageCode="en"
+        )
+        pii_entities = response.get("Entities", [])
+
+        # Replace PII entities with mask or redaction
+        for entity in pii_entities:
+            text = text.replace(entity["Text"], "[PII]")
```
collindutter, line 29: Let's move `[PII]` up to `BasePiiProcessor` as a new field `pii_replace_text`

griptape/processors/prompt_driver_processor.py
```diff
@@ -0,0 +1,24 @@
+from .base_processors import BasePromptStackProcessor
+
+
+class PromptDriverPiiProcessor(BasePromptStackProcessor):
```
collindutter, line 4: I think the idea behind `PromptDriverPiiProcessor` is to pass a prompt to a Prompt Driver that instructs the LLM to mask the text.

collindutter (changes requested):

vasinov (changes requested):
Please add type hints to all method signatures.

griptape/processors/prompt_driver_processor.py
collindutter: @vasinov we would need to do a Prompt Driver `run` call on each Input in the Prompt Stack, right? This could get slow/expensive..

vasinov (changes requested):
As far as I can tell, `before_run` and `after_run` are not invoked anywhere. They need to be invoked in prompt drivers individually since different LLM APIs work differently (e.g., use ChatML vs. single string).

griptape/processors/prompt_driver_processor.py
collindutter: One other approach could be to assemble the Prompt Stack into a prompt string, with clear separators between the Prompt Stack Inputs. Then we could make a single call to the LLM, and convert the response back to Prompt Stack Inputs.

Though this feels kind of hacky and might overwhelm LLMs that would otherwise be great at masking.

collindutter:
Thanks! Don't hesitate to reach out here or on Discord for any questions.

griptape/processors/amazon_comprehend_processor.py
ashkaaar: done, I also added some type hints for this class

griptape/processors/amazon_comprehend_processor.py
ashkaaar: Should I create a new `BasePiiProcessor` or did you mean `BasePromptStackProcessor`

griptape/processors/base_processors.py
ashkaaar: Done

griptape/processors/amazon_comprehend_processor.py
collindutter: Yes sorry, I meant `BasePromptStackProcessor`

griptape/processors/prompt_driver_processor.py
ashkaaar: another way could be to process the inputs in batches, rather than individually. This would reduce the number of function calls, which could potentially improve performance. Here's a rough idea of how this could be implemented.
```
def before_run(self, prompt_stack):
    batched_inputs = self.batch_inputs(prompt_stack["inputs"])
    for batch in batched_inputs:
        batch = self.mask_pii_func(batch)
    return prompt_stack

def after_run(self, prompt_stack):
    batched_inputs = self.batch_inputs(prompt_stack["inputs"])
    for batch in batched_inputs:
        batch = self.unmask_pii_func(batch)
    return prompt_stack 
```
Let me know what do you guys think

collindutter: Let's just get it working with the "slow" method for now and then we can optimize from there. Parallelization might be another approach if it comes down to it.

griptape/utils/prompt_stack.py
```diff
@@ -1,21 +1,27 @@
...
+from typing import Optional, TYPE_CHECKING, Any, List
+from dataclasses import dataclass, field
+from griptape.processors.base_processors import BasePromptStackProcessor
 
 if TYPE_CHECKING:
     from griptape.memory.structure import ConversationMemory
 
 
-@define
+@dataclass
```
vasinov, line 10: ?

griptape/processors/prompt_driver_processor.py
```diff
@@ -0,0 +1,26 @@
+from typing import Callable, Dict, List, Any
+from .base_processors import BasePromptStackProcessor
+import attr
+
+
+@attr.s
```
vasinov, line 6: Can you please use `attrs` syntax consistent with the rest of the framework? E.g., `define` for class and `field` for properties.

griptape/processors/base_processors.py
```diff
@@ -0,0 +1,16 @@
+from abc import ABC, abstractmethod
+from attr import define, field
+from typing import Any
+
+
+@define
+class BasePromptStackProcessor(ABC):
+    pii_replace_text: str = field(default="[PII]", kw_only=True)
```
vasinov, line 8: What is the meaning of this default? It looks like it's used in

```python
text.replace(entity["Text"], self.pii_replace_text)
```

which will just swap out allinstances of `entity["Text"]` in the text with `[PII]`. This is not the desired behavior.

griptape/processors/amazon_comprehend_processor.py
```diff
@@ -0,0 +1,33 @@
+from typing import Callable, Optional, List, Dict, Any, Union
+from attr import define, field, Factory
+from .base_processors import BasePromptStackProcessor
+import boto3
+from botocore.client import BaseClient
+
+
+@define
+class AmazonComprehendPiiProcessor(BasePromptStackProcessor):
+    comprehend_client: BaseClient = field(default=Factory(lambda: boto3.client("comprehend")), kw_only=True)
```
vasinov, line 10: Please use `session` to instantiate the client. See `AmazonBedrockPromptDriver` for reference.

griptape/processors/amazon_comprehend_processor.py
```diff
@@ -0,0 +1,33 @@
...
+    custom_filter_func: Optional[Callable[[str], str]] = field(default=None, kw_only=True)
+
+    def before_run(self, prompt_stack: Dict[str, List[Dict[str, Any]]]) -> Dict[str, List[Dict[str, Any]]]:
+        for input_item in prompt_stack["inputs"]:
+            input_item["content"] = self.filter_pii(input_item["content"])
+        return prompt_stack
+
+    def after_run(self, result: Any) -> Any:
+        ...
```
vasinov, line 19: Should `return None` with this type signature

griptape/processors/base_processors.py
```diff
@@ -0,0 +1,16 @@
...
+
+    @abstractmethod
+    def before_run(self, prompt: str) -> str:
+        ...
+
+    @abstractmethod
+    def after_run(self, result: Any) -> Any:
```
vasinov, line 15: Why `Any`?

griptape/processors/prompt_driver_processor.py
vasinov: > One other approach could be to assemble the Prompt Stack into a prompt string, with clear separators between the Prompt Stack Inputs.

We could convert it into a JSON object to make separation explicit. But parallelization is probably the better option.

griptape/processors/prompt_driver_processor.py
```diff
@@ -0,0 +1,26 @@
...
+class PromptDriverPiiProcessor(BasePromptStackProcessor):
+    prompt_driver = attr.ib()
+    mask_pii_func = attr.ib()
+    unmask_pii_func = attr.ib()
+
+    def before_run(self, prompt_stack: Dict[str, List[Dict[str, Any]]]) -> Dict[str, List[Dict[str, Any]]]:
+        for input_item in prompt_stack["inputs"]:
+            input_item["content"] = self.mask_pii_func(input_item["content"])
+        return prompt_stack
```
vasinov, line 15: This is inconsistent with the return type signature.

griptape/processors/prompt_driver_processor.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def after_run(self, prompt_stack: Dict[str, List[Dict[str, Any]]]) -> Dict[str, List[Dict[str, Any]]]:
+        for input_item in prompt_stack["inputs"]:
+            input_item["content"] = self.unmask_pii_func(input_item["content"])
+        return prompt_stack
```
vasinov, line 20: This is inconsistent with the return type signature.

tests/unit/processors/test_amazon_comprehend_processor.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@pytest.fixture
+def processor() -> AmazonComprehendPiiProcessor:
+    comprehend_client = MagicMock()
+    comprehend_client.detect_pii_entities.return_value = {"Entities": [{"Text": "John Doe"}, {"Text": "123 Main St"}]}
+    return AmazonComprehendPiiProcessor(comprehend_client=comprehend_client, custom_filter_func=custom_filter_func)
+
+
+def test_before_run_single_pii(processor: AmazonComprehendPiiProcessor) -> None:
```
vasinov, line 18: We don't use type signatures in tests.

vasinov (changes requested):
Thanks for the PR!

It doesn't seem like it would actually work. Have you tried testing it with Comprehend directly?

The point of this feature is to swap out all PII entries before they go out to the LLM provider and then put them back in once the LLM returns generated text. For example,

> Summarize Sam Altman's user profile: ...

"Sam Altman" here should be swapped out for something random like "John Doe" before it goes out to the LLM.

Once the LLM responds with something like

> Here is the summary of Jon Doe's profile: ...

We should swap out "John Doe" with "Sam Altman" again, so that the user sees the right PII entry (first and last name in this case). Does that make sense?

ashkaaar:
Thanks for the review @vasinov,  I am currently in the process of testing the feature using various approaches. I'll commit it with requested changes as soon as possible

ashkaaar:
I'll mark the PR ready to review when I am done with it

ashkaaar:
It's taking more time that I thoguht, I might need some more time. I'm also trying to get more familier with the codebase(apperciate the neatness by the way), thanks for the patience.

PR #442, opened 2023-11-14
Put action schema in template

collindutter:


collindutter:
Side discussion, should we hard fail if `ActionSubtask` returns `tool not found`? It seems like it's almost always a result of not enough Tools being provided. This would save the LLM from going on a hallucination streak of Tools.

griptape/tools/base_tool.py
```diff
@@ -83,6 +84,24 @@ def abs_file_path(self):
     def abs_dir_path(self):
         return os.path.dirname(self.abs_file_path)
 
+    def schema(self) -> dict:
```
vasinov, line 87: `@property`?

tests/unit/tools/test_email_client.py
collindutter, line 1: Weird test that relied on `_action.j2` for some reaosn.

vasinov (approved):

Issue #444, opened 2023-11-14
Enhancement: Would like to explicitly expose the encoding to be used by `open()` inside TextLoader


br-moriarty:
[x ] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
We would like to support collaborators working on different operating systems. Our source data is utf-8 encoded. For users on Windows, however, because `open` defaults to whatever encoding is returned by `locale.getencoding` we hit errors when characters in our source data are unknown to cp1252.

**Describe the solution you'd like**
It seems like a natural property to expose to expose on `TextLoader`.

**Describe alternatives you've considered**
If intention is to enforce utf-8, another option is to enforce it explicitly in the `open` call, as utf-8 is only default on some operating systems.

collindutter:
@br-moriarty this has been implemented in https://github.com/griptape-ai/griptape/pull/445 which should be released soon.

ashkaaar:
@collindutter that was super quick, impressive

PR #445, opened 2023-11-14
Add TextLoader encoding field

collindutter:
Closes https://github.com/griptape-ai/griptape/issues/444

### DIFF
--- a/griptape/loaders/text_loader.py
+++ b/griptape/loaders/text_loader.py
@@ -57,6 +57,7 @@ class TextLoader(BaseLoader):
                 chunk.generate_embedding(self.embedding_driver)
 
         for chunk in chunks:
+            chunk.encoding = self.encoding
             artifacts.append(chunk)
 
         return artifacts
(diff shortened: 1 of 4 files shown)


griptape/loaders/text_loader.py
```diff
@@ -41,7 +42,7 @@ def text_to_artifacts(self, text: str | Path) -> list[TextArtifact]:
         artifacts = []
 
         if isinstance(text, Path):
-            with open(text, "r") as file:
+            with open(text, "r", encoding=self.encoding) as file:
```
vasinov, line 45: Should we also integrate it into `TextArtifact` that we instantiate below? Similar to how we do it `BlobArtifact`.

vasinov (changes requested):

griptape/loaders/text_loader.py
collindutter: Would the field on `TextArtifact` just be for reference? Or would it enable some functionality like `to_bytes`?

vasinov: I think just for reference for now.

vasinov (approved):

PR #452, opened 2023-11-15
Rename TaskMemory directory

andrewfrench:
Renames the `/tool/` ToolMemory directory to `/task/` after ToolMemory was renamed to TaskMemory.


collindutter (approved):
Good catch

collindutter:
Ah should also do `templates/memory/tool.j2`

PR #453, opened 2023-11-16
Add check for empty pages

### LINKED ISSUE #403: WebLoader load call fails on URL with empty content
**Describe the bug**
The webloader fails on a web page that has valid html but no content. Can happen for certain pages. For example I passed in the URL for a page that had the following html - 
```
<html><head><meta http-equiv='refresh' content='0;url=/?_cookiecheck=1' /></head></html>
```

I was passing in multiple links using the `load_collection` call provided by the `WebLoader` and the entire call failed because of this one page.

collindutter:
Fixes #403

### DIFF
--- a/griptape/loaders/web_loader.py
+++ b/griptape/loaders/web_loader.py
@@ -35,6 +35,8 @@ class WebLoader(TextLoader):
         if page is None:
             raise Exception("can't access URL")
         else:
-            return json.loads(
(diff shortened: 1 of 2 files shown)


collindutter:
Yeah wasn't really sure how to add a test since it only happens on pages with no body.

andrewfrench (approved):

andrewfrench (approved):

PR #455, opened 2023-11-16
Remove ProxycurlClient tool

andrewfrench:
The ProxycurlClient tool has moved! You can now find the tool in its own repository [here](https://github.com/griptape-ai/proxycurl-client-tool). It can be imported into and used with a Griptape project just like before (check out examples/example.py).


collindutter:
Can you also please update docs?

collindutter (approved):

PR #459, opened 2023-11-17
Update README with contribution guidelines for new tools

andrewfrench:
New contributing guidelines outline our policy that new tools should be managed independently and not submitted to the core framework (unless we've explicitly asked for it in an issue).


collindutter (approved):
Love it!

PR #466, opened 2023-11-20
Support writing image to disk directly in ImageGenerator tool

andrewfrench:
Updates ImageGenerator tool to support writing an image artifact to disk directly from the tool itself. This is important because FileManager is not deterministic and it can be difficult to convince the LLM to save data where you want it. Offloading that functionality back to Python saves time and tokens.

### DIFF
--- a/griptape/tools/image_generator/tool.py
+++ b/griptape/tools/image_generator/tool.py
@@ -1,4 +1,8 @@
 from __future__ import annotations
+
+from os import path
+from typing import Optional
+
 from attrs import define, field
 from schema import Schema, Literal
 from griptape.artifacts import ErrorArtifact, ImageArtifact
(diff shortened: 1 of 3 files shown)


griptape/tools/image_generator/tool.py
```diff
@@ -9,7 +13,16 @@
 
 @define
 class ImageGenerator(BaseTool):
+    """ImageGenerator is a tool that can be used to generate an image.
+
+    Attributes:
+        image_generation_engine: The engine used to generate the image.
+        output_dir: If provided, the generated image will be written to disk in output_dir.
```
collindutter, line 20: Should we name this `workdir` for `FileManager` consistency?

andrewfrench: I can see the argument and don't strongly disagree, but in this case, I think `workdir` is less specific than `output_dir`. That makes sense for `FileManager`, but in this case, we're only outputting files and nothing else.

collindutter (approved):

PR #470, opened 2023-11-26
Specify image output file directly in image generation tool/task

andrewfrench:
Previously, the image generation tool and task accepted an output directory, but filenames were generated based on the artifact creation time. This allows a developer to specify the output filename directly, which helps when incorporating a generated image into the rest of a program.

### DIFF
--- a/griptape/tasks/image_generation_task.py
+++ b/griptape/tasks/image_generation_task.py
@@ -18,12 +18,14 @@ class ImageGenerationTask(BaseTextInputTask):
     Attributes:
         image_generation_engine: The engine used to generate the image.
         output_dir: If provided, the generated image will be written to disk in output_dir.
+        output_file: If provided, the generated image will be written to disk as output_file.
     """
 
     NEGATIVE_RULESET_NAME = "Negative Ruleset"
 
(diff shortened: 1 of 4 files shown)


griptape/tasks/image_generation_task.py
```diff
@@ -59,16 +77,20 @@ def run(self) -> ImageArtifact:
...
     def _write_to_file(self, image_artifact: ImageArtifact) -> None:
         # Save image to file. This is a temporary workaround until we update Task and Meta
         # Memory to persist artifacts from tasks.
-        outfile = path.join(self.output_dir, image_artifact.name)
+        if self.output_file:
+            outfile = self.output_file
+        else:
+            outfile = path.join(self.output_dir, image_artifact.name)
 
+        os.makedirs(self.output_dir, exist_ok=True)
```
collindutter, line 93: Should this be within the `else`? `output_dir` will be `None` if `output_file` is specified.

andrewfrench: Thanks! Updated to be `path.dirname(outfile)` to match the logic in tool.py

collindutter (approved):
🛳️

Issue #267, opened 2023-09-21
Add More Useful Logs


collindutter:
Developers want to debug their Griptape applications without having to add logs themselves. I propose adding `logging.debug` statements throughout the framework in critical areas like Drivers, and Tokenizers.

https://github.com/griptape-ai/griptape/issues/266

PR #478, opened 2023-12-05
Add prompt_stack and prompt to Prompt Events

collindutter:
These changes allow developers to access `prompt_stack` and `prompt` before it is sent to the LLM. This is useful when debugging applications and trying to determine what is being constructed by Griptape.

Partially implements #267


griptape/drivers/prompt/base_prompt_driver.py
collindutter, line 1: Unrelated, fixes missing type hint.

griptape/utils/stream.py
```diff
@@ -3,9 +3,11 @@
 from threading import Thread
 from queue import Queue
 from griptape.artifacts.text_artifact import TextArtifact
-from griptape.events import CompletionChunkEvent, EventListener
+from griptape.events.completion_chunk_event import CompletionChunkEvent
+from griptape.events.event_listener import EventListener
+from griptape.events.base_event import BaseEvent
+from griptape.events.finish_structure_run_event import FinishStructureRunEvent
```
collindutter, line 9: Resolves circular dependency issue :groan:

griptape/schemas/utils/prompt_stack_input_schema.py
andrewfrench, line 1: Neither here nor there in relation to your changes, but while it's on my mind: the `PromptStack` class seems pretty central to the framework at this point, should it still be in `utils`?

collindutter: I've been wondering the same thing, `utils` doesn't feel like like the right place for it. Any suggestions for where else it could go?

Issue #405, opened 2023-11-01
Add Stream functionality to the Chat utility


shhlife:
[x ] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Describe the solution you'd like**
Now that we have the Stream utility, it would be very helpful to add the functionality to the Chat utility. Preferably it would just be a parameter you pass.

`Chat(agent, stream=True).start()`

mattma1970:
Hey, I've been playing with the Streaming class a bit and would like to put this together.

shhlife:
I modified my initial request to focus on `stream=True`, so I replaced the method `run` with `start` (since that's what Chat uses), and also removed the initial message.

so instead of: `Chat(agent, stream=True).run("Howdy!")` it's `Chat(agent, stream=True).start()`

PR #481, opened 2023-12-05
Chat Streaming

collindutter:
Closing out @mattma1970's work.

Adds streaming support to the `Chat` utility. Also updates the messages from:
`Q: ` -> `User: `
`A: ` -> `Assistant: `
`processing...` -> `Thinking...`
`exiting...` -> `Exiting...`

Closes #409, #405


griptape/drivers/prompt/base_prompt_driver.py
```diff
@@ -68,6 +68,7 @@ def run(self, prompt_stack: PromptStack) -> TextArtifact:
                     for chunk in completion_chunks:
                         self.structure.publish_event(CompletionChunkEvent(token=chunk.value))
                         tokens.append(chunk.value)
+                    self.structure.publish_event(CompletionChunkEvent(token="\n"))
```
collindutter, line 71: Don't _love_ having to do this, but it's to resolve this issue:
image

Somehow we need to signify that we've sent the final completion chunk. This solution seems relatively unoffensive since users could just trim this off if they don't need it.

griptape/utils/chat.py
```diff
@@ -10,24 +11,30 @@
...
-    processing_text: str = field(default="processing...", kw_only=True)
+    exiting_text: str = field(default="Exiting...", kw_only=True)
+    processing_text: str = field(default="Thinking...", kw_only=True)
     intro_text: Optional[str] = field(default=None, kw_only=True)
-    prompt_prefix: str = field(default="Q: ", kw_only=True)
-    response_prefix: str = field(default="A: ", kw_only=True)
-    output_fn: Callable[[str], None] = field(default=Factory(lambda: print), kw_only=True)
+    prompt_prefix: str = field(default="User: ", kw_only=True)
+    response_prefix: str = field(default="Assistant: ", kw_only=True)
+    output_fn: Callable[[str], None] = field(default=lambda x: print(x, end="", flush=True), kw_only=True)
```
collindutter, line 19: Went back and forth on whether there should be two output functions. One for regular printing, one for stream printing. This feels kind of hacky, open to feedback.

andrewfrench (approved):

PR #482, opened 2023-12-06
Add simple tokenizer

collindutter:
This PR adds `SimpleTokenizer` which serves as an approximate Tokenizer for LLM providers that do not provide Tokenization APIs (Bedrock). The tokenizer allows for users to specify `max_tokens` and `characters_per_token`.


collindutter:
> is there a reason to give it a field tokenizer rather than just subclass SimpleTokenizer?

Good call

vachillo (approved):

Issue #425, opened 2023-11-07
Support for image input


derekpai:

derekpai:
need to generalize at the prompt driver level

related to #332

andrewfrench:
Closing as completed in #469

PR #494, opened 2023-12-12
Accept `TextArtifact` Input in `BaseTextInputTask`

collindutter:
Adds support for passing in `TextArtifact` for `BaseTextInputTask.input_template` either directly for through a Callable. Sets up #425 nicely, and enables use-cases like this:

```python
from typing import cast
from griptape.drivers import OpenAiDalleImageGenerationDriver
from griptape.tasks import ToolTask, ImageGenerationTask
from griptape.engines import ImageGenerationEngine


pipeline = Pipeline(
    prompt_driver=OpenAiChatPromptDriver(model="gpt-3.5-turbo"),
    tasks=[
        ToolTask(
            "Scrape https://griptape.ai",
            tool=WebScraper(),
            id="WebResults",
        ),
        ImageGenerationTask(
            lambda task: cast(
                TextArtifact,
                pipeline.task_memory.summarize_namespace(task.parents[0].output.name),
            ),
            id="ImageGeneration",
            image_generation_engine=ImageGenerationEngine(
                image_generation_driver=OpenAiDalleImageGenerationDriver(
                    model="dall-e-3"
                )
            ),
        ),
    ],
)

pipeline.run()
```

### DIFF
--- a/griptape/tasks/base_text_input_task.py
+++ b/griptape/tasks/base_text_input_task.py
@@ -15,7 +15,6 @@ class BaseTextInputTask(BaseTask, ABC):
     DEFAULT_RULESET_NAME = "Default Ruleset"
     ADDITIONAL_RULESET_NAME = "Additional Ruleset"
 
-    input_artifact_namespace: str | None = field(default=None, kw_only=True)
     input_template: str | TextArtifact | Callable[[BaseTask], TextArtifact] = field(default=DEFAULT_INPUT_TEMPLATE)
     context: dict[str, Any] = field(factory=dict, kw_only=True)
     rulesets: list[Ruleset] = field(factory=list, kw_only=True)
--- a/griptape/tasks/base_text_input_task.py
+++ b/griptape/tasks/base_text_input_task.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from abc import ABC
-from typing import Any, Optional, Callable
+from typing import Any, Callable
 
 from attr import define, field
 from griptape.artifacts import TextArtifact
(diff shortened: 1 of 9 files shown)


griptape/tasks/base_text_input_task.py
```diff
@@ -14,14 +15,19 @@ class BaseTextInputTask(BaseTask, ABC):
     DEFAULT_RULESET_NAME = "Default Ruleset"
     ADDITIONAL_RULESET_NAME = "Additional Ruleset"
 
-    input_template: str = field(default=DEFAULT_INPUT_TEMPLATE)
+    input_template: str | TextArtifact | Callable[[BaseTask], TextArtifact] = field(default=DEFAULT_INPUT_TEMPLATE)
```
collindutter, line 18: Should this be renamed to `_input`?

andrewfrench: I think so. We're only rendering a template in one of the three input cases so that name seems overly specific anymore

griptape/tasks/base_text_input_task.py
```diff
@@ -14,14 +15,19 @@ class BaseTextInputTask(BaseTask, ABC):
...
 
     @property
     def input(self) -> TextArtifact:
-        return TextArtifact(J2().render_from_string(self.input_template, **self.full_context))
+        if isinstance(self.input_template, TextArtifact):
+            return self.input_template
+        elif isinstance(self.input_template, Callable):
+            return self.input_template(self)
+        else:
+            return TextArtifact(J2().render_from_string(self.input_template, **self.full_context))
```
andrewfrench, line 30: I'm assuming this behaves just fine if the input isn't a template, but I wonder if there's an eventual improvement to be made by typing templates and distinguishing template inputs from raw string inputs, avoiding J2 initialization and template rendering step entirely.

andrewfrench (approved):

griptape/tasks/base_text_input_task.py
collindutter: If a user doesn't use any of the Jinja syntax, nothing will get rendered. I think the overhead of Jinja unnecessarily rendering is so minimal that it's not worth worrying about.

andrewfrench (approved):

PR #387, opened 2023-10-25
added sagemaker embedding driver

KaushikIyer16:
Added  AmazonSageMakerEmbeddingDriver to support SageMaker endpoints for generating embeddings.

I could not see any docs on what formatter is being used. I am happy to reformat the classes if I can get some info.

Closes #371

PR #498, opened 2023-12-13
SageMaker Embedding Driver

### LINKED ISSUE #371: Add `AmazonSageMakerEmbeddingDriver`
Add `AmazonSageMakerEmbeddingDriver` to support SageMaker endpoints for generating embeddings.

collindutter:
Closes #371, #387


griptape/drivers/embedding/sagemaker_embedding_driver.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+from typing import TYPE_CHECKING
+import json
+from typing import Any
+
+import boto3
```
andrewfrench, line 6: Elsewhere in the framework, like in bedrock_titan_embedding_driver.py, we conditionally `import boto3` under the `if TYPE_CHECKING` block like BaseEmbeddingModelDriver below, but it's not consistent. Should we do that here as well?

andrewfrench (commented):
One small comment

griptape/drivers/embedding/sagemaker_embedding_driver.py
collindutter: Good catch, that would've broken things. Also updated the non-typed import of boto3 to use `import_optional_dependency`.

andrewfrench (approved):

andrewfrench (approved):

griptape/drivers/embedding/base_embedding_driver.py
```diff
@@ -12,23 +12,25 @@
...
     """
     Attributes:
+        model: The name of the model to use.
         tokenizer: An instance of `BaseTokenizer` to use when calculating tokens.
-        chunker: An instance of `BaseChunker` to use when chunking long strings.
     """
 
-    tokenizer: BaseTokenizer = field(kw_only=True)
+    model: str = field(kw_only=True)
+    tokenizer: BaseTokenizer | None = field(kw_only=True)
```
vasinov, line 20: 1. `Optional[BaseTokenizer]`
2. Should it have `None` as the default?

griptape/drivers/embedding/base_embedding_driver.py
```diff
@@ -12,23 +12,25 @@
...
         return self.embed_string(artifact.to_text())
 
     def embed_string(self, string: str) -> list[float]:
         for attempt in self.retrying():
             with attempt:
-                if self.tokenizer.count_tokens(string) > self.tokenizer.max_tokens:
+                if self.tokenizer and self.tokenizer.count_tokens(string) > self.tokenizer.max_tokens:
                     return self._embed_long_string(string)
                 else:
                     return self.try_embed_chunk(string)
```
vasinov, line 36: Is this meaningful behavior if the tokenizer is not specified? What use case does it enable?

vasinov (changes requested):

griptape/drivers/embedding/base_embedding_driver.py
collindutter: A Tokenizer is only needed when embedding long strings that need to be chunked up. Some embedding models do not provide tokenization information (max tokens, tokenization algorithm, etc) so there is no great way to build a tokenizer. By making it optional, we hand the reigns to the developer to embed strings of a length they determine.

griptape/drivers/embedding/base_embedding_driver.py
collindutter: 1. This change comes from pyupgrade: https://github.com/asottile/pyupgrade?tab=readme-ov-file#pep-604-typing-rewrites
2. Yep.

vasinov (approved):

PR #515, opened 2023-12-19
Add pyright pre-commit hook

collindutter:


andrewfrench:
Do we want to add this to code checks as well on the Actions side?

collindutter:
Only challenge with that is that it'll run on the entire repo, not just changed files. Let me see if I can tweak it to only run on the diff, similar to how pre-commit works.

andrewfrench (approved):

PR #520, opened 2023-12-21
Add MediaArtifact

andrewfrench:
Adds a MediaArtifact class. ImageArtifact as well as future types inherit from a common MediaArtifact to capture common fields and functionality like artifact naming and MIME type support.

### DIFF
--- a/griptape/artifacts/base_media_artifact.py
+++ b/griptape/artifacts/base_media_artifact.py
@@ -1,0 +1,51 @@
+from __future__ import annotations
+
+from abc import abstractmethod, ABC
+
+import string
+import time
+import random
+from typing import Optional
+
+from attr import define, field, Factory
+from griptape.artifacts import BlobArtifact
+import base64
+
+
+@define
+class BaseMediaArtifact(BlobArtifact, ABC):
+    """MediaArtifact is a type of BlobArtifact that represents media (image, audio, video, etc.).
+
+    Attributes:
+        value: Raw bytes representing the media.
+        name: Artifact name, generated using creation time and a random string.
+        mime_type: The mime type of the image, like image/png or audio/wav.
+        model: Optionally specify the model used to generate the media.
+        prompt: Optionally specify the prompt used to generate the media.
+    """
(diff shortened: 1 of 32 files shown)


collindutter:
@andrewfrench would you like to keep this PR open?

andrewfrench:
Yes, I'll dust this off and we can get it in. It's still useful as we expand to other media artifact types.

griptape/artifacts/base_media_artifact.py
```diff
@@ -0,0 +1,54 @@
...
+
+@define
+class MediaArtifact(BlobArtifact):
+    """MediaArtifact is a type of BlobArtifact that represents media (image, audio, video, etc.)
+    and can be extended to support a specific media type.
+
+    Attributes:
+        value: Raw bytes representing media data.
+        name: Artifact name, generated using creation time and a random string.
+        mime_type: The mime type of the image, like image/png or audio/wav.
```
collindutter, line 23: Docstring appears a bit out of sync with actual fields.

griptape/artifacts/base_media_artifact.py
```diff
@@ -0,0 +1,54 @@
...
+        model: Optionally specify the model used to generate the media.
+        prompt: Optionally specify the prompt used to generate the media.
+    """
+
+    artifact_type: str = field(default="media", kw_only=True, metadata={"serializable": True})
+    format: str = field(kw_only=True, metadata={"serializable": True})
```
collindutter, line 29: Do these need to be split out like this? Do they serve any purpose other than building `mime_type`? I think `artifact_type` is a bit confusing because it's really the media type.

andrewfrench: These attributes contribute to `mime_type` as well as artifact name. Independent `format` metadata in particular is useful for things like enforcing format limitations on model inputs.

collindutter (commented):
Looks great, can you please update the changelog?

collindutter (approved):

Issue #519, opened 2023-12-20
Resolve Pyright errors


andrewfrench:
We're using [Pyright](https://github.com/microsoft/pyright) as a static type checker for our Python code. The rules enforced by Pyright are much stricter than those typically enforced natively by an IDE like Visual Studio, PyCharm, or other tools, so a large portion of the framework does not meet Pyright's standard. Pyright currently reports about 160 errors on our `dev` branch that we should resolve before we're able to enforce the standard on new PRs with code checks and pre-commit hooks.

- [x] #577
- [x] chunkers/
- [x] #578
- [x] #579
- [x] #580
- [x] #581
- [x] #582
- [x] #583
- [x] #584
- [x] #585
- [x] #586
- [x] mixins/
- [x] rules/
- [x] schemas/
- [x] structures/
- [x] summarizers/
- [x] #576
- [x] templates/
- [x] tokenizers/
- [x] tools/
- [x] utils/

derekpai:
break this issue down into multiple tasks

PR #528, opened 2023-12-21
Add overloads to tokenizers, fix type errors

collindutter:
#519


andrewfrench (dismissed):

griptape/tokenizers/base_tokenizer.py
```diff
@@ -21,6 +18,16 @@ def count_tokens_left(self, text: str | list) -> int:
...
+
+    @overload
+    @abstractmethod
+    def count_tokens(self, text: list[dict]) -> int:
+        ...
+
     @abstractmethod
-    def count_tokens(self, text: str | list) -> int:
+    def count_tokens(self, text: str | list[dict]) -> int:
         ...
```
vasinov, line 33: Why do we need both here?

collindutter: Both as in `str` and `list[dict]`? `OpenAiTokenizer` uses a list.

vasinov: Both as in a method with a union type hint _and_ an overload. Is it to make the type checker happy?

collindutter: Ah I see, I can remove the union type hint; is overloading even worth doing here? Originally I had done it to avoid the `isinstance` checks in the subclass methods but I think I had misunderstood python's overloading capabilities.

vasinov: I think it can just be `def count_tokens(self, text: str | list[dict])` without the `@overload` method.

collindutter: Yeah that's what it was originally. Adding `@overload` does give the benefit of multiple method signatures for things like auto complete, but maybe not worth the fork in consistency.

vasinov: Oh, I see! That's a good point about auto complete. Let's have an internal discussion about this since there are multiple places that this will touch.

vasinov (approved):

PR #532, opened 2023-12-23
Fix Llama Prompt, Improve Tool Usage

collindutter:


tests/utils/structure_tester.py
```diff
@@ -79,20 +77,23 @@ class StructureTester:
...
+            model="meta.llama2-70b-chat-v1", prompt_model_driver=BedrockLlamaPromptModelDriver(), max_attempts=1
         ),
+        # "SAGEMAKER_LLAMA_7B": AmazonSageMakerPromptDriver(
+        #     model=os.environ["SAGEMAKER_LLAMA_ENDPOINT_NAME"],
+        #     prompt_model_driver=SageMakerLlamaPromptModelDriver(max_tokens=4096),
+        # ),
+        # "SAGEMAKER_FALCON_7b": AmazonSageMakerPromptDriver(
+        #     model=os.environ["SAGEMAKER_FALCON_ENDPOINT_NAME"],
+        #     prompt_model_driver=SageMakerFalconPromptModelDriver(),
+        # ),
```
collindutter, line 93: Is there a better way to keep these around? Don't actively keep the infrastructure running.

griptape/drivers/prompt_model/bedrock_llama_prompt_model_driver.py
```diff
@@ -57,22 +57,25 @@ def prompt_stack_to_model_input(self, prompt_stack: PromptStack) -> str:
...
                 prompt_lines.append(f"<s>[INST] <<SYS>>\n{first.content}\n<</SYS>>\n\n")
-                if second and second.is_user():
-                    prompt_lines.append(f"{second.content} [/INST]")
-                else:
-                    raise Exception("System input must be followed by user input.")
+                if second:
+                    if second.is_user():
+                        prompt_lines.append(f"{second.content} [/INST]")
+                    else:
+                        raise Exception("System input must be followed by user input.")
```
andrewfrench, line 64: Doesn't the new nesting break this? If second is None the system input isn't followed by user input but we don't raise the exception. Or perhaps is it that `second` is optional and if it's present it must be user input?

griptape/drivers/prompt_model/bedrock_llama_prompt_model_driver.py
```diff
@@ -57,22 +57,25 @@ def prompt_stack_to_model_input(self, prompt_stack: PromptStack) -> str:
...
                 prompt_lines.append(f"<s>[INST] {first.content} [/INST]")
-                if second and second.is_assistant():
-                    prompt_lines.append(f" {second.content} </s>")
-                else:
-                    raise Exception("User input must be followed by assistant input.")
+                if second:
+                    if second.is_assistant():
+                        prompt_lines.append(f" {second.content} </s>")
+                    else:
+                        raise Exception("User input must be followed by assistant input.")
```
andrewfrench, line 78: Same points re: nested conditionals here and on line 71.

griptape/drivers/prompt_model/bedrock_llama_prompt_model_driver.py
vasinov: If it is indeed optional, it might be a good idea to add a type hint on line 56 to indicate that, so the code is easier to reason about.

griptape/templates/tasks/tool_task/system.j2
```diff
@@ -1,7 +1,6 @@
-Your output should always be an action in the form of a JSON object with the following schema: 
-{{ action_schema }}
+Output a JSON object that strictly adheres to the specified schema with no additional text or commentary. The object should only include fields defined in the schema
```
vasinov, line 1: Have you tried using "raw JSON object" language?

tests/utils/structure_tester.py
vasinov: We could add an object in the form of:

```python
class TesterPromptDriverOption:
    prompt_driver: BasePromptDriver
    enabled: bool
```

and then set `enabled` to `False` for cases that we don't want to be testing all the time. This will also allow us to run it "with defaults" and "force run all."

griptape/templates/tasks/tool_task/system.j2
collindutter: I had to change the wording quite a bit in order to get llama 13b happy. With the old wording (and "raw"):
```python
tests/integration/tasks/test_tool_task.py::TestToolTask::test_tool_task[AmazonBedrockPromptDriver-meta.llama2-13b-chat-v1] [01/05/24 12:15:18] INFO     ToolTask 0e5dbba866064f0b9a758fcaa7467711
                             Input: What is 7 times 3 divided by 5 plus 10.
[01/05/24 12:15:27] ERROR    Subtask 0e5dbba866064f0b9a758fcaa7467711
                             Error parsing tool action: Extra data: line 37 column 1 (char 781)
                    INFO     Subtask 606f804783074224b5d6be822b94036a
                             Action: Here is the calculation you requested as a JSON object:

                             {
                             "description": "Calculator action schema.",
                             "type": "object",
                             "properties": {
                             "name": {
                             "const": "Calculator"
                             },
                             "path": {
                             "description": "Can be used for making simple numeric or algebraic calculations in Python.",
                             "const": "calculate"
                             },
                             "input": {
                             "type": "object",
                             "properties": {
                             "values": {
                             "type": "object",
                             "properties": {
                             "expression": {
                             "description": "Arithmetic expression parsable in pure Python. Single line only. Don't use any imports or external libraries",
                             "type": "string"
                             }
                             },
                             "required": ["expression"],
                             "additionalProperties": false
                             }
                             },
                             "required": ["values"],
                             "additionalProperties": false
                             }
                             },
                             "required": ["name", "path", "input"],
                             "additionalProperties": false,
                             "$id": "Calculator Action Schema",
                             "$schema": "http://json-schema.org/draft-07/schema#"
                             }

                             And the answer to your calculation is:

                             {
                             "expression": "7 * 3 / 5 + 10",
                             "values": {
                             "expression": "7 * 3"
                             }
                             }
                    INFO     Subtask 606f804783074224b5d6be822b94036a
                             Response: {'error': 'Action input parsing error: Extra data: line 37 column 1 (char 781)'}
                    INFO     ToolTask 0e5dbba866064f0b9a758fcaa7467711
                             Output: {'error': 'Action input parsing error: Extra data: line 37 column 1 (char 781)'}
```

This new wording works with all Drivers (except Titan).

tests/utils/structure_tester.py
```diff
@@ -26,88 +28,155 @@
 )
 
 
+def get_enabled_prompt_drivers(prompt_drivers_options) -> list[BasePromptDriver]:
```
collindutter, line 31: Is there a better place to put this? Can't put it in the class as a `@staticmethod` because `StructureTester` does not exist yet. Putting it out here prevents me from using the `TesterPromptDriverOption` type hint.

I want to keep the fields static.

andrewfrench: would it make sense as a class method?

collindutter: Same issue: `"StructureTester" is not defined`

andrewfrench (approved):

Issue #530, opened 2023-12-22
Support deletion in vector store drivers


andrewfrench:
Relevant [Discord thread](https://discord.com/channels/1096466116672487547/1187737692033785858)

Griptape vector store drivers don't currently support deleting vectors from the vector store. Add a `delete` method to complete CRUD support.

PR #533, opened 2023-12-28
Add Delete Function to Vector Store Drivers

bjorger:
Partially implements #530 

Add Delete Function to All Classes That Inherit from the Base Vectorstore Class
Add Delete Function to Base Vectorstore Class
Implement Delete Function for MongoDB Vectorstore
Implement Delete Collection Function for MongoDB Vectorstore
Add Test for Delete Function for MongoDB Vectorstore


pyproject.toml
```diff
@@ -37,7 +37,10 @@ mail-parser = "^3.15.0"
 requests = "^2"
 lxml = "^4.9.3" # TODO: remove when https://github.com/adbar/trafilatura/issues/449 is fully resolved
 pillow = "^10.1.0"
-
+pytest = "^7.4.3"
+mongomock = "^4.1.2"
+pyright = "^1.1.343"
+google = "^3.0.0"
```
collindutter, line 43: Were these changes intentional? Some of them look like they belong in the `test` poetry group.

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -72,6 +72,9 @@ def upsert_text(
             **kwargs
         )
 
+    def delete(self):
+        raise NotImplementedError("This Method is not implemented yet")
```
collindutter, line 76: Let's update the message to match other `NotImplementedError`s in the codebase. Something like:
`raise NotImplementedError(f"{self.__class__.__name__} does not support deletion.")`

collindutter (changes requested):

pyproject.toml
bjorger: Pytest, mongomock and pyright should not be there, you are correct!

But for google, I received an error in the testing pipeline for python 3.11 (or 3.12) that the dependency was missing, so I don't know if this should be in the testing dependencies or in the normal ones!

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -72,6 +72,10 @@ def upsert_text(
             **kwargs
         )
 
+    @abstractmethod
+    def delete(self, vector_id: str):
```
collindutter, line 76: Let's rename to `delete_vector` for consistency with the other methods.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ mail-parser = "^3.15.0"
 requests = "^2"
 lxml = "^4.9.3" # TODO: remove when https://github.com/adbar/trafilatura/issues/449 is fully resolved
 pillow = "^10.1.0"
+google = "^3.0.0"
```
collindutter, line 40: Is this change intentional?

pyproject.toml
```diff
@@ -106,6 +107,8 @@ optional = true
 pytest = "~=7.1"
 pytest-mock = "*"
 mongomock = "*"
+pyright = "^1.1.343"
+
```
collindutter, line 111: Is this change intentional?

collindutter (changes requested):

pyproject.toml
bjorger: As stated above, without this dependency the python 3.11 pipeline failed for me - but I will remove the dependency and see if the pipeline runs successfully

collindutter: Ah sorry missed the previous review comment! Yeah let's try again without it, I can help troubleshoot if it still gives issues.

collindutter:
@bjorger FYI don't worry about the `type-check` action.

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -72,6 +72,10 @@ def upsert_text(
             **kwargs
         )
 
+    @abstractmethod
+    def delete_vector(self, vector_id: str):
```
collindutter, line 76: We should add an explicit return type of `None`.

pyproject.toml
```diff
@@ -107,6 +107,7 @@ pytest = "~=7.1"
 pytest-mock = "*"
 mongomock = "*"
 
+
```
collindutter, line 110: Extra line, let's remove to leave `pyproject.toml` out of the changed files.

collindutter (changes requested):

collindutter (dismissed):

griptape/drivers/vector/mongodb_vector_store_driver.py
```diff
@@ -135,3 +135,12 @@ def query(
...
         return results
+
+    def delete_vector(self, vector_id: str):
+        """Deletes the vector from the collection."""
+        collection = self.get_collection()
+        collection.delete_one({"_id": vector_id})
+
+    def delete_collection(self):
+        """Deletes the current collection from the database."""
+        self.get_collection().drop()
```
vasinov, line 146: Should the collection really be deleted here? If so, let's not move it into a separate method since it's a one-liner not used anywhere else.

vasinov (dismissed):

griptape/drivers/vector/mongodb_vector_store_driver.py
bjorger: I don't fully understand.
First of all, yes, the objective of this function is to delete the collection.
But where should I move it?

vasinov: Sorry, I misread the code. I don't think we need this method at all. The purpose of the driver is to handle vectors inside of a vector DB, not the DB itself.

bjorger: I, for once, need that functionality with our current use case, where we create embeddings of large documents and store them in individual collections.
Also, check out the screenshot attached. When you upload a big document as an embedding, it makes a bunch of vectors for each embedding. So, this is where the 'delete_collection' function comes in handy
image

vasinov: Your use case makes perfect sense but we are trying to keep all drivers pretty generic.

In this particular scenario, the user could chose between any number of ways to store vectors (e.g., single collection, collection per artifact, collection per artifact grouping, etc.), which could result in different `delete_*` functionality. We are also, in general, trying to reduce the number of one-off methods in drivers that could result in inconsistent UX

In your use case, would it make more sense to just use the Mongo SDK to delete the collection instead of moving that functionality into Griptape?

bjorger: I understand the UX aspect, but Griptape also creates the collection, therefore, it should provide a functionality to delete them as well, without going the way of using pymongo or any other SDK. 

Also, maybe I misunderstood you there, but there is no "out of the box" solution from griptape to store the vectors in the fashion that you just described. 
Furthermore, as I said above, when creating an embedding from a large document, you cannot store the entire thing in one vector, there will be several vectors created for that embedding, therefore, the delete_vector function would probably result in some weird work arounds to properly delete the entire embedding.

But this is my opinion - if you guys agree that the delete_collection function is not of use for griptape, I will remove it from the PR.

bjorger (commented):
Work in review

collindutter (approved):

PR #537, opened 2024-01-02
Support Other Tokenizers With `OpenAiChatPromptDriver`

### LINKED ISSUE #512: Together AI support

Ability to use TogetherAI with griptape

**Additional context**
I am a big fan of TogetherAI, I use it often, they offer a drop in open ai api replacement, but at the moment it doesn't seem to work

```

agent = Agent(
    prompt_driver=OpenAiChatPromptDriver(
        api_key=os.environ["OPENAI_API_KEY"],
        base_url=os.environ["OPENAI_BASE_URL"],
        model="mistralai/Mixtral-8x7B-Instruct-v0.1",
        max_tokens=1024
    )
)
agent.run("Return the height and width of an average grown plant in cm, rounded to the closest 5cm: dianthus picotee mix")
```
When running it I am getting back the following error

```
Traceback (most recent call last):
  File "c:\projects\rl\app.py", line 18, in <module>
    agent = Agent(
            ^^^^^^
  File "<attrs generated init griptape.structures.agent.Agent>", line 40, in __init__
  File "C:\anaconda3\Lib\site-packages\griptape\structures\structure.py", line 53, in <lambda>
    default=Factory(lambda self: self.default_task_memory, takes_self=True), kw_only=True
                                 ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\anaconda3\Lib\site-packages\griptape\structures\structure.py", line 126, in default_task_memory
    summary_engine=PromptSummaryEngine(prompt_driver=self.prompt_driver),
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<attrs generated init griptape.engines.summary.prompt_summary_engine.PromptSummaryEngine>", line 15, in __init__
  File "C:\anaconda3\Lib\site-packages\griptape\engines\summary\prompt_summary_engine.py", line 24, in <lambda>
    lambda self: TextChunker(tokenizer=self.prompt_driver.tokenizer, max_tokens=self.max_chunker_tokens),
                                                                                ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\anaconda3\Lib\site-packages\griptape\engines\summary\prompt_summary_engine.py", line 39, in max_chunker_tokens
    return round(self.prompt_driver.tokenizer.max_tokens * self.max_token_multiplier)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\anaconda3\Lib\site-packages\griptape\tokenizers\openai_tokenizer.py", line 47, in max_tokens
    tokens = next(v for k, v in self.MODEL_PREFIXES_TO_MAX_TOKENS.items() if self.model.startswith(k))
```

I am wondering if this would be something considered in the future?

Thank you

collindutter:
Closes #512

Enables the use of other Tokenizers when using services like TogetherAi that use OpenAI's SDK.
```python
import os
from griptape.tokenizers import HuggingFaceTokenizer
from transformers import AutoTokenizer

agent = Agent(
    prompt_driver=OpenAiChatPromptDriver(
        api_key=os.environ["TOGETHERAI_API_KEY"],
        base_url=os.environ["TOGETHERAI_BASE_URL"],
        model="mistralai/Mixtral-8x7B-Instruct-v0.1",
        tokenizer=HuggingFaceTokenizer(
            tokenizer=AutoTokenizer.from_pretrained(
                "mistralai/Mixtral-8x7B-Instruct-v0.1"
            )
        ),
    )
)
agent.run(
    "Return the height and width of an average grown plant in cm, rounded to the closest 5cm: dianthus picotee mix"
)
```


collindutter:
Contains changes from #528 that should be merged in first.

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -43,7 +43,7 @@ class OpenAiChatPromptDriver(BasePromptDriver):
         )
     )
     model: str = field(kw_only=True)
-    tokenizer: OpenAiTokenizer = field(
+    tokenizer: BaseTokenizer = field(
```
vasinov, line 46: Is there an actual use case for when we need to support other tokenizers here?

collindutter: The use-case on #512 requires it, I think. The alternative is that we add support for arbitrary models to `OpenAiTokenizer`.

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -119,7 +122,10 @@ def _base_params(self, prompt_stack: PromptStack) -> dict:
         # A max_tokens parameter is not required, but if it is specified by the caller, bound it to
         # the maximum value as determined by the tokenizer and pass it to the API.
         if self.max_tokens:
-            params["max_tokens"] = self.max_output_tokens(messages)
+            if isinstance(self.tokenizer, OpenAiTokenizer):
+                params["max_tokens"] = self.max_output_tokens(messages)
+            else:
+                params["max_tokens"] = self.max_output_tokens(self.prompt_stack_to_string(prompt_stack))
```
collindutter, line 128: Should this logic even exist? Seems like if a user specifies `max_tokens` we should pass that exact value. Otherwise we leave it empty and trust that the LLM provider will generate an appropriate amount of tokens.

griptape/drivers/prompt/openai_chat_prompt_driver.py
vasinov: I think this is confusing. Seems like `OpenAi*` should be focused on _OpenAI-only_ things. But we could add another more generic driver and tokenizer to support _OpenAI-like_ things.

griptape/drivers/prompt/openai_chat_prompt_driver.py
vasinov: I think it works for OpenAI APIs but not for some others. In this particular case, we can probably drop it. I actually thought that @andrewfrench worked on this at some point to deal with GPT-4 rate limiting issues but I might be wrong.

griptape/drivers/prompt/openai_chat_prompt_driver.py
collindutter: Agreed that it is kind of confusing, but there is a trend of using OpenAi's SDK as the SDK for non-OpenAi models. And we already partially support this functionality by allowing users to set the `base_url` field. So we could create a Driver, but it would still be tightly coupled to OpenAi.

At the very least I think we need more graceful error handling in the `OpenAiTokenizer` (maybe all Tokenizers?) so it doesn't error out when people set `model` to a non-OpenAi model.

griptape/drivers/prompt/openai_chat_prompt_driver.py
collindutter: The work @andrewfrench did was to add the `if max_tokens` check. But agreed, I think we should remove the use of `max_output_tokens` in this case.

griptape/drivers/prompt/openai_chat_prompt_driver.py
vasinov: I see...that makes sense. I think what we should fix then is the conditional code below (e.g., `if isinstance()...` by moving relevant parts into tokenizers.

vasinov (changes requested):

griptape/drivers/prompt/openai_chat_prompt_driver.py
collindutter: If I'm understanding correctly, that would require us to handle the result of `_prompt_stack_to_messages(prompt_stack)` (message list) in other Tokenizers which would not be straightforward. For `OpenAiTokenizer` we must have the message list, and for all other Tokenizers we must have a string.

Another potential avenue is to allow Tokenizers to take in a Prompt Stack. But the logic for converting the Prompt Stack into a string only exits in the Prompt Drivers.

griptape/drivers/prompt/openai_chat_prompt_driver.py
collindutter: Removed

andrewfrench (approved):
@vasinov's concerns above are valid and that conversation remains unresolved, but approving as this is a minimal solution to the problem described.

PR #538, opened 2024-01-04
Refactor task rule behavior to RuleMixin

andrewfrench:
A precursor PR to an incoming refactor of image generation tasks.

Refactoring rule and ruleset behavior to the RuleMixin allows image generation tasks to inherit them without having to inherit from BaseTextInputTask, inappropriately overloading the `input` methods (as per Pyright).

There are a couple bits of maybe controversial code in here that I'll call out in the comments.


griptape/tasks/base_rule_task.py
```diff
@@ -0,0 +1,69 @@
...
+from typing import Any
+
+from attr import define, field
+
+from griptape.rules import Ruleset, Rule
+from griptape.tasks import BaseTask
+
+
+@define
+class BaseRuleTask(BaseTask, ABC):
```
andrewfrench, line 13: The content of BaseRuleTask is copied directly from BaseTextInputTask, no new functionality here.

griptape/tasks/base_rule_task.py
```diff
@@ -0,0 +1,69 @@
...
+    @property
+    def full_context(self) -> dict[str, Any]:
+        if self.structure:
+            structure_context = self.structure.context(self)
+
+            structure_context.update(self.context)
+
+            return structure_context
+        else:
+            return {}
```
andrewfrench, line 69: Context-related code is relevant to image generation tasks as well, so I've moved it to BaseRuleTask. This dulls the precision of the naming a bit, but it could be overkill to introduce a BaseContextTask. Maybe context code should be promoted to BaseTask?

griptape/tasks/base_task.py
```diff
@@ -31,7 +34,7 @@ class State(Enum):
 
     @property
     @abstractmethod
-    def input(self) -> BaseArtifact:
+    def input(self) -> BaseArtifact | tuple[BaseArtifact, ...]:
```
andrewfrench, line 37: Descendent tasks are no longer guaranteed to return a single artifact here (see ImageVariationTask vs. ExtractionTask). Instead of widening this to `Any`, I've covered all our current bases by indicating either an artifact or a tuple of artifacts.

griptape/tasks/text_to_image_task.py
```diff
@@ -25,6 +25,10 @@ class TextToImageTask(BaseImageGenerationTask):
         output_file: If provided, the generated image will be written to disk as output_file.
     """
 
+    DEFAULT_INPUT_TEMPLATE = "{{ args[0] }}"
+
+    _input: str | TextArtifact | Callable[[BaseTask], TextArtifact] = field(default=DEFAULT_INPUT_TEMPLATE)
```
andrewfrench, line 30: There's some duplication between TextToImageTask and BaseTextInputTask to satisfy Pyright's clean inheritance standards. Could inputs be refactored to mixins to prevent this duplication?

griptape/tasks/base_task.py
```diff
@@ -107,3 +111,14 @@ def reset(self) -> BaseTask:
...
+    @property
+    def full_context(self) -> dict[str, Any]:
+        if self.structure:
+            structure_context = self.structure.context(self)
+
+            structure_context.update(self.context)
+
+            return structure_context
+        else:
+            return {}
```
andrewfrench, line 124: Context code is generally applicable, so moved to BaseTask.

griptape/mixins/rule_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@define(slots=False)
+class RuleMixin:
+    DEFAULT_RULESET_NAME = "Default Ruleset"
+    ADDITIONAL_RULESET_NAME = "Additional Ruleset"
+
+    rulesets: list[Ruleset] = field(factory=list, kw_only=True)
+    rules: list[Rule] = field(factory=list, kw_only=True)
+    structure: Any = field(default=None, kw_only=True)
```
andrewfrench, line 17: `structure` needs to be defined in this class for Pyright to allow us to check for `self.structure` below. We can default to None, and if the class using the mixin has a structure, it will be overridden, allowing the Task to access its structure's rules (confirmed experimentally and via unit tests below)

griptape/tasks/base_task.py
SavagePencil: when would we use tuples vs. ListArtifact? Do we eventually need a DictArtifact?

andrewfrench: @collindutter that's right, the task itself returns a single artifact. This was in reference to the input method specifically as its signature changes depending on the task.

andrewfrench: @SavagePencil tuples are used here vs. list artifacts due to the behavior of the tasks that accept multiple inputs, like, for example, ImageVariationTask. This accepts a specific number and type of artifacts as input. By instead accepting a ListArtifact we'd lose out on the explicit typing of the inputs.

tests/mocks/mock_image_generation_task.py
```diff
@@ -6,5 +6,7 @@
 
 @define
 class MockImageGenerationTask(BaseImageGenerationTask):
+    input: str = field(default="input")  # pyright: ignore
```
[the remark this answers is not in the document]
andrewfrench: Simple override of BaseTask's input

griptape/mixins/rule_mixin.py
andrewfrench: That raised instinctual circular import alarms to me but it checks out! Updated.

griptape/tasks/base_image_generation_task.py
```diff
@@ -82,10 +83,12 @@ def _read_from_file(self, path: str) -> ImageArtifact:
...
         # Save image to file. This is a temporary workaround until we update Task and Meta
         # Memory to persist artifacts from tasks.
-        if self.output_file is not None:
+        if self.output_file:
             outfile = self.output_file
-        else:
+        elif self.output_dir:
             outfile = os.path.join(self.output_dir, image_artifact.name)
+        else:
+            raise ValueError("No output_file or output_dir specified.")
```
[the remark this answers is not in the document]
andrewfrench: Not entirely related, but yes, a small change required to bring Pyright complaints to zero for this PR.

griptape/tasks/text_to_image_task.py
andrewfrench: In the past, yes, but this no longer inherits from BaseTextInputTask so DEFAULT_INPUT_TEMPLATE doesn't exist in any of its parents/ancestors.

tests/mocks/mock_image_generation_task.py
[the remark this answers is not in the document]
collindutter: What is the type error?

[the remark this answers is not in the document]
andrewfrench: `griptape/tests/mocks/mock_image_generation_task.py:9:5 - error: "input" incorrectly overrides property of same name in class "BaseTask" (reportIncompatibleMethodOverride)`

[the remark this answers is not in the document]
collindutter: Ah because it's a `@property` in the parent. I think this warning is valid then; it should be a property in the subclass.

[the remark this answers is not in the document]
andrewfrench: Fair; updated.

collindutter (approved):

PR #539, opened 2024-01-04
Update makefile

andrewfrench:
Screenshot 2024-01-03 at 11 01 57 PM

- Add Makefile help text
- Requires `v` to be set when bumping release version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 .PHONY: version
-version: ## Bump version and push to release branch. ex: make version v=0.1.0
+version: ## Bump version and push to release branch.
+	ifndef v
+		$(error v is required, ex. make version v=0.1.0)
 	@poetry version $(v)
 	@git add pyproject.toml
 	@git commit -m "Version bump v$$(poetry version -s)"
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,20 @@
-version:
+.PHONY: version
+version: ## Bump version and push to release branch. ex: make version v=0.1.0
 	@poetry version $(v)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,11 +1,22 @@
...
 	@git push --tags
 	@poetry build
 	@poetry publish
+
+.DEFAULT_GOAL := help
+.PHONY: help
+help: ## Print Makefile help text.
+	@grep -E '^[a-zA-Z_\/-]+:.*?## .*$$' $(MAKEFILE_LIST) \
+	| awk 'BEGIN {FS = ":.*?## "}; \
+		{printf "\033[36m%-20s\033[0m%s\n", $$1, $$2}'
```
collindutter, line 22: Can you explain this wizardry a bit 😅

andrewfrench: This uses grep to parse the Makefile for targets with comments in the format:

```
target: ## comment
```

These are passed to awk which then splits the incoming line into a target and comment variable, then prints them inside some terminal formatting directives. Any targets without comments are ignored and excluded from help text.

collindutter (approved):

vasinov (approved):
Are there other checks we should add? For example, I think it currently fails if the git tag already exists (in the scenario where a previous run failed at the poetry build/publish step).

andrewfrench:
I think there are -- it'd be cool if we could unify our test and static check entrypoints through the Makefile, giving us consistent behavior for testing pre-merge and with or without pre-commit hooks. I know we do at least some of these via marketplace actions so I can take a look later on to see how feasible that is.

PR #541, opened 2024-01-05
Add CodeExecutionTask for executing python code

aodhanroche:
Griptape users can now execute arbitrary Python code as a task within a Griptape structure such as a Pipeline, Agent, or Workflow. The CodeExecutionTask class allows the user to pass in a Python function without requiring LLM interaction. Authors can elect for their CodeExecutionTask callback to return a custom artifact to pass on to the owning structure.


griptape/tasks/python_task.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def hello_world(task: BaseTextInputTask) -> BaseArtifact:
+    greeting = "Hello World!"
+    # We don't want to create an artifact, so just pass the parent task's input back
+    return TextArtifact(greeting)
+
+
+def deliberate_exception(task: BaseTextInputTask) -> BaseArtifact:
+    raise ValueError("ugh blub")
```
andrewfrench, line 25: Are these functions a developer would call? At first glance they seem like examples that may be better suited for documentation.

aodhanroche: Yes, I should move those to documentation

griptape/tasks/python_task.py
```diff
@@ -0,0 +1,26 @@
+from attr import define, field
+from griptape.artifacts import TextArtifact, BaseArtifact, ErrorArtifact
```
SavagePencil, line 2: I think we can get rid of `TextArtifact` from the import list once you get rid of the example functions.

griptape/tasks/python_task.py
SavagePencil, line 25: I would look to see if there is a "better" exception to raise (and apologies for giving you the "ugh blub" string).

tests/unit/tasks/test_python_task.py
```diff
@@ -0,0 +1,39 @@
+from griptape.tasks import PythonTask
+from griptape.tasks.python_task import hello_world, deliberate_exception
```
SavagePencil, line 2: Per earlier comment, we can move these functions in here and out of the main class definition. Then you can eliminate the import.

SavagePencil (changes requested):

griptape/tasks/python_task.py
aodhanroche: Are you referring to just changing the error string, or what were you thinking

griptape/tasks/python_task.py
aodhanroche: Getting rid of this...

tests/unit/tasks/test_python_task.py
```diff
@@ -0,0 +1,26 @@
+from griptape.artifacts import BaseArtifact, TextArtifact
+from griptape.tasks import PythonTask, BaseTextInputTask, BaseTask
+from tests.mocks.mock_prompt_driver import MockPromptDriver
+from griptape.structures import Pipeline
+
+
+def hello_world(task: BaseTask) -> BaseArtifact:
+    greeting = "Hello World!"
+    # We don't want to create an artifact, so just pass the parent task's input back
+    return TextArtifact(greeting)
```
SavagePencil, line 10: this comment doesn't mesh with what's being returned. We want to tell customers that if they want essentially a no-op to return `task.input`, right?

tests/unit/tasks/test_python_task.py
```diff
@@ -0,0 +1,26 @@
...
+
+def deliberate_exception(task: BaseTask) -> BaseArtifact:
+    raise ValueError("ugh blub")
+
+
+class TestPromptSubtask:
+    def test_hello_world_fn(self):
+        task = PythonTask(run_fn=hello_world)
+
+        assert task.run().value == "Hello World!"
```
SavagePencil, line 21: follow up on the earlier comment, maybe we want THREE tests: one returning TextArtifact, one "no-op" one returning the task.input, and one deliberately raising the exception

tests/unit/tasks/test_python_task.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def hello_world(task: BaseTask) -> BaseArtifact:
+    greeting = "Hello World!"
+    # We don't want to create an artifact, so just pass the parent task's input back
+    return TextArtifact(greeting)
+
+
+def deliberate_exception(task: BaseTask) -> BaseArtifact:
+    raise ValueError("ugh blub")
```
SavagePencil, line 14: I will go on record and admit that my suggestion of "ugh blugh" was unprofessional and disrespectful and we should change this :)

SavagePencil (changes requested):
ensure we have the no-op case handled and working as expected. Clean up my silliness :)

tests/unit/tasks/test_python_task.py
```diff
@@ -0,0 +1,26 @@
...
+class TestPromptSubtask:
+    def test_hello_world_fn(self):
+        task = PythonTask(run_fn=hello_world)
+
+        assert task.run().value == "Hello World!"
+
+    def test_error_fn(self):
+        task = PythonTask(run_fn=deliberate_exception)
+
+        assert task.run().value == "error during Python Task execution: ugh blub"
```
SavagePencil, line 26: I think we should confirm that what was generated was an `ErrorArtifact` rather than peeking inside the text.

tests/unit/tasks/test_python_task.py
```diff
@@ -0,0 +1,43 @@
...
+
+def hello_world(task: BaseTask) -> BaseArtifact:
+    greeting = "Hello World!"
+    return TextArtifact(greeting)
+
+
+def non_outputting(task: BaseTask) -> BaseArtifact:
+    # If your task function doesn't have an output, return task.input
+    temp = task.input
+    return temp
```
SavagePencil, line 15: We were just using the temp line so we could inspect in the debugger while we were working through issues. Let's get rid of the `temp = task.input` line and just `return task.input`

SavagePencil (changes requested):
tiniest change to get rid of temp var

SavagePencil (dismissed):

andrewfrench:
Looking great @aodhanroche. To help out reviewers, could you add a descriptive title and some information on the contents of the PR? This is especially important for new functionality so we can understand the motivation behind the code and your intentions (what the code _should_ be doing) in addition to what it _is_ doing as implemented.

griptape/tasks/__init__.py
```diff
@@ -28,4 +29,5 @@
     "ImageOutpaintingTask",
     "ImageVariationTask",
     "BaseImageGenerationTask",
+    "PythonTask",
```
collindutter, line 32: How do we feel about this name @vasinov @andrewfrench @SavagePencil? Feels very literal 😅

griptape/tasks/python_task.py
```diff
@@ -0,0 +1,15 @@
+from attr import define, field
+from griptape.artifacts import BaseArtifact, ErrorArtifact
+from griptape.tasks import BaseTextInputTask, BaseTask
+from typing import Callable
+
+
+@define
+class PythonTask(BaseTextInputTask):
+    run_fn: Callable[[BaseTask], BaseArtifact] = field(kw_only=True)
```
collindutter, line 9: I think we can reduce the typing scope from `BaseTask` to `PythonTask` since it's always passing itself to `run_fn`

griptape/tasks/__init__.py
aodhanroche: `ArbitraryFunctionTask`?

SavagePencil: is this something that could expand beyond Python? Yes, we can see that. We could always make a BaseArbitraryCodeTask or what-have-you that eventually leads to it. But end of the day we supply a callback with context and it goes and does stuff without involving the LLM. "Does it what it says on the tin." Any suggestions for something less literal that leaves doors open?

andrewfrench: Pretty inconsequential overall but I'd favor CallableTask

collindutter: Agreed, I think I prefer like CallableTask the most.

SavagePencil: so it shall be

aodhanroche: CodeExecutionTask

collindutter (approved):

andrewfrench (approved):
Nice! 🎉

SavagePencil (approved):
let's get sweaty!

andrewfrench (approved):

PR #547, opened 2024-01-08
Feature/config

collindutter:


griptape/config/base_structure_config.py
```diff
@@ -0,0 +1,26 @@
...
+from griptape.drivers import BasePromptDriver
+from griptape.mixins import SerializableMixin
+from griptape.schemas.base_schema import BaseSchema
+from marshmallow import class_registry
+from marshmallow.exceptions import RegistryError
+
+
+@define
+class BaseStructureConfig(SerializableMixin, ABC):
+    prompt_driver: BasePromptDriver = field(kw_only=True, metadata={"save": True})
```
collindutter, line 14: Definitely need a better name than `save` but this new metadata field will be used when determining what fields to serialize.

griptape/config/base_structure_config.py
```diff
@@ -0,0 +1,26 @@
...
+from griptape.config import StructureTaskMemoryConfig
+from griptape.drivers import BasePromptDriver
+from griptape.mixins import SerializableMixin
+from griptape.schemas.base_schema import BaseSchema
+from marshmallow import class_registry
+from marshmallow.exceptions import RegistryError
+
+
+@define
+class BaseStructureConfig(SerializableMixin, ABC):
```
collindutter, line 13: New mixin.

griptape/schemas/base_schema.py
```diff
@@ -1,5 +1,63 @@
...
+    @classmethod
+    def from_attrscls(cls, attrscls):
+        """Generate a Schema from an attrs class."""
+        from marshmallow import post_load
+        from importlib import import_module
+
+        class SubSchema(cls):
+            @post_load
+            def make_obj(self, data, **kwargs):
+                package_name = ".".join(attrscls.__module__.split(".")[:2])
```
collindutter, line 20: Need advice here. `attrscls` will be something like `<class 'griptape.drivers.prompt.base_prompt_driver.BasePromptDriver'>` and we need to get the packages its subclasses will exist in. I.e. `griptape.drivers`.

Is there a better way to do extract this info from a class? Its top level package where it is exported?

pyproject.toml
```diff
@@ -54,6 +54,8 @@ opensearch-py = { version = "^2.3.1", optional = true }
 requests-aws4auth = { version = "1.2.3", optional = true }
 pgvector = { version = "^0.2.3", optional = true }
 psycopg2-binary = { version = "^2.9.9", optional = true }
+cattrs = {git = "https://github.com/python-attrs/cattrs.git", rev = "main"}
+marshmallow-dataclass = "^8.6.0"
```
collindutter, line 58: Leaving for now as artifacts of what else I tried. `cattrs` was somewhat promising but kept crapping out on complex classes. `marshmallow-dataclass` is exactly what we want but does not support `attrs` :(

griptape/config/base_structure_config.py
SavagePencil: since it's going both directions, maybe `serialize` or `serializable`

griptape/config/base_structure_config.py
SavagePencil: as we think about start/stop, should we consider what this might mean for serializing a structure, too?

griptape/config/openai_structure_config.py
```diff
@@ -0,0 +1,45 @@
...
+            lambda: StructureTaskMemoryConfig(
+                query_engine=StructureTaskMemoryQueryEngineConfig(
+                    prompt_driver=OpenAiChatPromptDriver(model="gpt-3.5-turbo-16k"),
+                    vector_store_driver=LocalVectorStoreDriver(
+                        embedding_driver=OpenAiEmbeddingDriver(model="text-embedding-ada-002")
+                    ),
+                ),
+                extraction_engine=StructureTaskMemoryExtractionEngineConfig(
+                    csv=StructureTaskMemoryExtractionEngineCsvConfig(
+                        prompt_driver=OpenAiChatPromptDriver(model="gpt-3.5-turbo-16k")
```
SavagePencil, line 32: I'd like to get input from the folks who have been working on Griptape Cloud how a text-based config file could be hitched up to BaseStructureConfig in order to make this truly data-driven

pyproject.toml
SavagePencil: what kind of crap-out were you seeing?

griptape/config/base_structure_config.py
SavagePencil: Unity has a cute little pair of decorators for C#: IIRC, they use `[Serializable]` if you want an entire struct/class serialized, or you can make it laser-precise with `[SerializeField]'.

griptape/config/openai_structure_config.py
vachillo: this config file can be set on the App level and loaded in when the app runs. it could also be set on the Deployment level, i think there is pros and cons to both.

griptape/drivers/vector/base_vector_store_driver.py
collindutter, line 1: When `from __future__ import annotations` is present, types like `BaseEmbeddingDriver` will be a string, not a class. This prevents us from inspecting it further. There must be a way to resolve the class from a string annotation.

griptape/config/base_structure_config.py
collindutter: Yeah there are a few different syntaxes we could play with here.

griptape/drivers/vector/base_vector_store_driver.py
collindutter: `typing.get_type_hints` appears to be the solution but I get `TypeError: 'BaseEmbeddingDriver' is not a module, class, method, or function.` in `BaseVectorStoreClient`. This is similar to the errors I was getting when using `cattrs`.

tests/unit/config/test_openai_structure_config.py
```diff
@@ -0,0 +1,74 @@
...
+            "type": "OpenAiStructureConfig",
+            "prompt_driver": {
+                "type": "OpenAiChatPromptDriver",
+                "temperature": 0.1,
+                "max_tokens": None,
+                "model": "gpt-4",
+                "stream": False,
+            },
+            "task_memory": {
+                "type": "StructureTaskMemoryConfig",
```
collindutter, line 21: Can probably dynamically figure this out from the level of nesting.

PR #548, opened 2024-01-09
Support task memory inputs in image generation tools

andrewfrench:
Image tools that take image artifact inputs (VariationImageGenerationClient, InpaintingImageGenerationClient, and OutpaintingImageGenerationClient) now support inputs in memory in addition to images on disk.

Other additions:
- FileLoader supports loading images through explicit file extension matching
- ImageLoader tests to prove image type support
- Utility function `load_artifact_from_memory`


griptape/utils/load_image_artifact_from_memory.py
collindutter, line 1: This seems like a handy util outside of images. Can we make this more generic for use with other Artifact types?

andrewfrench: We can, let's scope that outside this PR

collindutter: Even if we don't apply it anywhere else in this PR, I think it would be good to make it generic before merging. Should just be a rename and maybe an extra field for the artifact type?

collindutter (approved):

PR #561, opened 2024-01-11
Refactor/auto serialize

collindutter:
This PR introduces a new `SerializableMixin` and heavily modifies `BaseSchema` to allow for automatic serialization/deserialization of attrs classes. I have deleted all manually created Schemas in favor of this new method.


griptape/events/base_event.py
```diff
@@ -1,56 +1,39 @@
...
-            StartPromptEventSchema,
-            FinishPromptEventSchema,
-            StartTaskEventSchema,
-            FinishTaskEventSchema,
-            StartActionSubtaskEventSchema,
-            FinishActionSubtaskEventSchema,
-            StartStructureRunEventSchema,
-            FinishStructureRunEventSchema,
-            CompletionChunkEventSchema,
+    def try_get_schema(cls, obj_type: str) -> list[type[Schema]] | type[Schema]:
```
vasinov, line 16: Same here.

griptape/schemas/base_schema.py
```diff
@@ -1,5 +1,89 @@
...
 
+from griptape.schemas.bytes_field import Bytes
+
 
 class BaseSchema(Schema):
-    schema_namespace = fields.Str(allow_none=True)
+    DATACLASS_TYPE_MAPPING = {**Schema.TYPE_MAPPING, dict: fields.Dict, bytes: Bytes}
+
+    @classmethod
+    def from_attrscls(cls, attrscls):
```
vasinov, line 15: Should it be `from_attrs_cls`?

griptape/schemas/base_schema.py
```diff
@@ -1,5 +1,79 @@
...
+        attrs.resolve_types(
+            attrscls,
+            localns={
+                "PromptStack": PromptStack,
+                "Input": PromptStack.Input,
+                "Structure": Structure,
+                "BaseConversationMemoryDriver": BaseConversationMemoryDriver,
+                "BasePromptDriver": BasePromptDriver,
+            },
+        )
```
collindutter, line 40: Semi arbitrary feeling imports to make this work. Without these we get a `NameError` on `resolve_types`. I think it has to do with import order.

griptape/artifacts/base_artifact.py
```diff
@@ -31,54 +34,39 @@ def value_to_dict(cls, value: any) -> dict:
...
-    def from_dict(cls, artifact_dict: dict) -> BaseArtifact:
-        from griptape.schemas import (
-            TextArtifactSchema,
-            InfoArtifactSchema,
-            ErrorArtifactSchema,
-            BlobArtifactSchema,
-            CsvRowArtifactSchema,
-            ListArtifactSchema,
-            ImageArtifactSchema,
+    def try_get_schema(cls, obj_type: str) -> list[type[Schema]] | type[Schema]:
```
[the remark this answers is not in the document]
collindutter: There a may be a more marshmallow native way to do it, but at the very least we could probably move this logic to a class decorator.

griptape/memory/meta/action_subtask_meta_entry.py
```diff
@@ -13,11 +13,8 @@ class ActionSubtaskMetaEntry(BaseMetaEntry):
         answer: tool-generated and memory-processed response from Griptape.
     """
 
-    thought: str = field(kw_only=True)
-    action: str = field(kw_only=True)
-    answer: str = field(kw_only=True)
+    type: str = field(default=BaseMetaEntry.__name__, kw_only=True, metadata={"serializable": False})
```
collindutter, line 16: Note the explicit False.

griptape/artifacts/base_artifact.py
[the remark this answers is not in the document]
collindutter: Refactored

griptape/drivers/image_generation_model/bedrock_stable_diffusion_image_generation_model_driver.py
```diff
@@ -115,7 +115,7 @@ def _request_parameters(
         }
 
         if image:
-            request["init_image"] = image.base64
+            request["init_image"] = image.value
```
andrewfrench, line 118: This change passes raw bytes to the image model APIs that expect base64-encoded images.

griptape/artifacts/image_artifact.py
```diff
@@ -37,14 +38,5 @@ def make_name(self) -> str:
 
         return f"image_artifact_{fmt_time}_{entropy}.{extension}"
 
-    @property
-    def base64(self) -> str:
-        return base64.b64encode(self.value).decode("utf-8")
-
```
andrewfrench, line 43: I know that we now automatically base64-encode when serializing, but I think we still need this property to be able to pull the base64-encoded image when required.

griptape/utils/prompt_stack.py
```diff
@@ -1,23 +1,24 @@
...
     SYSTEM_ROLE = "system"
 
-    @dataclass
-    class Input:
-        content: str | list[dict]
-        role: str
+    @define
+    class Input(SerializableMixin):
+        content: str = field(metadata={"serializable": True})
+        role: str = field(metadata={"serializable": True})
```
andrewfrench, line 21: minor, but is there any value in defining a reusable import to denote serializable/non-serializable fields, something like `field(**utils.serializable)`? Reduces _some_ repetition and addresses the issue that a typo in "serializable" wouldn't be caught by a type check, but it prevents us from easily using `metadata` for other uses.

andrewfrench (changes requested):
Huge wins and a couple image artifact value issues

griptape/artifacts/image_artifact.py
collindutter: Can we just encode at the time it is required? Like in the above example I think we can just do:
`request["init_image"] = base64.b64encode(image.value)`

griptape/utils/prompt_stack.py
collindutter: Interesting point... Another option could be to create a custom field as outlined [here](https://www.attrs.org/en/stable/extending.html#metadata). Something like `serializable_field(...)`. Not sure I love this one though, defining a serializable metadata constant somewhere might be easier.

griptape/artifacts/image_artifact.py
andrewfrench: I assume the `base64` property causes issues (data duplication?) with serialization? Yeah, that's workable, just slightly less convenient.

collindutter: It just requires us implementing [this logic](https://github.com/griptape-ai/griptape/blob/dev/griptape/schemas/artifacts/image_artifact_schema.py#L17) in the automatically generated schemas. Doable, but I think the work required outweighs the minor inconvenience here.

Are there other instances of this that we need to update?

griptape/mixins/serializable_mixin.py
```diff
@@ -0,0 +1,89 @@
...
+            module_name: The module name.
+            class_name: The class name.
+
+        Returns:
+            The imported class if found. Raises `ValueError` if not found.
+        """
+        try:
+            module = import_module(module_name)
+            test = getattr(module, class_name, None)
+        except ModuleNotFoundError:
```
vasinov, line 78: Should we log a warning here? Seems like it could be useful to notify the user.

vasinov (approved):
Looks great! So many important changes!

griptape/mixins/serializable_mixin.py
collindutter: It may have to traverse through several directory layers before it finds it though. These are expected exceptions that I don't think we want to bug the user with.

griptape/drivers/image_generation_model/bedrock_stable_diffusion_image_generation_model_driver.py
```diff
@@ -133,7 +133,7 @@ def _request_parameters(
                 raise ValueError("mask_source must be provided when mask is provided")
 
             request["mask_source"] = mask_source
-            request["mask_image"] = mask.base64
+            request["mask_image"] = mask.value
```
andrewfrench, line 136: Base64-encode the value

griptape/drivers/image_generation_model/bedrock_titan_image_generation_model_driver.py
```diff
@@ -64,7 +64,7 @@ def image_variation_request_parameters(
 
         request = {
             "taskType": "IMAGE_VARIATION",
-            "imageVariationParams": {"text": prompt, "images": [image.base64]},
+            "imageVariationParams": {"text": prompt, "images": [image.value]},
```
andrewfrench, line 67: Base64-encode the value

griptape/drivers/image_generation_model/bedrock_titan_image_generation_model_driver.py
```diff
@@ -94,7 +94,7 @@ def image_inpainting_request_parameters(
 
         request = {
             "taskType": "INPAINTING",
-            "inPaintingParams": {"text": prompt, "image": image.base64, "maskImage": mask.base64},
+            "inPaintingParams": {"text": prompt, "image": image.value, "maskImage": mask.value},
```
andrewfrench, line 97: Base64-encode the value

griptape/drivers/image_generation_model/bedrock_titan_image_generation_model_driver.py
```diff
@@ -116,8 +116,8 @@ def image_outpainting_request_parameters(
             "taskType": "OUTPAINTING",
             "outPaintingParams": {
                 "text": prompt,
-                "image": image.base64,
-                "maskImage": mask.base64,
+                "image": image.value,
+                "maskImage": mask.value,
```
andrewfrench, line 120: Base64-encode these values

andrewfrench (changes requested):
Marked the instances where image values must be base64-encoded

andrewfrench (approved):
woo! 😎

PR #562, opened 2024-01-11
Revert Pyupgrade Pep 604 Typing Rewrites

collindutter:
Reverts pyupgrade's [pep 604 typing rewrites](https://github.com/asottile/pyupgrade?tab=readme-ov-file#pep-604-typing-rewrites) which causes issues with https://github.com/griptape-ai/griptape/pull/561 on versions greater than python 3.9.

Thank goodness for regexes.

### DIFF
--- a/griptape/artifacts/blob_artifact.py
+++ b/griptape/artifacts/blob_artifact.py
@@ -8,7 +8,7 @@ from griptape.artifacts import BaseArtifact
 @define(frozen=True)
 class BlobArtifact(BaseArtifact):
     value: bytes = field(converter=BaseArtifact.value_to_bytes)
-    dir_name: str | None = field(default=None, kw_only=True)
+    dir_name: Optional[str] = field(default=None, kw_only=True)
     encoding: str = field(default="utf-8", kw_only=True)
     encoding_error_handler: str = field(default="strict", kw_only=True)
 
--- a/griptape/drivers/embedding/amazon_sagemaker_embedding_driver.py
+++ b/griptape/drivers/embedding/amazon_sagemaker_embedding_driver.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING
 import json
 from typing import Any
 
--- a/griptape/drivers/embedding/amazon_sagemaker_embedding_driver.py
+++ b/griptape/drivers/embedding/amazon_sagemaker_embedding_driver.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import TYPE_CHECKING, Optional$
+from typing import TYPE_CHECKING, Optional
 import json
 from typing import Any
 
--- a/griptape/drivers/embedding/amazon_sagemaker_embedding_driver.py
+++ b/griptape/drivers/embedding/amazon_sagemaker_embedding_driver.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional$
 import json
 from typing import Any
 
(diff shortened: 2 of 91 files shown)


collindutter:
So this is really only an issue in the context of #561 when the union syntax is used in a field since those are the types that are resolved when building a schema.

Feels kind of weird to require `Union` for fields, but allow `|` syntax elsewhere. Should we just drop 3.9?

collindutter:
Relevant https://github.com/tiangolo/typer/issues/371

tests/unit/mixins/test_seriliazable_mixin.py
```diff
@@ -0,0 +1,3 @@
+import pytest
```
SavagePencil, line 1: did you intend to include this file in this PR?

griptape/artifacts/blob_artifact.py
```diff
@@ -8,7 +8,7 @@
 @define(frozen=True)
 class BlobArtifact(BaseArtifact):
     value: bytes = field(converter=BaseArtifact.value_to_bytes)
-    dir_name: str | None = field(default=None, kw_only=True)
+    dir_name: Optional[str] = field(default=None, kw_only=True)
```
SavagePencil, line 11: Is there any nuance lost between `| None` and `Optional` that could be a gotcha?

SavagePencil (changes requested):
Two questions that do not impact approval but want addressed before we move forward with this one.

tests/unit/mixins/test_seriliazable_mixin.py
collindutter: I did not!

griptape/artifacts/blob_artifact.py
collindutter: No gotchas from a day-to-day standpoint. Only when you start doing reflective operations like I'm doing in #561.

SavagePencil (approved):
bless this mess

andrewfrench:
> So this is really only an issue in the context of #561 when the union syntax is used in a field since those are the types that are resolved when building a schema.
> 
> Feels kind of weird to require `Union` for fields, but allow `|` syntax elsewhere. Should we just drop 3.9?

That does feel weird, and maybe would be easier to drop 3.9. It'd be an easier decision if we had any idea what the version breakdown was among users. I found [these survey results](https://lp.jetbrains.com/python-developers-survey-2022/#PythonVersions) from Jetbrains (the 2023 survey results aren't available yet) implying 3.9 has a pretty large share of users, but clearly pretty out of date by now as it only includes versions as high as 3.10.

collindutter:
Another option could be to go all in on the revert so consistent throughout. I only reverted to `T | None` to `Optional[T]`. We could also revert `T | U | V` to `Union[T, U, V]`, but this would undo a decision made in https://github.com/griptape-ai/griptape/issues/226.

Dropping 3.9 feels wrong, also I think 3.10 still has some issues with `|` syntax.

griptape/drivers/embedding/amazon_sagemaker_embedding_driver.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
```
andrewfrench, line 2: A slightly overzealous find/replace leading to a few unused imports.

griptape/drivers/embedding/base_multi_model_embedding_driver.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from abc import ABC
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
```
andrewfrench, line 3: unused import

griptape/drivers/embedding/huggingface_hub_embedding_driver.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
```
andrewfrench, line 2: unused import

andrewfrench (approved):
Nothing worth blocking approval over but several unused `Optional` imports added here. 

The question around 3.9 support generally is worth discussing but can happen outside this PR.

PR #572, opened 2024-01-13
Fix pyright issues in tools directory

collindutter:
#519

### DIFF
--- a/griptape/tools/base_tool.py
+++ b/griptape/tools/base_tool.py
@@ -53,6 +53,8 @@ class BaseTool(ActivityMixin, ABC):
             for activity_name, memory_list in output_memory.items():
                 if not self.find_activity(activity_name):
                     raise ValueError(f"activity {activity_name} doesn't exist")
+                if memory_list is None:
+                    raise ValueError(f"memory list for activity '{activity_name}' can't be None")
 
                 output_memory_names = [memory.name for memory in memory_list]
 
@@ -92,7 +94,9 @@ class BaseTool(ActivityMixin, ABC):
                 {
                     Literal("name"): self.name,
                     Literal("path", description=self.activity_description(activity)): self.activity_name(activity),
-                    Literal("input"): {"values": activity.config["schema"]} if self.activity_schema(activity) else {},
+                    Literal("input"): {"values": getattr(activity, "config")["schema"]}
+                    if self.activity_schema(activity)
+                    else {},
                 }
             )
             for activity in self.activities()
@@ -126,7 +130,8 @@ class BaseTool(ActivityMixin, ABC):
     def after_run(self, activity: Callable, subtask: ActionSubtask, value: BaseArtifact) -> BaseArtifact:
         if value:
             if self.output_memory:
-                for memory in activity.__self__.output_memory.get(activity.name, []):
+                output_memories = self.output_memory[getattr(activity, "name")] or []
+                for memory in output_memories:
                     value = memory.process_output(activity, subtask, value)
 
                 if isinstance(value, BaseArtifact):
@@ -156,7 +161,7 @@ class BaseTool(ActivityMixin, ABC):
 
         return os.path.dirname(os.path.abspath(class_file))
 
-    def install_dependencies(self, env: dict[str, Optional[str]] = None) -> None:
+    def install_dependencies(self, env: Optional[dict[str, str]] = None) -> None:
         env = env if env else {}
 
         command = [sys.executable, "-m", "pip", "install", "-r", "requirements.txt"]
--- a/griptape/tools/computer/tool.py
+++ b/griptape/tools/computer/tool.py
@@ -6,6 +6,7 @@ import tempfile
 from pathlib import Path
 from typing import Optional, TYPE_CHECKING
 from attr import define, field, Factory
+from docker.models.containers import Container
 from schema import Schema, Literal
 import stringcase
 import docker
@@ -36,7 +37,7 @@ class Computer(BaseTool):
         default=Factory(lambda self: self.default_docker_client(), takes_self=True), kw_only=True
     )
 
-    __tempdir: Optional[tempfile.TemporaryDirectory] = field(default=None, kw_only=True)
+    _tempdir: Optional[tempfile.TemporaryDirectory] = field(default=None, kw_only=True)
 
     def __attrs_post_init__(self) -> None:
         super().__attrs_post_init__()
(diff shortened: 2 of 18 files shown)


griptape/tools/base_tool.py
```diff
@@ -126,7 +130,8 @@ def run(self, activity: Callable, subtask: ActionSubtask, value: Optional[dict])
     def after_run(self, activity: Callable, subtask: ActionSubtask, value: BaseArtifact) -> BaseArtifact:
         if value:
             if self.output_memory:
-                for memory in activity.__self__.output_memory.get(activity.name, []):
```
collindutter, line 129: Is this change fine? When would `activity.__self.__` not equal `self`?

griptape/tools/google_drive/tool.py
```diff
@@ -167,8 +167,8 @@ def save_content_to_drive(self, params: dict) -> ErrorArtifact | InfoArtifact:
         }
     )
     def download_files(self, params: dict) -> ListArtifact | ErrorArtifact:
-        from google.auth.exceptions import MalformedError
-        from googleapiclient.errors import HttpError
+        from google.auth.exceptions import MalformedError  # pyright: ignore
+        from googleapiclient.errors import HttpError  # pyright: ignore
```
collindutter, line 171: Not a fan of ignores, but these dependencies are installed at runtime.

griptape/tools/computer/tool.py
```diff
@@ -106,18 +107,21 @@ def execute_command_in_container(self, command: str) -> BaseArtifact:
...
 
-            if stderr:
-                return ErrorArtifact(stderr)
+                if stderr:
+                    return ErrorArtifact(stderr)
+                else:
+                    return TextArtifact(stdout)
             else:
-                return TextArtifact(stdout)
+                return ErrorArtifact("error running container")
```
emjay07, line 124: what are the guidelines for when to return `ErrorArtifact` vs raising an exception? Are ErrorArtifacts specific to tools?

griptape/tools/email_client/tool.py
```diff
@@ -100,25 +100,30 @@ def retrieve(self, params: dict) -> ListArtifact | ErrorArtifact:
     )
     def send(self, params: dict) -> InfoArtifact | ErrorArtifact:
         values = params["values"]
+        if self.smtp_user is None:
+            return ErrorArtifact("smtp_user is required")
```
emjay07, line 104: We should update the attribute helper comment at the top of this file to specify when these Optional parameters are actually required.

griptape/tools/web_scraper/tool.py
```diff
@@ -25,7 +25,7 @@ def get_content(self, params: dict) -> ListArtifact | ErrorArtifact:
         if isinstance(page, ErrorArtifact):
             return page
         else:
-            return ListArtifact(TextLoader().text_to_artifacts(page.get("text")))
+            return ListArtifact(TextLoader().text_to_artifacts(page["text"]))
```
emjay07, line 28: this seems like it could open us up to improperly handled errors when the key "text" does not exist in order to get rid of the pyright errors. Is it a guarantee that "text" always exists?

If not, we could do something like this:
```
text = page.get("text")
if text is None:
    return ErrorArtifact("")
return ListArtifact(TextLoader().text_to_artifacts(text))
```
Funny enough, this gives me a different Pyright error about wanting Sequence instead of list.

griptape/tools/computer/tool.py
collindutter: `ErrorArtifact`s are not specific to Tools necessarily, but all Tools should return an Artifact of some kind. This information is passed back to the LLM and allows it to potentially course correct.

Beyond Tools, it's kind of situationally dependent on whether a `raise` or `ErrorArtifact` is the correct choice.

griptape/tools/google_drive/tool.py
```diff
@@ -148,7 +148,7 @@ def save_content_to_drive(self, params: dict) -> ErrorArtifact | InfoArtifact:
         try:
             self._save_to_drive(filename, content)
 
-            return InfoArtifact(f"saved successfully")
+            return InfoArtifact("saved successfully")
```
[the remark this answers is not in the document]
collindutter: Probably, but I think improving Artifact messages is out of scope for this PR. This change was only to remove the unnecessary f-string.

griptape/tools/sql_client/tool.py
```diff
@@ -21,7 +21,7 @@ def full_table_name(self) -> str:
         return f"{self.schema_name}.{self.table_name}" if self.schema_name else self.table_name
 
     @property
-    def table_schema(self) -> str:
+    def table_schema(self) -> Optional[str]:
```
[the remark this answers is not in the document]
collindutter: Use `Optional` for optional types, use pipe syntax `|` for non-instance variables, use `Union` for instance variables. If we drop 3.9 in the future we can switch to pipe syntax for Unions in all places.

griptape/tools/web_scraper/tool.py
collindutter: These keys are guaranteed to exist, yes. Here is the result of a [simple site](https://cv.bt.ht/):

```json
{
  "title": "CV",
  "author": null,
  "hostname": null,
  "date": "2023-01-01",
  "fingerprint": "81ccfd77a5f0f861",
  "id": null,
  "license": null,
  "comments": "",
  "raw_text": "Hi, I'm Brad. Resume: Fullstack Engineer @ SerpApi, 2023-Sr. Product Designer @ Donorbox, 2021-2023Web Designer @ Purism, 2019-2021Product Designer @ Benbria, 2013-2019Web Designer @ Netvatise, 2009-2013 Links: bt.htsr.ht/~bt brad@bt.ht",
  "text": "Hi, I'm Brad.\nResume:\nFullstack Engineer @ SerpApi, 2023-Sr. Product Designer @ Donorbox, 2021-2023Web Designer @ Purism, 2019-2021Product Designer @ Benbria, 2013-2019Web Designer @ Netvatise, 2009-2013\nLinks:\nbt.htsr.ht/~bt\nbrad@bt.ht",
  "language": null,
  "image": null,
  "pagetype": null,
  "source": null,
  "source-hostname": null,
  "excerpt": null,
  "categories": "",
  "tags": ""
}
```

I updated the logic though, it was previously checking for an `ErrorArtifact` from `WebLoader` which it does not return.

emjay07 (approved):

PR #573, opened 2024-01-16
Add support for Dall-E 2 image to image generation

andrewfrench:


griptape/drivers/image_generation/openai_image_generation_driver.py
```diff
@@ -95,7 +118,32 @@ def try_image_inpainting(
...
         negative_prompts: Optional[list[str]] = None,
     ) -> ImageArtifact:
-        raise NotImplementedError(f"{self.__class__.__name__} does not support inpainting")
+        if self.model != "dall-e-2":
+            raise NotImplementedError(f"{self.__class__.__name__} only supports dall-e-2 for image variation")
+
+        if self.image_size not in {"256x256", "512x512", "1024x1024"}:
+            raise ValueError(f"dall-e-2 image variation support image sizes 256x256, 512x512, and 1024x1024")
+
+        image_size = cast(Literal["256x256", "512x512", "1024x1024"], self.image_size)
```
andrewfrench, line 127: Annoying stuff happening here: while both `dall-e-2` and `dall-e-3` use the same SDK, only `dall-e-2` supports editing (inpainting) and variations, at least for now. This means we need some checks to determine if the driver configuration makes sense for the method call. 

Maybe we need to split into two distinct model drivers given the distinct behavior?

griptape/drivers/image_generation/openai_image_generation_driver.py
```diff
@@ -86,7 +86,30 @@ def try_text_to_image(self, prompts: list[str], negative_prompts: Optional[list[
     def try_image_variation(
         self, prompts: list[str], image: ImageArtifact, negative_prompts: Optional[list[str]] = None
     ) -> ImageArtifact:
-        raise NotImplementedError(f"{self.__class__.__name__} does not support variation")
+        if self.model != "dall-e-2":
```
cjkindel, line 89: Nit: Should we consolidate the model/resolution checks into a common fn? Do we see image variation & inpainting having identical supported implementation long-term or are they just identical for now?

griptape/drivers/image_generation/openai_image_generation_driver.py
```diff
@@ -86,7 +86,30 @@ def try_text_to_image(self, prompts: list[str], negative_prompts: Optional[list[
...
+        if self.image_size not in {"256x256", "512x512", "1024x1024"}:
+            raise ValueError(f"dall-e-2 image variation support image sizes 256x256, 512x512, and 1024x1024")
+
+        image_size = cast(Literal["256x256", "512x512", "1024x1024"], self.image_size)
+        response = self.client.images.create_variation(
+            image=image.value, n=1, response_format=self.response_format, size=image_size
+        )
+
+        if not response.data[0] or not response.data[0].b64_json:
+            raise Exception("Failed to generate image")
```
collindutter, line 101: Are we looking for `None`? If so, we should be explicit with `response.data[0] is None or response.data[0].b64.json is None`.

https://google.github.io/styleguide/pyguide.html#2144-decision

griptape/drivers/image_generation/openai_image_generation_driver.py
```diff
@@ -86,7 +86,30 @@ def try_text_to_image(self, prompts: list[str], negative_prompts: Optional[list[
...
+
+        if self.image_size not in {"256x256", "512x512", "1024x1024"}:
+            raise ValueError(f"dall-e-2 image variation support image sizes 256x256, 512x512, and 1024x1024")
+
+        image_size = cast(Literal["256x256", "512x512", "1024x1024"], self.image_size)
+        response = self.client.images.create_variation(
+            image=image.value, n=1, response_format=self.response_format, size=image_size
+        )
+
+        if not response.data[0] or not response.data[0].b64_json:
```
collindutter, line 100: Will `response.data` always have at least a single element even in a failure case?

griptape/drivers/image_generation/openai_image_generation_driver.py
collindutter: That is annoying, but the Model Driver pattern feels more applicable to platforms that provide models from different vendors. I think these one off checks are fine for now as long as it's minimal.

griptape/drivers/image_generation/openai_image_generation_driver.py
```diff
@@ -95,7 +118,32 @@ def try_image_inpainting(
...
+            raise ValueError(f"dall-e-2 image variation support image sizes 256x256, 512x512, and 1024x1024")
+
+        image_size = cast(Literal["256x256", "512x512", "1024x1024"], self.image_size)
+        prompt = ", ".join(prompts)
+        response = self.client.images.edit(
+            prompt=prompt, image=image.value, mask=mask.value, response_format=self.response_format, size=image_size
+        )
+
+        if not response.data[0] or not response.data[0].b64_json:
+            raise Exception("Failed to generate image")
```
collindutter, line 134: Same points as above.

collindutter (changes requested):

griptape/drivers/image_generation/openai_image_generation_driver.py
andrewfrench: Consolidated. The two implementations are pretty similar, but not quite identical.

collindutter (approved):

PR #603, opened 2024-01-22
Fix pyright errors in /tasks

### LINKED ISSUE #576: Resolve Pyright errors for Tasks

collindutter:
Resolves #576

### DIFF
--- a/griptape/tasks/action_subtask.py
+++ b/griptape/tasks/action_subtask.py
@@ -32,7 +32,7 @@ class ActionSubtask(BaseTextInputTask):
         },
     )
 
-    _input: Optional[str] = field(default=None)
+    _input: Optional[str | TextArtifact | Callable[[BaseTask], TextArtifact]] = field(default=None)
     parent_task_id: Optional[str] = field(default=None, kw_only=True)
     thought: Optional[str] = field(default=None, kw_only=True)
     action_name: Optional[str] = field(default=None, kw_only=True)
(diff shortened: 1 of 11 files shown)


cjkindel (dismissed):

griptape/tasks/action_subtask.py
```diff
@@ -47,17 +46,33 @@ class ActionSubtask(PromptTask):
     def input(self) -> TextArtifact:
         return TextArtifact(self._input)
 
+    @input.setter
+    def input(self, value: str | TextArtifact | Callable[[BaseTask], TextArtifact]) -> None:
+        if isinstance(value, str):
+            self._input = value
+        else:
+            raise Exception("ActionSubtask input must be a string.")
```
andrewfrench, line 54: I think the signature and the implementation are misleading here

griptape/tasks/prompt_task.py
```diff
@@ -17,7 +17,7 @@ class PromptTask(BaseTextInputTask):
         default=Factory(lambda self: self.default_system_template_generator, takes_self=True), kw_only=True
     )
 
-    output: TextArtifact | ErrorArtifact | Optional[InfoArtifact] = field(default=None, init=False)
+    output: BaseArtifact = field(default=None, init=False)
```
andrewfrench, line 20: Optional?

andrewfrench (changes requested):
comments above

griptape/tasks/action_subtask.py
collindutter: Maybe we should just remove this override? It's not really clear to me why it was overridden in the first place.

griptape/tasks/prompt_task.py
collindutter: Good catch

griptape/tasks/action_subtask.py
andrewfrench: Looks like it'd be fine if we removed the override.

collindutter: Updated. The key part of the override is that we don't render a jinja template. We want the input exactly as it is.

andrewfrench (approved):

PR #604, opened 2024-01-22
Address Dependabot isses for dependencies

### LINKED ISSUE #496: Address Dependabot issue for pypdf (link to issue inside)
https://github.com/griptape-ai/griptape/security/dependabot/4

cjkindel:
Increased specificity of jinja2 version as well (already using >=3.1.3).
Updated dev dependency install script in README
Ignored pyright complaint on renaming of load/load_collection parameters for PdfLoader from parent TextLoader. (Necessary because I updated the file)
Resolves #496


griptape/loaders/pdf_loader.py
```diff
@@ -16,10 +16,10 @@ class PdfLoader(TextLoader):
         kw_only=True,
     )
 
-    def load(self, stream: str | IO | Path, password: Optional[str] = None) -> list[TextArtifact]:
+    def load(self, stream: str | IO | Path, password: Optional[str] = None) -> list[TextArtifact]:  # type: ignore
         return self._load_pdf(stream, password)
 
-    def load_collection(
+    def load_collection(  # type: ignore
```
collindutter, line 22: Let's resolve these type errors while we're already in here. What was the pyright output?

collindutter: Or if you'd like to help resolve https://github.com/griptape-ai/griptape/issues/585, we can merge that in first.

collindutter (commented):
Can you please also add a `Security` note in the CHANGELOG.md?

griptape/loaders/pdf_loader.py
collindutter: Hmm ok this probably requires a bit of thinking on how we want to structure this. Let's still remove the `type: ignore` so we don't mask the errors. You can skip the git precommit hooks with the `--no-verify` commit flag.

collindutter (approved):

PR #612, opened 2024-01-24
Ignore pyright issues in /schemas

aodhanroche:
Fixes: #615


collindutter (approved):
Approved. We adapted this file from another library, and the pyright errors are not worth the investigation effort to resolve.

PR #620, opened 2024-01-27
Remove Tokenizer default models

andrewfrench:
- Removed default models from Tokenizers
- Added MODEL_PREFIX_TO_MAX_TOKENS mapping to relevant Tokenizers
- Audited existing model, max token mappings

### DIFF
--- a/griptape/chunkers/base_chunker.py
+++ b/griptape/chunkers/base_chunker.py
@@ -14,9 +14,7 @@ class BaseChunker(ABC):
     separators: list[ChunkSeparator] = field(
         default=Factory(lambda self: self.DEFAULT_SEPARATORS, takes_self=True), kw_only=True
     )
-    tokenizer: BaseTokenizer = field(
-        default=Factory(lambda: OpenAiTokenizer(model=OpenAiTokenizer.DEFAULT_OPENAI_GPT_3_CHAT_MODEL)), kw_only=True
-    )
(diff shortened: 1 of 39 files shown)


andrewfrench:
Pyright checks are failing due to unrelated issues in test files.

griptape/chunkers/base_chunker.py
```diff
@@ -14,9 +14,7 @@ class BaseChunker(ABC):
     separators: list[ChunkSeparator] = field(
         default=Factory(lambda self: self.DEFAULT_SEPARATORS, takes_self=True), kw_only=True
     )
-    tokenizer: BaseTokenizer = field(
-        default=Factory(lambda: OpenAiTokenizer(model=OpenAiTokenizer.DEFAULT_OPENAI_GPT_3_CHAT_MODEL)), kw_only=True
-    )
+    tokenizer: BaseTokenizer = field(default=Factory(lambda: OpenAiTokenizer(model="gpt-3.5-turbo")), kw_only=True)
```
cjkindel, line 17: What is the value of shifting from an global enum to a local string? I understand removing defaults to require defining a model, but in this case we are just replacing a enum defined to be "gpt-3.5-turbo" with the string "gpt-3.5-turbo".

collindutter: I'm actually kind of inclined to remove from the use of `OpenAiTokenizer` entirely. Chunkers are low level enough where this should be something another class passes in.

collindutter:
Implemented in https://github.com/griptape-ai/griptape/pull/690

PR #622, opened 2024-01-29
Fix Amazon Opensearch Vector Store Driver

collindutter:
Improve class field signatures, and add special upsert check for aoss which does not allow for custom ids during insertion.

### DIFF
--- a/griptape/drivers/vector/amazon_opensearch_vector_store_driver.py
+++ b/griptape/drivers/vector/amazon_opensearch_vector_store_driver.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 from attr import define, field, Factory
-from typing import Optional, Tuple, TYPE_CHECKING
+from typing import Optional, TYPE_CHECKING
 from griptape.drivers import OpenSearchVectorStoreDriver
-from griptape.utils import import_optional_dependency
+from griptape.utils import import_optional_dependency, str_to_hash
 
 if TYPE_CHECKING:
     from boto3 import Session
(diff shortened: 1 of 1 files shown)


griptape/drivers/vector/amazon_opensearch_vector_store_driver.py
```diff
@@ -45,3 +43,27 @@ class AmazonOpenSearchVectorStoreDriver(OpenSearchVectorStoreDriver):
...
+        meta: Optional[dict] = None,
+        **kwargs,
+    ) -> str:
+        """Inserts or updates a vector in OpenSearch.
+
+        If a vector with the given vector ID already exists, it is updated; otherwise, a new vector is inserted.
+        Metadata associated with the vector can also be provided.
+        """
+
+        vector_id = vector_id if vector_id else str_to_hash(str(vector))
```
andrewfrench, line 61: I realize that this isn't new functionality, but curious about creating an ID directly from the vector. I guess a vector is already a kind of non-cryptographic hash so the risk of collision is low and duplicate vectors (and therefore duplicate hashes and IDs) would strongly imply duplicate underlying data, but even for duplicate data the metadata could differ. 

Any particular reason to build the ID from the vector rather than assign a new UUID?

andrewfrench (approved):

griptape/drivers/vector/amazon_opensearch_vector_store_driver.py
collindutter: I suspect it is to allow for upserting a vector. The metadata may change but you still want to update the same vector entry.

andrewfrench (approved):

PR #624, opened 2024-01-30
Move loader dependencies to extras

collindutter:
Move dependencies for Loaders to extras. This drastically reduces the core installation size, `pandas` was taking up nearly 50MB and is only used in a relatively niche Loader.


pyproject.toml
```diff
@@ -22,24 +22,17 @@ marshmallow-enum = ">=1.5"
 tiktoken = ">=0.3"
 rich = ">=13"
 schema = ">=0.7"
-pyyaml = ">=6"
```
collindutter, line 25: Wasn't being used at all.

andrewfrench (approved):

vasinov (approved):

PR #625, opened 2024-01-30
OpenAI Vision image query functionality

andrewfrench:
This PR introduces image query functionality using the OpenAI Vision model.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `JsonExtractionTask` for convenience over using `ExtractionTask` with a `JsonExtractionEngine`.
 - `CsvExtractionTask` for convenience over using `ExtractionTask` with a `CsvExtractionEngine`.
 - `OpenAiVisionImageQueryDriver` to support queries on images using OpenAI's Vision model.
-- `ImageQueryClient` allowing an agent to make queries on images on disk or in memory.
-- `ImageQueryEngine` and `ImageQueryTask`.
+- `ImageQueryClient` allowing an Agent to make queries on images on disk or in Task Memory.
+- `ImageQueryTask` and `ImageQueryEngine`.
 
 ### Fixed 
 - `BedrockStableDiffusionImageGenerationModelDriver` request parameters for SDXLv1.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `StructureConfig` for building your own Structure configuration.
 - `JsonExtractionTask` for convenience over using `ExtractionTask` with a `JsonExtractionEngine`.
 - `CsvExtractionTask` for convenience over using `ExtractionTask` with a `CsvExtractionEngine`.
+- `OpenAiVisionImageQueryDriver` to support queries on images using OpenAI's Vision model.
+- `ImageQueryClient` allowing an agent to make queries on images on disk or in memory.
+- `ImageQueryEngine` and `ImageQueryTask`.
 
 ### Fixed 
 - `BedrockStableDiffusionImageGenerationModelDriver` request parameters for SDXLv1.
--- a/griptape/config/openai_structure_config.py
+++ b/griptape/config/openai_structure_config.py
@@ -26,7 +26,7 @@ class OpenAiStructureConfig(BaseStructureConfig):
             lambda: StructureGlobalDriversConfig(
                 prompt_driver=OpenAiChatPromptDriver(model="gpt-4"),
                 image_generation_driver=OpenAiImageGenerationDriver(model="dall-e-2", image_size="512x512"),
-                image_query_driver=OpenAiVisionImageQueryDriver(model="gpt-4-vision-preview"),
+                image_query_driver=OpenAiVisionImageQueryDriver(model="gpt-4-vision-preview", max_tokens=300),
                 embedding_driver=OpenAiEmbeddingDriver(model="text-embedding-ada-002"),
                 vector_store_driver=LocalVectorStoreDriver(
                     embedding_driver=OpenAiEmbeddingDriver(model="text-embedding-ada-002")
--- a/griptape/config/openai_structure_config.py
+++ b/griptape/config/openai_structure_config.py
@@ -15,6 +15,7 @@ from griptape.drivers import (
     OpenAiChatPromptDriver,
     OpenAiEmbeddingDriver,
     OpenAiImageGenerationDriver,
+    OpenAiVisionImageQueryDriver,
 )
 
 
@@ -25,6 +26,7 @@ class OpenAiStructureConfig(BaseStructureConfig):
             lambda: StructureGlobalDriversConfig(
                 prompt_driver=OpenAiChatPromptDriver(model="gpt-4"),
                 image_generation_driver=OpenAiImageGenerationDriver(model="dall-e-2", image_size="512x512"),
+                image_query_driver=OpenAiVisionImageQueryDriver(model="gpt-4-vision-preview"),
                 embedding_driver=OpenAiEmbeddingDriver(model="text-embedding-ada-002"),
                 vector_store_driver=LocalVectorStoreDriver(
                     embedding_driver=OpenAiEmbeddingDriver(model="text-embedding-ada-002")
--- a/griptape/drivers/image_query/base_image_query_driver.py
+++ b/griptape/drivers/image_query/base_image_query_driver.py
@@ -1,0 +1,45 @@
+from __future__ import annotations
+
+from abc import abstractmethod, ABC
+from typing import Optional, TYPE_CHECKING
+
+from attr import define, field
+
+from griptape.artifacts import TextArtifact, ImageArtifact
+from griptape.events import StartImageQueryEvent, FinishImageQueryEvent
+from griptape.mixins import ExponentialBackoffMixin, SerializableMixin
+
+if TYPE_CHECKING:
+    from griptape.structures import Structure
+
+
+@define
+class BaseImageQueryDriver(SerializableMixin, ExponentialBackoffMixin, ABC):
+    structure: Optional[Structure] = field(default=None, kw_only=True)
+
+    def before_run(self, query: str, images: list[ImageArtifact]) -> None:
+        if self.structure:
+            self.structure.publish_event(
+                StartImageQueryEvent(query=query, images_info=[image.to_text() for image in images])
+            )
+
+    def after_run(self, result: str) -> None:
+        if self.structure:
+            self.structure.publish_event(FinishImageQueryEvent(result=result))
+
+    def query(self, query: str, images: list[ImageArtifact]) -> TextArtifact:
+        for attempt in self.retrying():
+            with attempt:
+                self.before_run(query, images)
+
+                result = self.try_query(query, images)
+
+                self.after_run(result.value)
+
+                return result
+        else:
+            raise Exception("image query driver failed after all retry attempts")
+
+    @abstractmethod
+    def try_query(self, query: str, images: list[ImageArtifact]) -> TextArtifact:
+        ...
--- a/griptape/drivers/image_query/openai_vision_image_query_driver.py
+++ b/griptape/drivers/image_query/openai_vision_image_query_driver.py
@@ -1,0 +1,58 @@
+from __future__ import annotations
+
+from typing import Literal, Optional
+
+from attr import define, field, Factory
+from openai.types.chat import (
+    ChatCompletionUserMessageParam,
+    ChatCompletionContentPartParam,
+    ChatCompletionContentPartTextParam,
+    ChatCompletionContentPartImageParam,
+)
+import openai
+
+from griptape.artifacts import ImageArtifact, TextArtifact
+from griptape.drivers.image_query.base_image_query_driver import BaseImageQueryDriver
+
+
+@define
+class OpenAiVisionImageQueryDriver(BaseImageQueryDriver):
+    model: str = field(default="gpt-4-vision-preview", kw_only=True, metadata={"serializable": True})
+    api_type: str = field(default=openai.api_type, kw_only=True, metadata={"serializable": True})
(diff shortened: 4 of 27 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `AzureMongoDbVectorStoreDriver` for using CosmosDB with MongoDB vCore API.
 - `vector_path` field on `MongoDbAtlasVectorStoreDriver`.
 - `LeonardoImageGenerationDriver` supports image to image generation.
+- `OpenAiVisionImageQueryDriver` to support queries on images using OpenAI's Vision model.
+- `ImageQueryClient` allowing an agent to make queries on images on disk or in memory.
+- Image Query engine and task.
```
collindutter, line 17: Capitalize `engine` and `task`

griptape/drivers/image_query/base_image_query_driver.py
```diff
@@ -0,0 +1,45 @@
...
+        if self.structure:
+            self.structure.publish_event(
+                StartImageQueryEvent(query=query, images_info=[image.to_text() for image in images])
+            )
+
+    def after_run(self, result: str) -> None:
+        if self.structure:
+            self.structure.publish_event(FinishImageQueryEvent(result=result))
+
+    def query(self, query: str, images: list[ImageArtifact]) -> TextArtifact:
```
collindutter, line 30: Should we refactor to `ListArtifact`? Thinking not now, but just raising it.

griptape/drivers/image_query/openai_vision_image_query_driver.py
```diff
@@ -0,0 +1,56 @@
...
+from griptape.artifacts import ImageArtifact, TextArtifact
+from griptape.drivers.image_query.base_image_query_driver import BaseImageQueryDriver
+
+
+@define
+class OpenAiVisionImageQueryDriver(BaseImageQueryDriver):
+    model: str = field(default="gpt-4-vision-preview", kw_only=True, metadata={"serializable": True})
+    api_type: str = field(default=openai.api_type, kw_only=True, metadata={"serializable": True})
+    api_version: Optional[str] = field(default=openai.api_version, kw_only=True, metadata={"serializable": True})
+    base_url: str = field(default=None, kw_only=True, metadata={"serializable": True})
```
collindutter, line 23: Optional?

griptape/drivers/image_query/openai_vision_image_query_driver.py
```diff
@@ -0,0 +1,56 @@
...
+
+        messages = ChatCompletionUserMessageParam(content=message_parts, role="user")
+        params = {"model": self.model, "messages": [messages]}
+
+        if self.max_tokens is not None:
+            params["max_tokens"] = self.max_tokens
+
+        response = self.client.chat.completions.create(**params)
+
+        return TextArtifact(response.choices[0].message.content)
```
collindutter, line 56: Should we add a check similar to the chat prompt driver that verifies that there is exactly one choice?

griptape/events/base_task_event.py
```diff
@@ -12,7 +13,7 @@ class BaseTaskEvent(BaseEvent, ABC):
     task_parent_ids: list[str] = field(kw_only=True, metadata={"serializable": True})
     task_child_ids: list[str] = field(kw_only=True, metadata={"serializable": True})
 
-    task_input: Union[BaseArtifact, BaseArtifact, tuple[BaseArtifact, ...]] = field(
-        kw_only=True, metadata={"serializable": True}
-    )
+    task_input: Union[
+        BaseArtifact, BaseArtifact, tuple[BaseArtifact, ...], tuple[BaseArtifact, Sequence[BaseArtifact]]
```
collindutter, line 17: Is that a duplicate `BaseArtifact`? I know not related to your change, but can we clean that up?

griptape/tasks/image_query_task.py
```diff
@@ -0,0 +1,51 @@
...
+from griptape.tasks import BaseTask
+from griptape.utils import J2
+
+
+@define
+class ImageQueryTask(BaseTask):
+    image_query_engine: ImageQueryEngine = field(kw_only=True)
+    _input: tuple[str, list[ImageArtifact]] | tuple[TextArtifact, list[ImageArtifact]] | Callable[
+        [BaseTask], tuple[TextArtifact, list[ImageArtifact]]
+    ] = field(default=None)
```
collindutter, line 18: If you add `alias="input"` users can set the field using `input` instead of `_input`.

griptape/tasks/image_query_task.py
```diff
@@ -0,0 +1,51 @@
...
+
+from griptape.artifacts import ImageArtifact, TextArtifact
+from griptape.engines import ImageQueryEngine
+from griptape.tasks import BaseTask
+from griptape.utils import J2
+
+
+@define
+class ImageQueryTask(BaseTask):
+    image_query_engine: ImageQueryEngine = field(kw_only=True)
```
collindutter, line 15: May want to merge in #591 first so that we can do something similar to [this](https://github.com/griptape-ai/griptape/pull/591/files#diff-30ce1ef1ca99f2869a727dd437c0a9985d34cddc1e9a922142bdf4020bd366bfR49)

griptape/tasks/image_query_task.py
```diff
@@ -0,0 +1,51 @@
...
+                query_text = self._input[0]
+            else:
+                query_text = TextArtifact(J2().render_from_string(self._input[0], **self.full_context))
+
+            return query_text, self._input[1]
+        elif isinstance(self._input, Callable):
+            return self._input(self)
+        else:
+            raise ValueError(
+                "Input must be a tuple of a text artifact and a list of image artifacts or a callable that "
```
collindutter, line 33: Capitalize artifact names?

griptape/tasks/image_query_task.py
```diff
@@ -0,0 +1,51 @@
...
+                "returns a tuple of a text artifact and a list of image artifacts."
+            )
+
+    @input.setter
+    def input(
+        self,
+        value: tuple[TextArtifact, list[ImageArtifact]]
+        | Callable[[BaseTask], tuple[TextArtifact, list[ImageArtifact]]],
+    ) -> None:
+        self._input = value
```
collindutter, line 43: I _think_ with the `alias` change we can get rid of the setter.

griptape/tasks/image_query_task.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def run(self) -> TextArtifact:
+        query = self.input[0]
+        image_artifacts = self.input[1]
```
collindutter, line 47: Total nit but if we wanna get pythonic with it we can do `query, image_artifacts = self.input`

griptape/tools/image_query_client/tool.py
```diff
@@ -0,0 +1,92 @@
...
+                        {
+                            Literal(
+                                "image_artifact_namespace", description="The namespace of the image artifact in memory."
+                            ): str,
+                            Literal(
+                                "image_artifact_name", description="The name of the image artifact in memory."
+                            ): str,
+                        }
+                    ],
+                    "memory_name": str,
```
collindutter, line 65: Is this usually how we structure the schema for Task Memory fields? Isn't it typically without descriptions and flat?

griptape/drivers/image_query/openai_vision_image_query_driver.py
```diff
@@ -0,0 +1,56 @@
...
+            takes_self=True,
+        )
+    )
+
+    def try_query(self, query: str, images: list[ImageArtifact]) -> TextArtifact:
+        message_parts: list[ChatCompletionContentPartParam] = [
+            ChatCompletionContentPartTextParam(type="text", text=query)
+        ]
+
+        for image in images:
```
emjay07, line 40: what is the expected behavior for a list of images? does it string all of the text together? wondering why defaulting to multiple?

griptape/drivers/image_query/openai_vision_image_query_driver.py
```diff
@@ -0,0 +1,56 @@
...
+                    type="image_url",
+                    image_url={"url": f"data:{image.mime_type};base64,{image.base64}", "detail": self.image_quality},
+                )
+            )
+
+        messages = ChatCompletionUserMessageParam(content=message_parts, role="user")
+        params = {"model": self.model, "messages": [messages]}
+
+        if self.max_tokens is not None:
+            params["max_tokens"] = self.max_tokens
```
emjay07, line 52: if the user doesn't provide max_tokens, what does OpenAI do? does it have a default?

griptape/tasks/image_query_task.py
emjay07, line 33: nit: can we add docs to this task explaining the required order params? I saw that the image_eneration tasks had helpful class-level docs.

griptape/tools/image_query_client/tool.py
```diff
@@ -0,0 +1,92 @@
...
+                    Literal(
+                        "query",
+                        description="A detailed question to be answered using the contents of the provided images.",
+                    ): str,
+                    Literal("image_paths", description="The paths to an image files on disk."): list[str],
+                }
+            ),
+        }
+    )
+    def query_image_from_disk(self, params: dict) -> TextArtifact | ErrorArtifact:
```
emjay07, line 35: not sure if this is a pattern we've used yet: but should this be using the file_manager tool? it has a whole advanced interface for picking the proper loader and error checking for it.

griptape/tasks/image_query_task.py
andrewfrench: I think this just allows us to reference it by the alias in the constructor, like `ImageQueryTask(input=...`

>    :param str | None alias: Override this attribute's parameter name in the
        generated ``__init__`` method. If left `None`, default to ``name``
        stripped of leading underscores. See `private-attributes`.

griptape/tasks/image_query_task.py
andrewfrench: Clean.

griptape/tools/image_query_client/tool.py
andrewfrench: Good point, I removed the descriptions. I don't think we have other examples that behave quite like this. Here, the agent can specify a series of namespace-artifact combinations, which would be important if multiple images are provided as inputs from independent tasks.

griptape/drivers/image_query/openai_vision_image_query_driver.py
andrewfrench: This enables fundamentally different behavior than operating serially over a list of images. As implemented, we can provide many images to Vision and query it in reference to _all the images at once_, not just one at a time. An example would be providing two images and asking the model "What's the difference between these two images?" to understand that one contains a dog and the other contains a cat.

Of course, we can always provide a list of length one to query a single image.

griptape/drivers/image_query/openai_vision_image_query_driver.py
andrewfrench: Unfortunate backstory: OpenAI's APIs have an outstanding issue where your account's rate limit is penalized for the max tokens requested, not just the tokens returned. This is why the `OpenAiChatPromptDriver` only provides `max_tokens` only when explicitly provided. Vision, specifically, also has another complicating issue where it will default to a very small number of output tokens (like 16) unless otherwise requested. 

The interaction of both of these issues means there really isn't a great way to define a default here. Requesting a high `max_tokens` value will very quickly consume a typical user's rate limit. The right number will depend on the context.

griptape/tools/image_query_client/tool.py
andrewfrench: This pattern is used in text+image to image generation tools like VariationImageGenerationClient, InpaintingImageGenerationClient, and OutpaintingImageGenerationClient to provide some flexibility on the source of the input images. I don't think FileManager would make sense here because we only want to load images, so we don't need that loader selection logic.

griptape/tasks/image_query_task.py
andrewfrench: Good call! Added a docstring.

griptape/tasks/image_query_task.py
collindutter: Reposting conversation from [other PR](https://github.com/griptape-ai/griptape/pull/591/files#r1482277530)

griptape/tools/image_query_client/tool.py
emjay07: That's fair. It's just using the ImageLoader. I was more thinking on the validation side. Are there any validations we need to do here to make sure the user is handing us an image file? does it just blow up if it's not an image file?

griptape/tasks/image_query_task.py
andrewfrench: Changed to avoid Pyright errors in the future.

collindutter (dismissed):
Overall looks great. Up to you whether you'd like to wait until #591 is merged in first.

emjay07 (dismissed):

griptape/tools/image_query_client/tool.py
emjay07: discussed in person. the underlying parser will properly surface errors if there are problems parsing.

collindutter:
@andrewfrench 591 went in, let me know if you have any questions on how to integrate here.

griptape/tasks/image_query_task.py
```diff
@@ -0,0 +1,75 @@
...
+    @property
+    def image_query_engine(self) -> ImageQueryEngine:
+        if self._image_query_engine is None:
+            if self.structure is not None:
+                self._image_query_engine = ImageQueryEngine(
+                    image_query_driver=self.structure.config.global_drivers.image_query_driver
+                )
+            else:
+                raise ValueError("Image Query Engine is not set.")
+        return self._image_query_engine
```
collindutter, line 64: Can you please add a unit test for this logic? Should be nearly identical to the unit tests for the other Tasks.

collindutter (changes requested):

griptape/config/openai_structure_config.py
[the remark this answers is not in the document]
andrewfrench: Right, Bedrock will inherit the Dummy driver.

collindutter (approved):

Issue #634, opened 2024-02-05
Allow to set top_n in query_engine.query in TextQueryTask like in VectorStoreClient tool.


igor-2lemetry:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
It would be great to have an ability to configure top n in TextQueryTask by the same way as in VectorStoreClient tool.

**Describe the solution you'd like**
Add optional attribute to TextQueryTask and pass in to query_engine.query.

**Describe alternatives you've considered**
Alternative is using default value 5.

**Additional context**
No additional context.

PR #635, opened 2024-02-05
Added top_n attribute to TextQueryTask

igor-2lemetry:
Closes https://github.com/griptape-ai/griptape/issues/634

### DIFF
--- a/griptape/tasks/text_query_task.py
+++ b/griptape/tasks/text_query_task.py
@@ -11,6 +11,9 @@ class TextQueryTask(BaseTextInputTask):
     query_engine: BaseQueryEngine = field(kw_only=True)
     loader: TextLoader = field(default=Factory(lambda: TextLoader()), kw_only=True)
     namespace: Optional[str] = field(default=None, kw_only=True)
+    top_n: Optional[int] = field(default=None, kw_only=True)
 
     def run(self) -> TextArtifact:
-        return self.query_engine.query(self.input.to_text(), namespace=self.namespace, rulesets=self.all_rulesets)
+        return self.query_engine.query(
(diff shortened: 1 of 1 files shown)


griptape/tasks/text_query_task.py
```diff
@@ -11,6 +11,9 @@ class TextQueryTask(BaseTextInputTask):
     query_engine: BaseQueryEngine = field(kw_only=True)
     loader: TextLoader = field(default=Factory(lambda: TextLoader()), kw_only=True)
     namespace: Optional[str] = field(default=None, kw_only=True)
+    top_n: Optional[int] = field(default=None, kw_only=True)
 
     def run(self) -> TextArtifact:
-        return self.query_engine.query(self.input.to_text(), namespace=self.namespace, rulesets=self.all_rulesets)
+        return self.query_engine.query(
+            self.input.to_text(), namespace=self.namespace, rulesets=self.all_rulesets, top_n=self.top_n
```
andrewfrench, line 18: I believe this is failing static analysis because `query_engine` is a `BaseQueryEngine` that does not support `top_n` in its `query()` method. I think you'll need to update the base class to match.

andrewfrench (changes requested):
Thanks @igor-2lemetry! One small change to make our type checker happy.

collindutter (approved):

andrewfrench (approved):

collindutter:
@andrewfrench i'm actually still getting the `top_n` pyright error, was the test passing when you merged?

andrewfrench:
@collindutter it was, but that may be because we were only checking changed files and the blast radius exceeds that. I'll take a look.

PR #636, opened 2024-02-05
Create Griptape Cloud Knowledge Base Client

collindutter:


griptape/tools/griptape_cloud_knowledge_base_client/tool.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Attributes:
+        description: LLM-friendly knowledge base description.
+        base_url: Base URL for the Griptape Cloud Knowledge Base API.
+        api_key: API key for Griptape Cloud.
+        headers: Headers for the Griptape Cloud Knowledge Base API.
+        knowledge_base_id: ID of the Griptape Cloud Knowledge Base.
+    """
+
+    description: str = field(kw_only=True)
```
vasinov, line 21: Can we pull description dynamically with an optional override (what I have in the POC)? Could be a very nice UX feature for true end-to-end.

vasinov (changes requested):

griptape/tools/griptape_cloud_knowledge_base_client/tool.py
collindutter: Added

vasinov (approved):

PR #638, opened 2024-02-06
Fix index name retrieval on Marqo Index object

andrewfrench:
This PR fixes our final Pyright error by correcting how we grab Marqo index names from Index objects.


collindutter:
What was reasoning for closing?

andrewfrench:
@collindutter We ran Pyright on my machine to confirm zero errors after the last batch of fixes and found one error remaining. It turned out the error was caused by my downgrading the version of our Marqo dependency to reproduce a Discord issue and never reverting that change. There were zero errors after re-upgrading to a recent version of Marqo.

PR #639, opened 2024-02-06
Run pyright on all framework files

andrewfrench:
A change in one file can cause Pyright errors in other files.

Previously, we only checked files touched in the diff because there were hundreds of unrelated errors that would be hard to ignore. Now that we have a baseline of zero Pyright errors, we should check the entire project to ensure new changes don't break any associated code.


collindutter (approved):

collindutter:
Nice catch

Issue #641, opened 2024-02-07
Make preamble in vector query prompt customisable


igor-2lemetry:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
This is a good widely used approach to have a preamble in the very beginning to let LLM to take up a role, before providing the rest of data like user query, contex, etc. It is great to have a default value for it, but many cases and advances users it may matter. So, I believe it is important to make it customisable.

**Describe the solution you'd like**
Provide an ability to overwrite default value by adding new attribute to the TextQueryTask.

**Describe alternatives you've considered**
Alternative is to accept the fate and use the default value.

**Additional context**
No additional context here.

collindutter:
This issue is quite old but for visibility the recommended solution is to pass in a new value for `generate_system_template` into `PromptResponseRagModule`. The [default implementation](https://github.com/griptape-ai/griptape/blob/a9943fb4921014857841ad19490b3f91f92f10bd/griptape/engines/rag/modules/response/prompt_response_rag_module.py#L56-L65) can serve useful resource when creating a new function.

PR #642, opened 2024-02-07
Make preamble in vector query prompt customisable

igor-2lemetry:
Closes https://github.com/griptape-ai/griptape/issues/641

### DIFF
--- a/griptape/engines/query/vector_query_engine.py
+++ b/griptape/engines/query/vector_query_engine.py
@@ -23,6 +23,8 @@ class VectorQueryEngine(BaseQueryEngine):
     )
     template_generator: J2 = field(default=Factory(lambda: J2("engines/query/vector_query.j2")), kw_only=True)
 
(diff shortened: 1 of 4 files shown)


griptape/engines/query/vector_query_engine.py
```diff
@@ -23,6 +23,8 @@ class VectorQueryEngine(BaseQueryEngine):
     )
     template_generator: J2 = field(default=Factory(lambda: J2("engines/query/vector_query.j2")), kw_only=True)
```
collindutter, line 24: While I do like the granularity of being able to override specific bits of the prompt, I'm not sure this change is necessary since users can override `template_generator` to completely change the prompt.

@andrewfrench @vasinov do you have any suggestions for how we might improve this?

collindutter:
Closing. For now the best solution is to override `template_generator`. And this process may become easier with some upcoming changes to how Prompt Stack works.

PR #658, opened 2024-02-26
Fix initialization when overriding drivers

collindutter:
Fixes logic so that if a user is providing a `prompt_driver` or `embedding_driver`, it will stick those into a generic `StructureConfig`, not an `OpenAiStructureConfig`. Additionally, if a user provides only one of `prompt_driver` or `embedding_driver`, it will default the null one to the defaults we provide in `<0.23`.

All of this can be removed after we remove the `prompt_driver` and `embedding_driver` fields for good.

### DIFF
--- a/griptape/structures/structure.py
+++ b/griptape/structures/structure.py
@@ -10,8 +10,16 @@ from attrs import Factory, define, field
 from rich.logging import RichHandler
 
 from griptape.artifacts import BlobArtifact, TextArtifact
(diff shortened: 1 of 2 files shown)


collindutter:
Also including a lock file regeneration because tests were failing without it

andrewfrench:
is this intended to fix the missing `prompt_driver` exceptions when running integration tests?

andrewfrench (approved):

andrewfrench (approved):

PR #660, opened 2024-02-27
Lock pinecone to major version

andrewfrench:


collindutter (commented):
Do we need to regenerate the lock file?

andrewfrench:
I don't think so -- we're currently using v2.2.4 on the Framework, but when installing for integration tests it pulls the newer v3.

PR #667, opened 2024-02-28
Support querying against images stored as Blob Artifacts in Task Memory

andrewfrench:
- Updates the schema for task memory retrievals for better performance
- Allows images in Task Memory to be stored as Blob Artifacts, enabling S3 as a data source


collindutter (commented):
Is this easily unit testable?

andrewfrench:
Not easily, but I am validating manually on a local project.

collindutter (approved):

PR #670, opened 2024-03-04
Add OpenAI Embedding Model 'text-embedding-3-small'

bjorger:
As OpenAI has introduced a new embedding model, I added it in the OpenAITokenizer, so I can be used.


griptape/tokenizers/openai_tokenizer.py
collindutter, line 1: Please add the model to this list as well.

collindutter, line 1: Thanks for the PR! Would you mind also adding `large` and updating the `MODEL_PREFIXES_TO_MAX_TOKENS`?
image

collindutter (changes requested):

griptape/tokenizers/openai_tokenizer.py
bjorger: hey collin, sure thing!

bjorger: Sure, will add :)

PR #674, opened 2024-03-07
Don't raise exception when action input not present

collindutter:
Reproduce:
```python
from griptape.tools import DateTime

agent = Agent(
    tools=[DateTime(off_prompt=False)],
)

agent.run("Whats the time?")
```

Output:
```
                           Thought: The user wants to know the current time. I can use the "get_current_datetime" action to get this information.

                             Action:
                             {
                               "name": "DateTime",
                               "path": "get_current_datetime"
                             }
                    INFO     Subtask 00da8251ac4e4bc19155786c505eabc7
                             Response: {'error': 'Action input parsing error: Action input not found.'}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.23.1] - 2024-03-07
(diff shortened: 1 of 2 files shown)


vasinov (approved):
I removed this and some other checks in the parallel actions PR but let's merge this in as well.

PR #685, opened 2024-03-13
Make MarkdownChunker inherit from TextChunker

collindutter:
Fixes type error of not being able to pass a `MarkdownChunker` into a `TextLoader`


vasinov (changes requested):
I think we should change the `TextLoader.chunker` type to `BaseChunker` instead. It should work with any chunker.

vasinov (approved):

PR #688, opened 2024-03-14
Fix/structure config

collindutter:
Update `StructureConfig.task_memory` to use the Drivers set in `StructureConfig.global_drivers` by default. Without this, users would run into a `DummyException` when doing something such as:

```python
import os
from griptape.drivers import AnthropicPromptDriver, OpenAiEmbeddingDriver
from griptape.tools import WebScraper, TaskMemoryClient
from griptape.config import StructureConfig, StructureGlobalDriversConfig

agent = Agent(
    config=StructureConfig(
        global_drivers=StructureGlobalDriversConfig(
            prompt_driver=AnthropicPromptDriver(
                model="claude-2.1", api_key=os.environ["ANTHROPIC_API_KEY"]
            ),
            embedding_driver=OpenAiEmbeddingDriver(),
        )
    ),
    tools=[WebScraper(off_prompt=True), TaskMemoryClient(off_prompt=False)],
)

agent.run("Tell me more about https://griptape.ai")
```

Open question for @vasinov @andrewfrench, should this change be applied to `OpenAiStructureConfig` and `AmazonBedrockStructureConfig`? Currently, their `task_memory` Drivers are set separately from their `global_drivers`.


vachillo (approved):

andrewfrench:
@collindutter wouldn't it be inconsistent behavior if we didn't apply this same change to `OpenAiStructureConfig` and `AmazonBedrockStructureConfig`? I feel that if someone chooses to use `AmazonBedrockStructureConfig` they would be okay with (and probably intend to) use those defaults for Task Memory as well.

vasinov (dismissed):

andrewfrench (dismissed):

andrewfrench (approved):

vasinov (approved):

PR #690, opened 2024-03-14
Fix/tokenizer output

collindutter:


griptape/tokenizers/base_tokenizer.py
```diff
@@ -4,10 +4,22 @@
...
-    max_tokens: int = field(kw_only=True)
+    max_tokens: int = field(kw_only=True, default=None)
+    max_output_tokens: int = field(kw_only=True, default=None)
+
+    def __attrs_post_init__(self) -> None:
+        if self.max_tokens is None:
+            self.max_tokens = self._default_max_tokens()
+
+        if self.max_output_tokens is None:
+            self.max_output_tokens = self._default_max_output_tokens()
```
collindutter, line 22: Our usual method of handling attrs defaults that depend on each other does not work well here because of partially initialized objects.

griptape/tokenizers/base_tokenizer.py
```diff
@@ -4,10 +4,22 @@
...
-@define(frozen=True)
+@define()
 class BaseTokenizer(ABC):
+    MODEL_PREFIXES_TO_MAX_OUTPUT_TOKENS = {}
+    MODEL_PREFIXES_TO_MAX_TOKENS = {}
+
+    model: str = field(kw_only=True)
     stop_sequences: list[str] = field(default=Factory(lambda: [utils.constants.RESPONSE_STOP_SEQUENCE]), kw_only=True)
-    max_tokens: int = field(kw_only=True)
+    max_tokens: int = field(kw_only=True, default=None)
```
collindutter, line 14: `max_tokens` means _input_ tokens in Tokenizers, but `max_tokens` means _output_ tokens in Drivers. Should we rename this to `max_input_tokens`?

vasinov: I think we should!

griptape/tokenizers/base_tokenizer.py
```diff
@@ -17,6 +29,32 @@ def count_tokens_left(self, text: str | list) -> int:
...
+    def count_output_tokens_left(self, text: str | list) -> int:
+        diff = self.max_output_tokens - self.count_tokens(text)
+
+        if diff > 0:
+            return diff
+        else:
+            return 0
+
     @abstractmethod
     def count_tokens(self, text: str | list[dict]) -> int:
```
vasinov, line 41: We should keep this as `count_tokens`, right? Since it can be used in any context.

collindutter: Correct

griptape/tokenizers/base_tokenizer.py
andrewfrench: Agreed, I think it would be confusing otherwise. If we're making this naming change, though, should we rip the bandaid off and change `MODEL_PREFIXES_TO_MAX_TOKENS` to `MODEL_PREFIXES_TO_MAX_INPUT_TOKENS`?

griptape/engines/summary/prompt_summary_engine.py
```diff
@@ -53,7 +53,7 @@ def summarize_artifacts_rec(
             summary=summary, text=artifacts_text, rulesets=J2("rulesets/rulesets.j2").render(rulesets=rulesets)
         )
 
-        if self.prompt_driver.tokenizer.count_tokens_left(full_text) >= self.min_response_tokens:
+        if self.prompt_driver.tokenizer.count_input_tokens_left(full_text) >= self.min_response_tokens:
```
collindutter, line 56: Am I reasoning through this logic correct? This should be `count_input_tokens_left`, not `count_output_tokens_left`, right?

griptape/tokenizers/base_tokenizer.py
collindutter: Done!

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -114,6 +114,7 @@ def _base_params(self, prompt_stack: PromptStack) -> dict:
             "stop": self.tokenizer.stop_sequences,
             "user": self.user,
             "seed": self.seed,
+            "max_tokens": self.max_output_tokens(self.prompt_stack_to_string(prompt_stack)),
```
collindutter, line 117: @vasinov @andrewfrench let me know what you think about this. I wonder if we used to get rate limited because we were specifying a `max_tokens` in the order of _input_ tokens instead of output tokens.

This is the only Prompt Driver that had this one-off hack.

andrewfrench: Oh, wow. I can totally see that being the case.

griptape/engines/summary/prompt_summary_engine.py
andrewfrench: I think you're right here, just based off the context of chunking the input and recursively summarizing.

tests/mocks/mock_tokenizer.py
```diff
@@ -1,11 +1,13 @@
...
-@define(frozen=True)
+@define()
 class MockTokenizer(BaseTokenizer):
     model: str = field(kw_only=True)
-    max_tokens: int = field(default=1000, kw_only=True)
+    max_input_tokens: int = field(default=1000, kw_only=True)
+    max_output_tokens: int = field(default=1000, kw_only=True)
 
-    def count_tokens(self, text: str) -> int:
+    def count_tokens(self, text: str | list[dict]) -> int:
```
andrewfrench, line 12: In the `list[dict]` case do we need to loop through the elements of the dict to determine text length?

collindutter: This was just to fix a pyright error. I think all instances of `MockTokenizer` pass a `str`.

andrewfrench (approved):

vasinov (approved):

PR #691, opened 2024-03-15
Add web scraper drivers

dylanholmes:
The `WebLoader` currently uses [trafilatura](https://github.com/adbar/trafilatura) to scrape the content from web pages. We would like to allow alternative web scraping implementations that do not use trafilatura because (1) trafilatura is GPL3 and therefore may not be appropriate for all use cases and (2) does not work well for all web pages (particularly those with dynamically loaded content).

Changes:
- Add a `BaseWebScraperDriver`
- Add `TrafilaturaWebScraperDriver`, the original implementation for backwards compatibility.
- Add `MarkdownWebScraperDriver`, a new implementation that uses [playwright](https://github.com/microsoft/playwright-python) to render webpages using a headless browser, then [beautiful soup](https://www.crummy.com/software/BeautifulSoup/) + [markdownify](https://github.com/matthewwithanm/python-markdownify) to convert the webpage into markdown and remove much of the noise (e.g. style tags and script tags)
- Edit `WebLoader` to delegate actual scraping to a `BaseWebScraperDriver`, with a default of `TrafilaturaWebScraperDriver` for backwards compatibility.
- Edit changelog

Testing:
- Add unit tests for `TrafilaturaWebScraperDriver` and `MarkdownWebScraperDriver`

Manual testing:
```python
from griptape.drivers import MarkdownifyWebScraperDriver
from griptape.loaders import WebLoader
from griptape.tools import TaskMemoryClient, WebScraper

load_dotenv()


url = "https://www.kantata.com/careers"

agent = Agent(
    tools=[
        WebScraper(
            web_loader=WebLoader(
                web_scraper_driver=MarkdownifyWebScraperDriver(timeout=1000),
            ),
            off_prompt=True,
        ),
        TaskMemoryClient(off_prompt=False),
    ],
)
agent.run(f"List all job openings at '{url}' in a flat numbered markdown list.")


# Ouput:
# ```md
# 1. Senior Functional Consultant in London / Remote, United Kingdom
# 2. Senior Functional Consultant in Remote, United States
# 3. Senior Business Intelligence Consultant in London / Remote, United Kingdom
# 4. Senior Project Manager in Remote, United States
# 5. Technical Account Manager in London / Remote, United Kingdom
# 6. Technical Account Manager in Remote, United States
# 7. Technical Account Manager in Remote, United States
# 8. Business Development Representative in London (Hybrid), United Kingdom
# 9. Business Development Representative in London (Hybrid), United Kingdom
# 10. Business Development Representative in Boston (Hybrid), United States
# 11. Business Development Representative in Irvine (Hybrid), CA, United States
# 12. Business Development Representative, German Speaking in London (Hybrid), United Kingdom
# 13. Senior Solutions Engineer in London / Remote, United Kingdom
# 14. Commercial Account Executive in London / Remote, United Kingdom
# 15. Customer Community Manager in Remote, United States
# 16. Senior Customer Success Manager in London / Remote, United Kingdom
# 17. Customer Success Manager in Remote, United States
# ```
```


griptape/drivers/web_scraper/markdown_web_scraper_driver.py
```diff
@@ -0,0 +1,61 @@
...
+
+        include_links = self.include_links
+
+        # Custom MarkdownConverter to optionally linked urls. If include_links is False only
+        # the text of the link is returned.
+        class OptionalLinksMarkdownConverter(MarkdownConverter):
+            def convert_a(self, el, text, convert_as_inline):
+                if include_links:
+                    return super().convert_a(el, text, convert_as_inline)
+                return text
```
collindutter, line 41: This should be moved up to the class level and be an attrs class. See `PromptStack.Input` for reference.

griptape/drivers/web_scraper/base_web_scraper_driver.py
```diff
@@ -0,0 +1,8 @@
+from abc import ABC, abstractmethod
+from typing import Optional
+
+
+class BaseWebScraperDriver(ABC):
+    @abstractmethod
+    def scrape_url(self, url: str, *args, **kwargs) -> str:
```
collindutter, line 7: Should probably return a `TextArtifact`.

collindutter, line 7: Do we need `*args` and `**kwargs`?

griptape/drivers/web_scraper/trafilatura_web_scraper_driver.py
```diff
@@ -0,0 +1,38 @@
...
+            raise Exception("can't access URL")
+        else:
+            extracted_page = trafilatura.extract(
+                page, include_links=self.include_links, output_format="json", config=config
+            )
+
+            if extracted_page:
+                return json.loads(extracted_page).get("text")
+            else:
+                raise Exception("can't extract page")
```
collindutter, line 38: Should we return `ErrorArtifacts`s or raise exceptions in Drivers? I don't know for sure. CC @andrewfrench @vasinov

griptape/tools/web_scraper/requirements.txt
collindutter, line 1: I forget, do we need an empty `requirements.txt` in the Tool directory?

pyproject.toml
```diff
@@ -81,10 +83,13 @@ drivers-embedding-amazon-bedrock = ["boto3"]
...
 drivers-embedding-huggingface = ["huggingface-hub", "transformers"]
 
+drivers-web-scraper-trafilatura = ["trafilatura"]
+drivers-web-scraper-markdown = ["playwright", "beautifulsoup4", "markdownify"]
+
 loaders-dataframe = ["pandas"]
 loaders-pdf = ["pypdf"]
 loaders-image = ["pillow"]
-loaders-web = ["trafilatura"]
+loaders-web = []
```
collindutter, line 92: Can we just remove this entirely?

griptape/drivers/web_scraper/markdown_web_scraper_driver.py
```diff
@@ -0,0 +1,61 @@
+from typing import Optional
+from attr import define, field, Factory
+from griptape.drivers import BaseWebScraperDriver
+from griptape.utils import import_optional_dependency
+
+
+@define
+class MarkdownWebScraperDriver(BaseWebScraperDriver):
```
collindutter, line 8: Not sure how I feel about the name `MarkdownWebScraperDriver`. The Driver names usually imply the tools they use, not their output format. Is there a reason we're returning in markdown for this Driver but not others?

collindutter (commented):
Nice work!

griptape/tools/web_scraper/requirements.txt
dylanholmes: There is no requirements.txt under other tools (e.g. `aws_iam_client` and `computer`), so I'm assuming it is not required.

griptape/drivers/web_scraper/markdown_web_scraper_driver.py
dylanholmes: The base `MarkdownConverter` class comes from an optional dependency. I'm not sure if there is a way to move it up without making the optional importing mechanism more complicated. Let me know what you think

collindutter: Ah didn't see that, a necessary evil in that case.

griptape/drivers/web_scraper/markdown_web_scraper_driver.py
dylanholmes: About the naming:

The `TrafilaturaWebScraperDriver` only relies on a single dependency (`Trafilatura`, obviously) to fetch the web page and to extract the meaningful content from it, so here the naming is pretty straight forward, but with this new web scraper, it's using three different dependencies, `playwright` to fetch the web page, `beautifulsoup4` to remove to filter out some of the html, and `markdownify` to convert the html to markdown. Using a name to match the functionality is a little awkward here, since it should effectively be called `PlaywrightBeautifulSoupMarkdownifyWebScraperDriver`, which is horrible.

---

About the output formatting:

The `TrafilaturaWebScraperDriver` currently does apply markdown formatting for links. The class has an `include_links` parameter, which if true (the default) links are converted into markdown (e.g. `<a href="foo.com">foo</a>` to `[foo](foo.com`).

Replicating this behavior with beautiful soup alone is rather low level and doesn't seem worth the trouble. That said,  after seeing the markdown outputted by chat gpt when I told it to 'Extract the meaningful content from this html`, it seems like retaining some minimal document structure would be a good trade off between a conciseness and accuracy (plus its much more human readable if you are debugging). In other words, it seems like the purpose of our web scrapers is to extract the meaningful content from the webpage, and I think markdown is a good choice for that. Let me know if you think otherwise

griptape/drivers/web_scraper/base_web_scraper_driver.py
dylanholmes: I believe that @vasinov chose this interface to keep the options open in the derived classes, though I think we should be able to pass them via the constructor args instead.

Actually this is what we did in the derived classes, I'll either need to remove the `include_links` parameter from `Webloader.extract_page`, or start using it in the derived classes and remove the constructor args.

I would prefer moving `include_links` to a constructor arg because it is not determined via a parameter in the `WebScraper` tool, so I see a reason to pass it to `Webloader.extract_page` on each invocation.

dylanholmes: Updated

dylanholmes: Removed `*args` and `**kwargs` from method signature

pyproject.toml
dylanholmes: Removed

griptape/drivers/web_scraper/markdown_web_scraper_driver.py
dylanholmes: Renamed to `MarkdownifyWebScraperDriver` in order follow the convention without being too verbose.

dylanholmes:
Note that I add some manual testing to the overview (with links to gists that also serve as example usages)

griptape/tools/web_scraper/tool.py
```diff
@@ -10,7 +9,7 @@
 
 @define
 class WebScraper(BaseTool):
-    include_links: bool = field(default=True, kw_only=True)
+    web_loader: WebLoader = field(default=Factory(lambda: WebLoader()))
```
vasinov, line 12: Should include `kw_only=True`.

collindutter (approved):

griptape/tools/web_scraper/tool.py
dylanholmes: Updated

dylanholmes:
I updated the change log in most recent revision

collindutter (approved):

PR #695, opened 2024-03-19
Edit MarkdownifyWebScraperDriver

dylanholmes:
After reviewing the output from the example usage I was adding to the docs, I wanted to remove unsightly triple+ newlines from the output markdown. While I made I was making the change, I thought it would be a good idea to also decrease the assumptions made in the default `exclude_tags` list. Then, while submitting the PR, then I realize that I had needed to update `poetry.lock` one more time for [Add web scraper drivers #691](https://github.com/griptape-ai/griptape/pull/691/files). (Surprised this wasn't caught by the PR checks...)

Changes:
- Edit `poetry.lock` via `poetry lock --no-update`. This was missed after renaming the `drivers-web-scraper-markdown` extra to `drivers-web-scraper-markdownify` in `pyproject.toml`.
- Remove `header`, `footer`, and `svg` from default `exclude_tags`, since it is reasonable to assume that it may contain meaningful content. (`svg`s sometimes contain text too)
- Add a constant to hold the default `exclude_tags` value. This makes it easier for client code to add exclude tags without duplicating the defaults. (e.g. they can now do  `exclude_tags = DEFAULT_EXCLUDE_TAGS + ["foo"]`)
- Remove trailing whitespace from lines.
- Indent using 2 spaces instead of tabs for more human readable output. (Like tabs are human readable too, but they depend on you tab size, so it can be ugly if you don't normally use tabs, and indenting with spaces is more of the convention we are following with our own use of markdown).
- Limit occurrences contiguous newline to 2. By default markdownify does not do this and it can lead to unsightly whitespace in the output. We need to allow 2 contiguous newlines to separate paragraphs.

Testing:
- Compared [output before change](https://gist.github.com/dylanholmes/b81dbee4b0632ccebd698db1592f10c2) and [output after change](https://gist.github.com/dylanholmes/7b1712ed95e1b8fa822b81f66ca4ce9c) for url: https://docs.griptape.ai/latest/griptape-framework/drivers/sql-drivers/

### DIFF
--- a/griptape/drivers/web_scraper/markdownify_web_scraper_driver.py
+++ b/griptape/drivers/web_scraper/markdownify_web_scraper_driver.py
@@ -1,3 +1,4 @@
+import re
 from typing import Optional
 from attr import define, field, Factory
 from griptape.artifacts import TextArtifact
@@ -22,9 +23,11 @@ class MarkdownifyWebScraperDriver(BaseWebScraperDriver):
             the browser has emitted the "load" event.
     """
 
+    DEFAULT_EXCLUDE_TAGS = ["script", "style", "head"]
+
     include_links: bool = field(default=True, kw_only=True)
     exclude_tags: list[str] = field(
-        default=Factory(lambda: ["script", "style", "head", "header", "footer", "svg"]), kw_only=True
+        default=Factory(lambda self: self.DEFAULT_EXCLUDE_TAGS, takes_self=True), kw_only=True
     )
     exclude_classes: list[str] = field(default=Factory(list), kw_only=True)
     exclude_ids: list[str] = field(default=Factory(list), kw_only=True)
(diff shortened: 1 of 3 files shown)


griptape/drivers/web_scraper/markdownify_web_scraper_driver.py
```diff
@@ -1,10 +1,14 @@
...
 from typing import Optional
 from attr import define, field, Factory
 from griptape.artifacts import TextArtifact
 from griptape.drivers import BaseWebScraperDriver
 from griptape.utils import import_optional_dependency
 
 
+DEFAULT_EXCLUDE_TAGS = ["script", "style", "head"]
+
+
```
collindutter, line 11: Let's move this into the class.

griptape/drivers/web_scraper/markdownify_web_scraper_driver.py
```diff
@@ -23,9 +27,7 @@ class MarkdownifyWebScraperDriver(BaseWebScraperDriver):
     """
 
     include_links: bool = field(default=True, kw_only=True)
-    exclude_tags: list[str] = field(
-        default=Factory(lambda: ["script", "style", "head", "header", "footer", "svg"]), kw_only=True
-    )
+    exclude_tags: list[str] = field(default=Factory(lambda: DEFAULT_EXCLUDE_TAGS), kw_only=True)
```
collindutter, line 30: Will need to add `takes_self=True` after moving `DEFAULT_EXCLUDE_TAGS` into class.

collindutter (changes requested):
Just minor thing

collindutter (approved):

PR #696, opened 2024-03-19
Add filter to pgvector query

zachgiordano:
Adding the capability to pass in an arbitrary `filter` dict to the query engine. 

* Changes
  * BaseQueryEngine and VectorQueryEngine accept an optional `filter` dict parameter in the `query` method
  * The PgVectorVectorStoreDriver utilizes kwargs in the instantiation of its model, and checks for a `filter` kwarg to use in the `filter_by` method

### DIFF
--- a/griptape/drivers/vector/pgvector_vector_store_driver.py
+++ b/griptape/drivers/vector/pgvector_vector_store_driver.py
@@ -10,6 +10,7 @@ from sqlalchemy import create_engine, Column, String, JSON
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.orm import Session
+from collections import OrderedDict
 
 
 @define
@@ -82,7 +83,7 @@ class PgVectorVectorStoreDriver(BaseVectorStoreDriver):
     ) -> str:
         """Inserts or updates a vector in the collection."""
         with Session(self.engine) as session:
-            obj = self._model(id=vector_id, vector=vector, namespace=namespace, meta=meta)
+            obj = self._model(id=vector_id, vector=vector, namespace=namespace, meta=meta, **kwargs)
 
             obj = session.merge(obj)
             session.commit()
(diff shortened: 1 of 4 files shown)


tests/unit/drivers/vector/test_pgvector_vector_store_driver.py
```diff
@@ -33,7 +33,7 @@ def test_initialize_requires_engine_or_connection_string(self, embedding_driver)
 
     def test_initialize_accepts_engine(self, embedding_driver):
         engine = create_engine(self.connection_string)
-        driver = PgVectorVectorStoreDriver(embedding_driver=embedding_driver, engine=engine, table_name=self.table_name)
+        driver = PgVectorVectorStoreDriver(embedding_driver=embedding_driver, engine=engine, table_name=self.table_name)  # type: ignore
```
collindutter, line 36: Reason for the type ignore?

zachgiordano: It seems pyright doesn't like the mock engine. Is there another way to satisfy pyright?

```
/Users/zach/workplace/Griptape/griptape/tests/unit/drivers/vector/test_pgvector_vector_store_driver.py:36:86 - error: Argument of type "MockConnection | Unknown" cannot be assigned to parameter "engine" of type "Engine | None" in function "__init__" (reportGeneralTypeIssues)
```

zachgiordano: Actually it looks like typing the mock engine as `Any` does the trick:

```
engine: Any = create_engine(self.connection_string)
```

Updated

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -148,8 +149,19 @@ def query(
...
-                query_result = query_result.filter_by(namespace=namespace)
+                filter_kwargs = OrderedDict(namespace=namespace)
+
+            if "filter" in kwargs and isinstance(kwargs["filter"], dict):
+                if filter_kwargs:
+                    filter_kwargs.update(kwargs["filter"])
+                else:
+                    filter_kwargs = OrderedDict(kwargs["filter"])
+
+            if filter_kwargs:
```
collindutter, line 163: Be explicit when checking for `None` https://google.github.io/styleguide/pyguide.html#2144-decision

griptape/drivers/vector/pgvector_vector_store_driver.py
```diff
@@ -148,8 +149,19 @@ def query(
...
+                query_result = query_result.filter_by(**filter_kwargs)
```
collindutter, line 164: Can we simplify this entire thing to something like:

```python
        if "filter" in kwargs and isinstance(kwargs["filter"], dict):
                query_result.filter_by(OrderedDict(kwargs["filter"]))
```

poetry.lock
collindutter, line 1: Why do we need to update the poetry lock? The version on `dev` is passing tests, right?

collindutter (changes requested):

poetry.lock
zachgiordano: Fixed by this [PR](https://github.com/griptape-ai/griptape/pull/697). I'll merge into mine

griptape/drivers/vector/pgvector_vector_store_driver.py
zachgiordano: I want the `namespace` to be added to the OrderedDict first if it is present so that it is applied first. But I simplified this logic with the most recent commit update, so please take a look.

collindutter (approved):

PR #694, opened 2024-03-18
Add Claude 3 Support

collindutter:

PR #700, opened 2024-03-19
Claude 3 Image Query Support

emjay07:
Adding Claude 3 Image Query support via both Anthropic and Bedrock.

```
image_query_driver=AnthropicImageQueryDriver(
    api_key=anthropic_api_key,
    model="claude-3-sonnet-20240229"
)
```

```
image_query_driver=AmazonBedrockImageQueryDriver(
    image_query_model_driver=BedrockClaudeImageQueryModelDriver(),
    model="anthropic.claude-3-sonnet-20240229-v1:0"
)
```


collindutter:
Once things stabilize on a handful of PRs we should add the vision drivers to https://github.com/griptape-ai/griptape/pull/702

pyproject.toml
```diff
@@ -34,7 +34,7 @@ boto3-stubs = {extras = ["bedrock", "iam", "opensearch", "s3", "sagemaker"], ver
 
 # drivers
 cohere = { version = ">=4", optional = true }
-anthropic = { version = "^0.7.8", optional = true }
+anthropic = { version = "^0.17.0", optional = true }
```
collindutter, line 37: Lets bump this up to be in sync with the [other PR](https://github.com/griptape-ai/griptape/pull/694/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R37)

griptape/drivers/image_query/anthropic_image_query_driver.py
```diff
@@ -0,0 +1,64 @@
...
+class AnthropicImageQueryDriver(BaseImageQueryDriver):
+    """
+    Attributes:
+        api_key: Anthropic API key.
+        model: Anthropic model name.
+        client: Custom `Anthropic` client.
+        max_output_tokens: Max output tokens to return.
+    """
+
+    api_key: str = field(kw_only=True, metadata={"serializable": True})
```
vasinov, line 22: This should have a default set to `None`, so that it can be automatically loaded from an environment variable.

griptape/drivers/image_query/anthropic_image_query_driver.py
```diff
@@ -0,0 +1,64 @@
...
+    model: str = field(default="claude-3-sonnet-20240229", kw_only=True, metadata={"serializable": True})
+    client: Anthropic = field(
+        default=Factory(
+            lambda self: import_optional_dependency("anthropic").Anthropic(api_key=self.api_key), takes_self=True
+        ),
+        kw_only=True,
+    )
+    max_output_tokens: Optional[int] = field(default=4096, kw_only=True, metadata={"serializable": True})
+
+    def try_query(self, query: str, images: list[ImageArtifact]) -> TextArtifact:
```
vasinov, line 32: Could you normalize empty lines in this method (and other class methods)? Assignments shouldn't be separated by empty lines but groups of assignments should be separated from method calls and for/if structures for better readability.

vasinov (changes requested):

griptape/drivers/image_query/anthropic_image_query_driver.py
emjay07: I'm good on most of these. I have always followed a standard though if you are initializing a variable (like an empty list) that is specifically going to be populated by a for-loop or if/else statement, then that remains "grouped" with that statement. Thoughts?

griptape/drivers/image_query/anthropic_image_query_driver.py
emjay07: I was following a standard that was set in the [AnthropicPromptDriver](https://github.com/griptape-ai/griptape/blob/dev/griptape/drivers/prompt/anthropic_prompt_driver.py#L24). I can update there as well to stay consistent? but I definitely think they need to follow the same pattern.

emjay07: I see Collin updated that in his [structure PR](https://github.com/griptape-ai/griptape/pull/702/files#diff-388135d66c742b44c2b3cd526877c0d0d34ca1938daa41523ade65a86a3f24f5R21) will follow that standard.

griptape/drivers/image_query/amazon_bedrock_image_query_driver.py
```diff
@@ -0,0 +1,35 @@
...
+    def try_query(self, query: str, images: list[ImageArtifact]) -> TextArtifact:
+        payload = self.image_query_model_driver.construct_image_query_request_parameters(query, images)
+
+        response = self.bedrock_client.invoke_model(
+            modelId=self.model, contentType="application/json", accept="application/json", body=json.dumps(payload)
+        )
+
+        response_body = json.loads(response.get("body").read())
+
+        if not response_body:
```
collindutter, line 29: Explicitly check for `is None`. https://google.github.io/styleguide/pyguide.html#2144-decision

griptape/drivers/image_query/anthropic_image_query_driver.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Attributes:
+        api_key: Anthropic API key.
+        model: Anthropic model name.
+        client: Custom `Anthropic` client.
+        max_output_tokens: Max output tokens to return.
+    """
+
+    api_key: str = field(kw_only=True, metadata={"serializable": True})
+    model: str = field(default="claude-3-sonnet-20240229", kw_only=True, metadata={"serializable": True})
```
collindutter, line 23: Remove default model value.

griptape/drivers/image_query/anthropic_image_query_driver.py
```diff
@@ -0,0 +1,64 @@
...
+            lambda self: import_optional_dependency("anthropic").Anthropic(api_key=self.api_key), takes_self=True
+        ),
+        kw_only=True,
+    )
+    max_output_tokens: Optional[int] = field(default=4096, kw_only=True, metadata={"serializable": True})
+
+    def try_query(self, query: str, images: list[ImageArtifact]) -> TextArtifact:
+        content = []
+        for image in images:
+            content.append(self._construct_image_message(image))
```
collindutter, line 35: This is a good place for list comprehension

griptape/drivers/image_query/anthropic_image_query_driver.py
```diff
@@ -0,0 +1,64 @@
...
+
+    api_key: str = field(kw_only=True, metadata={"serializable": True})
+    model: str = field(default="claude-3-sonnet-20240229", kw_only=True, metadata={"serializable": True})
+    client: Anthropic = field(
+        default=Factory(
+            lambda self: import_optional_dependency("anthropic").Anthropic(api_key=self.api_key), takes_self=True
+        ),
+        kw_only=True,
+    )
+    max_output_tokens: Optional[int] = field(default=4096, kw_only=True, metadata={"serializable": True})
```
collindutter, line 30: Do we want to provide a default value here?

griptape/drivers/image_query/anthropic_image_query_driver.py
```diff
@@ -0,0 +1,64 @@
...
+        for image in images:
+            content.append(self._construct_image_message(image))
+
+        content.append(self._construct_text_message(query))
+        messages = self._construct_messages(content)
+
+        params = {"model": self.model, "messages": messages}
+
+        if self.max_output_tokens is not None:
+            params["max_tokens"] = self.max_output_tokens
```
collindutter, line 43: Can we move this logic to a separate method like the Prompt Drivers have? Something like `OpenAiChatPromptDriver._base_params`

griptape/drivers/image_query_model/base_image_query_model_driver.py
```diff
@@ -0,0 +1,16 @@
...
+from abc import ABC, abstractmethod
+from attr import define
+from griptape.artifacts import TextArtifact, ImageArtifact
+from griptape.mixins import SerializableMixin
+
+
+@define
+class BaseImageQueryModelDriver(SerializableMixin, ABC):
+    @abstractmethod
+    def construct_image_query_request_parameters(self, query: str, images: list[ImageArtifact]) -> dict:
```
collindutter, line 11: For consistency with `BaseImageGenerationModelDriver`, I think we should shorten to `image_query_request_parameters`

griptape/drivers/image_query_model/bedrock_claude_image_query_model_driver.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@define
+class BedrockClaudeImageQueryModelDriver(BaseImageQueryModelDriver):
+    """
+    Attributes:
+        max_output_tokens: Max output tokens to return.
+    """
+
+    max_output_tokens: Optional[int] = field(default=4096, kw_only=True, metadata={"serializable": True})
```
collindutter, line 15: Do we want a default set here?

griptape/drivers/image_query_model/bedrock_claude_image_query_model_driver.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def construct_image_query_request_parameters(self, query: str, images: list[ImageArtifact]) -> dict:
+        content = []
+        for image in images:
+            content.append(self._construct_image_message(image))
```
collindutter, line 20: Good candidate for list comprehension

griptape/drivers/image_query_model/bedrock_claude_image_query_model_driver.py
```diff
@@ -0,0 +1,50 @@
...
+
+        content.append(self._construct_text_message(query))
+        messages = self._construct_messages(content)
+
+        input_params = {"messages": messages, "anthropic_version": "bedrock-2023-05-31"}
```
collindutter, line 25: Let's make `anthropic_version` an field variable for consistency with `AnthropicPromptDriver`

collindutter (changes requested):

griptape/drivers/image_query/anthropic_image_query_driver.py
emjay07: Can do, but I was curious about this. I saw we set defaults for OpenAI drivers, but not others. Why is that?

griptape/drivers/image_query/anthropic_image_query_driver.py
emjay07: Anthropic requires *some* value to be specified. I tried with None and it gets angry. So I thought a default would be best because then it's more upfront to the customer vs. hiding it in our internal call. Thoughts?

griptape/drivers/image_query_model/bedrock_claude_image_query_model_driver.py
emjay07: I think we should stay consistent with whatever we decide on the native `AnthropicImageQueryDriver`. So whatever comes out of that conversation I'll update here as well.

griptape/drivers/image_query/anthropic_image_query_driver.py
emjay07: All of these cases are no longer relevant due to changing to list comprehension, so no longer need to discuss this.

griptape/drivers/image_query/anthropic_image_query_driver.py
collindutter: Got it, thanks Anthropic. If we do this, we should probably standardize across the Query Drivers. OpenAi does not have one set, for instance.

Any thoughts @andrewfrench?

griptape/drivers/image_query_model/bedrock_claude_image_query_model_driver.py
collindutter: Ah sorry I meant this guy: https://github.com/griptape-ai/griptape/blob/dev/griptape/drivers/prompt_model/bedrock_claude_prompt_model_driver.py#L17.

Agreed that it probably won't change on us, but I think it should be moved to at least a static class field since it's kind of a magic string. Maybe we don't need to expose it to the user via an instance field.

griptape/drivers/image_query/anthropic_image_query_driver.py
andrewfrench: OpenAI's default no-value behavior is unhelpful enough as to consider it a required parameter as well. Perhaps we should update both the OpenAI and Anthropic drivers to explicitly require this field? In either case, a refactor of the tokenizers + prompt stack to support image and media inputs will go a long way to help here.

andrewfrench (commented):
Spent some time this morning playing with Claude 3 image queries with great success, awesome stuff! If we can choose a path on [this conversation](https://github.com/griptape-ai/griptape/pull/700#discussion_r1532893642) I think we're good to go.

griptape/drivers/image_query/anthropic_image_query_driver.py
emjay07: I'm good with staying consistent across all `image_query` drivers and making `max_output_tokens` a required value for now. That means no default and put the onus on the customer, or should we provide a "middle-of-the-road" default for both OpenAI and Anthropic? I was thinking yes to still providing a default so that customers can use it out-of-the-box without needing to go deeply understand each LLMs tokenization. Thoughts?

griptape/drivers/image_query/anthropic_image_query_driver.py
collindutter: Ah I didn't realize we set in openai. That probably shouldn't be set their either. But for sake of consistency we can keep it.

@andrewfrench thoughts?

emjay07:
> Once things stabilize on a handful of PRs we should add the vision drivers to #702

Added these in

griptape/drivers/image_query_model/bedrock_claude_image_query_model_driver.py
emjay07: I made it a static field and updated `BedrockClaudePromptModelDriver` as well to stay consistent.

griptape/drivers/image_query/anthropic_image_query_driver.py
emjay07: Put `max_output_tokens` on the base class and made the default 256.

griptape/drivers/image_query/anthropic_image_query_driver.py
emjay07: I had already removed the anthropic default. I can remove the openai default as well in the name of consistency.

andrewfrench (approved):

griptape/drivers/image_query/anthropic_image_query_driver.py
collindutter: Let's keep it named as `max_tokens` for consistency with the Prompt Drivers. We can do a full rename in the future.

griptape/drivers/prompt_model/bedrock_claude_prompt_model_driver.py
```diff
@@ -14,7 +14,7 @@ class BedrockClaudePromptModelDriver(BasePromptModelDriver):
     top_k: int = field(default=250, kw_only=True, metadata={"serializable": True})
     _tokenizer: BedrockClaudeTokenizer = field(default=None, kw_only=True)
     prompt_driver: Optional[AmazonBedrockPromptDriver] = field(default=None, kw_only=True)
-    anthropic_version: str = field(default="bedrock-2023-05-31", kw_only=True, metadata={"serializable": True})
+    anthropic_version: str = "bedrock-2023-05-31"  # static string for AWS: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html#api-inference-examples-claude-multimodal-code-example
```
collindutter, line 17: If we're treating this as a class field, it should be `ANTHROPIC_VERSION = "bedrock-2023-05-31"`. But I think we should keep it as an attrs instance field for consistency with the other implementation.

griptape/drivers/image_query_model/bedrock_claude_image_query_model_driver.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
+from typing import Optional
+from attr import field, define
+from griptape.artifacts import ImageArtifact, TextArtifact
+from griptape.drivers import BaseImageQueryModelDriver
+
+
+@define
+class BedrockClaudeImageQueryModelDriver(BaseImageQueryModelDriver):
+    anthropic_version: str = "bedrock-2023-05-31"  # static string for AWS: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html#api-inference-examples-claude-multimodal-code-example
```
collindutter, line 10: Same comment as below.

griptape/drivers/image_query/base_image_query_driver.py
```diff
@@ -16,6 +16,7 @@
 @define
 class BaseImageQueryDriver(SerializableMixin, ExponentialBackoffMixin, ABC):
     structure: Optional[Structure] = field(default=None, kw_only=True)
+    max_output_tokens: int = field(default=256, kw_only=True, metadata={"serializable": True})
```
collindutter, line 19: Lets keep it named as `max_tokens` for consistency with other Drivers.

CHANGELOG.md
```diff
@@ -33,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `BaseTextLoader` to accept a `BaseChunker`.
 - Default model of `AmazonBedrockStructureConfig` to `anthropic.claude-3-sonnet-20240229-v1:0`.
 - `AnthropicPromptDriver` and `BedrockClaudePromptModelDriver` to use Anthropic's Messages API.
-
+- `OpenAIVisionImageQueryDriver` now has a required field `max_output_tokens` that defaults to 256
```
collindutter, line 38: Nit: `OpenAiVisionImageQueryDriver`

griptape/drivers/image_query/amazon_bedrock_image_query_driver.py
```diff
@@ -0,0 +1,35 @@
...
+
+        response_body = json.loads(response.get("body").read())
+
+        if response_body is None:
+            raise ValueError("Model response is empty")
+
+        try:
+            return self.image_query_model_driver.process_output(response_body)
+        except Exception as e:
+            raise e
```
collindutter, line 35: This try/except is redundant if we're just immediately reraising.

griptape/drivers/prompt_model/bedrock_claude_prompt_model_driver.py
emjay07: Which other implementation? I thought we discussed making this a static class field?

griptape/drivers/image_query/amazon_bedrock_image_query_driver.py
emjay07: Ah, I actually meant to update that in the exception case.

griptape/drivers/prompt_model/bedrock_claude_prompt_model_driver.py
collindutter: Oh shoot, I didn't realize the file I was looking at. Ok then yeah let's just change it to `ANTHROPIC_VERSION = "bedrock-2023-05-31"` (no type) before the rest of the fields.

griptape/drivers/image_query/base_image_query_driver.py
collindutter: Agreed that `max_output_tokens` is probably a better field name, but the other PR was only in the context of the Tokenizers, not the Drivers.

emjay07: All the other drivers have inputs as text, not images. so to clarify my question: are the other drivers that have `max_tokens` intended to be for the output tokens or the input tokens?

emjay07: Updated back to `max_tokens`

griptape/drivers/image_query/anthropic_image_query_driver.py
emjay07: Done.

vasinov (approved):

andrewfrench (approved):

collindutter (approved):

PR #701, opened 2024-03-19
Add support for VoyageAi

collindutter:
Adds VoyageAi as an Embedding Driver since they are [recommended by Anthropic](https://docs.anthropic.com/claude/docs/embeddings#can-i-count-the-number-of-tokens-in-a-string-before-embedding-it)

### DIFF
--- a/griptape/drivers/embedding/voyageai_embedding_driver.py
+++ b/griptape/drivers/embedding/voyageai_embedding_driver.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+from typing import Optional, TYPE_CHECKING
+from attr import define, field, Factory
+from griptape.utils import import_optional_dependency
+from griptape.drivers import BaseEmbeddingDriver
+from griptape.tokenizers import VoyageAiTokenizer
+
+if TYPE_CHECKING:
+    from voyageai import Client
+
+
+@define
+class VoyageAiEmbeddingDriver(BaseEmbeddingDriver):
+    """
+    Attributes:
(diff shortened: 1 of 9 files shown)


griptape/drivers/embedding/voyageai_embedding_driver.py
```diff
@@ -0,0 +1,39 @@
...
+        api_key: API key to pass directly. Defaults to `VOYAGE_API_KEY` environment variable.
+        tokenizer: Optionally provide custom `VoyageAiTokenizer`.
+        client: Optionally provide custom VoyageAI `Client`.
+        input_type: VoyageAI input type. Defaults to `document`.
+    """
+
+    DEFAULT_MODEL = "voyage-large-2"
+
+    model: str = field(default=DEFAULT_MODEL, kw_only=True, metadata={"serializable": True})
+    api_key: Optional[str] = field(default=None, kw_only=True, metadata={"serializable": True})
```
vasinov, line 26: Should this not have a default?

vasinov (changes requested):

griptape/drivers/embedding/voyageai_embedding_driver.py
collindutter: It will pull from `VOYAGE_API_KEY` is left blank. This is consistent with how we've got OpenAi clients.

vasinov: Gotcha. Ignore my comments in the Gemini PR then as well.

vasinov (approved):

PR #705, opened 2024-03-21
Fix calculation of max tokens

collindutter:
Fixes calculation of `max_tokens` to use Prompt Stack as ChatML messages instead of Prompt Stack as string. Other tokenizers besides `OpenAiTokenizer` will still convert Prompt Stack to string for token calculation.

### DIFF
--- a/griptape/drivers/prompt/openai_chat_prompt_driver.py
+++ b/griptape/drivers/prompt/openai_chat_prompt_driver.py
@@ -123,7 +123,10 @@ class OpenAiChatPromptDriver(BasePromptDriver):
 
         messages = self._prompt_stack_to_messages(prompt_stack)
 
-        params["max_tokens"] = self.max_output_tokens(self.prompt_stack_to_string(prompt_stack))
+        if isinstance(self.tokenizer, OpenAiTokenizer):
+            params["max_tokens"] = self.max_output_tokens(messages)
+        else:
+            params["max_tokens"] = self.max_output_tokens(self.prompt_stack_to_string(prompt_stack))
         params["messages"] = messages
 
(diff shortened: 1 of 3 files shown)


griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -123,7 +123,10 @@ def _base_params(self, prompt_stack: PromptStack) -> dict:
 
         messages = self._prompt_stack_to_messages(prompt_stack)
 
-        params["max_tokens"] = self.max_output_tokens(self.prompt_stack_to_string(prompt_stack))
+        if isinstance(self.tokenizer, OpenAiTokenizer):
```
vasinov, line 126: Should we change the tokenizer type to `OpenAiTokenizer` instead?

collindutter: This came from https://github.com/griptape-ai/griptape/pull/537

dylanholmes, line 126: Small suggestion - add a helper method to reduce duplication:

```py
def _tokens(self, prompt_stack: PromptStack) -> str | list:
    if isinstance(self.tokenizer, OpenAiTokenizer):
        return self._prompt_stack_to_messages(prompt_stack)
    else:
        return self.prompt_stack_to_string(prompt_stack)
```

Then you can use this helper to set `params["max_tokens"]` here and to return token count in the `token_count` method without adding branching in two places.

dylanholmes (approved):

andrewfrench (approved):

PR #708, opened 2024-03-24
Improve TookitTask prompts for executing parallel actions

vasinov:


dylanholmes (approved):
Overall LGTM.

I like the renaming of output_label to tag.

The prompt template changes make sense, but it would be helpful to see some before and after examples demonstrating the change in behavior.

collindutter (approved):

PR #713, opened 2024-03-27
Feature/query engine improvements

collindutter:


griptape/engines/query/vector_query_engine.py
```diff
@@ -36,25 +37,25 @@ def query(
...
 
         for artifact in artifacts:
             text_segments.append(artifact.value)
 
-            message = self.template_generator.render(
+            user_message = self.user_template_generator.render(
                 metadata=metadata,
                 query=query,
                 text_segments=text_segments,
                 rulesets=J2("rulesets/rulesets.j2").render(rulesets=rulesets),
```
vasinov, line 49: Should remove rulesets from here.

griptape/engines/query/vector_query_engine.py
```diff
@@ -36,25 +37,25 @@ def query(
...
         text_segments = []
-        message = ""
+        user_message = ""
 
         for artifact in artifacts:
             text_segments.append(artifact.value)
 
-            message = self.template_generator.render(
+            user_message = self.user_template_generator.render(
                 metadata=metadata,
```
vasinov, line 46: Metadata should be in the system message.

griptape/engines/query/vector_query_engine.py
```diff
@@ -36,25 +37,25 @@ def query(
...
                 query=query,
                 text_segments=text_segments,
```
vasinov, line 48: Segments can be moved into the system message as well, so the LLM is more "grounded."

vasinov (changes requested):

griptape/engines/query/vector_query_engine.py
```diff
@@ -36,34 +37,42 @@ def query(
...
             if message_token_count + self.answer_token_offset >= tokenizer.max_input_tokens:
                 text_segments.pop()
 
-                message = self.template_generator.render(
-                    metadata=metadata,
-                    query=query,
-                    text_segments=text_segments,
-                    rulesets=J2("rulesets/rulesets.j2").render(rulesets=rulesets),
-                )
+                user_message = self.user_template_generator.render(query=query, text_segments=text_segments)
```
vasinov, line 64: Don't need to pass text_segments here.

vasinov (approved):
One minor cleanup needed but approving, so it's not blocked.

PR #718, opened 2024-03-28
Removed the `jsonschema` dependency

vasinov:
We only used `jsonschema` for JSON object validations, which we can do with `schema`.


collindutter (approved):
Nice!

collindutter (approved):

andrewfrench (approved):

PR #724, opened 2024-04-01
Output single artifact, move value to name

collindutter:
This PR makes the following changes:
- Don't return a `ListArtifact` from `ToolTask`. Consistent with old behavior.
- Move the formatted output of `ActionsSubtask` to `TextArtifact.name` instead of `TextArtifact.value`. This allows developers to have a formatted output in additional to the raw output (for parsing).

### DIFF
--- a/griptape/tasks/actions_subtask.py
+++ b/griptape/tasks/actions_subtask.py
@@ -93,8 +93,6 @@ class ActionsSubtask(BaseTextInputTask):
         self.structure.logger.info(f"Subtask {self.id}\n{self.input.to_text()}")
 
     def run(self) -> BaseArtifact:
-        format_output: Callable[[tuple[str, BaseArtifact]], str] = lambda o: f"{o[0]} output: {o[1].to_text()}"
-
         try:
             if any(a.name == "error" for a in self.actions):
                 errors = [a.input["error"] for a in self.actions if a.name == "error"]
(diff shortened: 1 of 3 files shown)


griptape/tasks/actions_subtask.py
```diff
@@ -103,7 +101,9 @@ def run(self) -> BaseArtifact:
             else:
                 results = self.execute_actions(self.actions)
 
-                self.output = ListArtifact([TextArtifact(format_output(r)) for r in results])
+                self.output = ListArtifact(
+                    [TextArtifact(name=f"{r[0]} output: {r[1].to_text()}", value=r[1].to_text()) for r in results]
```
collindutter, line 105: Is this the correct approach? Using `name` doesn't quite feel right.

griptape/tasks/tool_task.py
```diff
@@ -63,8 +63,8 @@ def run(self) -> BaseArtifact:
                 subtask.run()
                 subtask.after_run()
 
-                if subtask.output:
-                    self.output = subtask.output
+                if subtask.output and isinstance(subtask.output, ListArtifact):
```
vasinov, line 66: Can just be `if isinstance(subtask.output, ListArtifact)`.

griptape/tasks/actions_subtask.py
vasinov: I think `name` should just be `{r[0]} output`.

vasinov (changes requested):

PR #725, opened 2024-04-02
Add minItems to ToolkitTask actions schema

collindutter:


tests/unit/tasks/test_tool_task.py
```diff
@@ -10,6 +10,133 @@
 
 
 class TestToolTask:
+    TARGET_TOOLS_SCHEMA = {
```
collindutter, line 13: Originally added this to validate the existence of `minItems` in the schema, but now it is being added during prompt rendering. Can remove these tests, but I think there is value in having them.

griptape/tasks/actions_subtask.py
```diff
@@ -207,6 +207,15 @@ def __init_from_prompt(self, value: str) -> None:
...
                 if isinstance(self.origin_task, ActionsSubtaskOriginMixin):
                     self.origin_task.actions_schema().validate(actions_list)
+                if not actions_list:
+                    error_message = "Array item count 0 is less than minimum count of 1."
+                    self.structure.logger.error(
+                        f"Subtask {self.origin_task.id}\nInvalid activity input JSON: {error_message}"
+                    )
+
+                    self.actions.append(
+                        self.__error_to_action(f"Activity input JSON validation error: {error_message}")
```
vasinov, line 217: Can we raise a `schema.SchemaError` instead so it's more consistent with the rest of the validation logic?

vasinov (changes requested):

griptape/tasks/actions_subtask.py
collindutter: Good call, updated.

vasinov (approved):

PR #735, opened 2024-04-05
Feature/event listener drivers

collindutter:


griptape/utils/stream.py
```diff
@@ -53,11 +53,14 @@ def run(self, *args) -> Iterator[TextArtifact]:
         t.join()
 
     def _run_structure(self, *args):
+        from griptape.drivers import LocalEventListenerDriver
+
         def event_handler(event: BaseEvent):
             self._event_queue.put(event)
 
         stream_event_listener = EventListener(
-            event_handler, event_types=[CompletionChunkEvent, FinishPromptEvent, FinishStructureRunEvent]
```
dylanholmes, line 60: Now I see why you overrode `publish_event` in `LocalEventListenerDriver`

dylanholmes (dismissed):
🙌

griptape/utils/stream.py
collindutter: 🧵

andrewfrench (approved):
Very cool.

PR #736, opened 2024-04-05
Make loader sources more consistent

dylanholmes:
We will want to create a `BaseFileManagerDriver` to enable the `FileManager` Tool to interact with alternative (e.g. non-local) filesystems. We'd like to be able to reuse the existing loaders to load in artifacts from files, though some of the loaders used by the `FileManager` currently do not support loading files from memory, and thus would subclasses of `BaseFileManagerDriver` to write to a file before invoking the loader. While this may be beneficial in certain circumstances, it seems more natural to make the loader interfaces somewhat more consistent in the types of sources that they accept. Thus, that is the goal of this PR.

While I was making this change, I though it a good opportunity to reduce a some code duplication - by providing a single default implementation of `load_collection` for all loaders. Note that it is still beneficial to override `load_collection` in subclasses for the sake of typing (implementation just calls the super() implementation).

Changes:
- Add `to_key` method on `BaseLoader` with a reasonable implementation and move `load_collection` implementation in to the `BaseLoader`. The idea here is to encapsulate the part of `load_collection` that varies. This allows there to be a single implementation that just submits `load` to the executor for each source.
- Remove duplicate loads in `load_collection` for sources that are listed more than once in the input (or at least that have the same key).
- Edit `CsvLoader.load()` to accept `bytes | str | IO | Path` as source (previously was only `str). Note that this is a breaking change. Source as `str` used to be interpreted as a file path, but now it is interpreted as the actual file content instead. To pass a file path, source must now be typed as `Path`. This is more consistent with `TextLoader.load()`.
- Edit `ImageLoader.load()` to accept `bytes | IO | Path` as source (previously was only `bytes`).
- Edit `PdfLoader.load()` to accept `bytes | IO | Path` as source (previously was `str | IO | Path`). Similarly to the changes to `CsvLoader.load()`, `str` can no longer be used as a file path, you must pass a `Path` object. Since you can't pdfs are a binary format, the option to pass `str` was remove and replace with `bytes` for when the pdf is already in memory.
- Refactored unit tests a bit for code reuse and consistency.

### DIFF
--- a/griptape/loaders/csv_loader.py
+++ b/griptape/loaders/csv_loader.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 import csv
-from io import IOBase, StringIO
+from io import StringIO, TextIOBase
 from pathlib import Path
 from typing import IO, Optional, cast
 from collections.abc import Sequence
@@ -16,6 +16,7 @@ from griptape.loaders import BaseLoader
 class CsvLoader(BaseLoader):
     embedding_driver: Optional[BaseEmbeddingDriver] = field(default=None, kw_only=True)
     delimiter: str = field(default=",", kw_only=True)
+    encoding: str = field(default="utf-8", kw_only=True)
 
     def load(self, source: bytes | str | IO | Path, *args, **kwargs) -> list[CsvRowArtifact]:
         artifacts = []
@@ -35,13 +36,13 @@ class CsvLoader(BaseLoader):
 
     def _stream_from_source(self, source: bytes | str | IO | Path) -> IO:
         if isinstance(source, bytes):
-            return StringIO(source.decode())
+            return StringIO(source.decode(encoding=self.encoding))
         elif isinstance(source, str):
             return StringIO(source)
-        elif isinstance(source, IOBase):
+        elif isinstance(source, TextIOBase):
             return cast(IO, source)
         elif isinstance(source, Path):
-            return open(source, encoding="utf-8")
+            return open(source, encoding=self.encoding)
         else:
             raise ValueError(f"Unsupported source type: {type(source)}")
 
--- a/griptape/loaders/pdf_loader.py
+++ b/griptape/loaders/pdf_loader.py
@@ -30,8 +30,6 @@ class PdfLoader(BaseTextLoader):
     def _stream_from_source(self, source: bytes | IO | Path) -> IO:
         if isinstance(source, bytes):
             return BytesIO(source)
-        elif isinstance(source, str):
-            return open(source, "rb")
         elif isinstance(source, (RawIOBase, BufferedIOBase)):
             return cast(IO, source)
         elif isinstance(source, Path):
--- a/griptape/loaders/text_loader.py
+++ b/griptape/loaders/text_loader.py
@@ -45,7 +45,7 @@ class TextLoader(BaseTextLoader):
 
     def _stream_from_source(self, source: bytes | str | IO | Path) -> IO:
         if isinstance(source, bytes):
-            return StringIO(source.decode())
+            return StringIO(source.decode(encoding=self.encoding))
         elif isinstance(source, str):
             return StringIO(source)
         elif issubclass(type(source), TextIOBase):
--- a/tests/unit/loaders/test_csv_loader.py
+++ b/tests/unit/loaders/test_csv_loader.py
@@ -4,9 +4,13 @@ from tests.mocks.mock_embedding_driver import MockEmbeddingDriver
 
 
 class TestCsvLoader:
-    @pytest.fixture
-    def loader(self):
-        return CsvLoader(embedding_driver=MockEmbeddingDriver())
+    @pytest.fixture(params=["ascii", "utf-8", None])
+    def loader(self, request):
+        encoding = request.param
+        if encoding is None:
+            return CsvLoader(embedding_driver=MockEmbeddingDriver())
+        else:
+            return CsvLoader(embedding_driver=MockEmbeddingDriver(), encoding=encoding)
 
     @pytest.fixture
     def loader_with_pipe_delimiter(self):
--- a/tests/unit/loaders/test_text_loader.py
+++ b/tests/unit/loaders/test_text_loader.py
@@ -11,7 +11,11 @@ MAX_TOKENS = 50
 class TestTextLoader:
     @pytest.fixture(params=["ascii", "utf-8", None])
     def loader(self, request):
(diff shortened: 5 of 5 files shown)


griptape/loaders/base_text_loader.py
```diff
@@ -33,19 +33,16 @@ class BaseTextLoader(BaseLoader, ABC):
     embedding_driver: Optional[BaseEmbeddingDriver] = field(default=None, kw_only=True)
     encoding: str = field(default="utf-8", kw_only=True)
 
-    def _text_to_artifacts(self, text: str | Path) -> list[TextArtifact]:
-        artifacts = []
+    def load_collection(self, sources: Sequence[Any], *args, **kwargs) -> dict[str, list[TextArtifact]]:
+        return cast(Any, super().load_collection(sources, *args, **kwargs))
```
collindutter, line 37: Is a overridden method that immediately calls `super` necessary?

collindutter, line 37: Can we refine the cast type from `Any` to `dict[str, list[TextArtifact]]`?

griptape/loaders/csv_loader.py
```diff
@@ -37,3 +32,20 @@ def _load_file(self, filename: str) -> list[CsvRowArtifact]:
...
+            return cast(IO, source)
+        elif isinstance(source, Path):
+            return open(source, encoding="utf-8")
+        else:
+            raise ValueError(f"Unsupported source type: {type(source)}")
+
+    def load_collection(
+        self, sources: Sequence[bytes | str | IO | Path], *args, **kwargs
+    ) -> dict[str, list[CsvRowArtifact]]:
+        return cast(Any, super().load_collection(sources, *args, **kwargs))
```
collindutter, line 51: Can we reduce the cast type from `Any` to `dict[str, list[CsvRowArtifact]]`?

griptape/loaders/dataframe_loader.py
```diff
@@ -43,7 +33,10 @@ def _load_file(self, dataframe: DataFrame) -> list[CsvRowArtifact]:
 
         return artifacts
 
-    def _dataframe_to_hash(self, dataframe: DataFrame) -> str:
+    def load_collection(self, sources: Sequence[DataFrame], *args, **kwargs) -> dict[str, list[CsvRowArtifact]]:
+        return cast(Any, super().load_collection(sources, *args, **kwargs))
```
collindutter, line 37: Cast type

griptape/loaders/email_loader.py
```diff
@@ -88,3 +78,8 @@ def _retrieve_email(self, query: EmailQuery) -> ListArtifact | ErrorArtifact:
 
     def _count_messages(self, message_numbers: bytes):
         return len(list(filter(None, message_numbers.decode().split(" "))))
+
+    def load_collection(
+        self, sources: Sequence[EmailQuery], *args, **kwargs
+    ) -> dict[str, ListArtifact | ErrorArtifact]:
+        return cast(Any, super().load_collection(sources, *args, **kwargs))
```
collindutter, line 85: Cast type

griptape/loaders/file_loader.py
```diff
@@ -16,28 +16,20 @@ class FileLoader(BaseLoader):
...
+                    return BlobArtifact(file.read(), name=file_name, dir_name=os.path.dirname(source))
         except FileNotFoundError:
             return ErrorArtifact(f"file {file_name} not found")
         except Exception as e:
             return ErrorArtifact(f"error loading file: {e}")
+
+    def load_collection(
+        self, sources: Sequence[str | Path], *args, **kwargs
+    ) -> dict[str, TextArtifact | BlobArtifact | ErrorArtifact]:
+        return cast(Any, super().load_collection(sources, *args, **kwargs))
```
collindutter, line 35: Cast type

griptape/loaders/image_loader.py
```diff
@@ -67,3 +74,6 @@ def _get_mime_type(self, image_format: str | None) -> str:
             raise ValueError(f"Unsupported image format {image_format}")
 
         return self.FORMAT_TO_MIME_TYPE[image_format.lower()]
+
+    def load_collection(self, sources: Sequence[bytes | IO | Path], *args, **kwargs) -> dict[str, ImageArtifact]:
+        return cast(Any, super().load_collection(sources, *args, **kwargs))
```
collindutter, line 79: Cast type

griptape/loaders/csv_loader.py
```diff
@@ -37,3 +32,20 @@ def _load_file(self, filename: str) -> list[CsvRowArtifact]:
...
+
+    def _stream_from_source(self, source: bytes | str | IO | Path) -> IO:
+        if isinstance(source, bytes):
+            return StringIO(source.decode())
+        elif isinstance(source, str):
+            return StringIO(source)
+        elif isinstance(source, IOBase):
+            return cast(IO, source)
+        elif isinstance(source, Path):
+            return open(source, encoding="utf-8")
```
collindutter, line 44: Should this use `self.encoding`?

griptape/loaders/sql_loader.py
```diff
@@ -38,3 +30,6 @@ def _load_query(self, query: str) -> list[CsvRowArtifact]:
             artifacts.append(chunk)
 
         return artifacts
+
+    def load_collection(self, sources: Sequence[str], *args, **kwargs) -> dict[str, list[CsvRowArtifact]]:
+        return cast(Any, super().load_collection(sources, *args, **kwargs))
```
collindutter, line 35: Cast type

tests/unit/loaders/conftest.py
collindutter, line 1: How does this file work?

griptape/loaders/base_text_loader.py
dylanholmes: Sure! (I had chosen `Any` as it is shorter without any functional difference, but I'm fine either way)

tests/unit/loaders/conftest.py
collindutter: Sweet!

griptape/loaders/csv_loader.py
dylanholmes: Now I remember why I didn't do this. `CsvLoader` extends `BaseLoader` rather than `BaseTextLoader`, so it doesn't have the `encoding` field.

I feel like it should for usability, so I'll add one and pass it through.

dylanholmes:
We've decided to remove the responsibility of reading from the filesystem. I will put out a new PR for this approach.

PR #737, opened 2024-04-08
Remove file reading responsibility from loaders

dylanholmes:
We will want to create a `BaseFileManagerDriver` to enable the `FileManager` Tool to interact with alternative (e.g. non-local) filesystems. We'd like to be able to reuse the existing loaders to load in artifacts from files, though some of the loaders used by the `FileManager` currently do not support loading files from memory, and thus would subclasses of `BaseFileManagerDriver` to write to a file before invoking the loader.

To make this easier, this change will remove the responsibility of reading from the filesystem from all loaders. Since this is the only reason the `FileLoader` loader exists, we will remove that.

Changes:
- Remove `IO` and `Path` type options as source. All loaders that are related to specific file formats will accept `bytes` as source (and `str` for text file formats).
- Remove `FileLoader`
- Edit `FileManager` to have no default loader and to return a `BlobArtifact` by default if there is no default loader. This is essentially the same behavior. (The default used to be `FileLoader` with no args, which would always return a `BlobArtifact` and if you changed the default loader to be `FileLoader` with `encoding` set, then it would be equivalent to passing a `TextLoader` for the default)
- Add `to_key` method on `BaseLoader` with a reasonable implementation and move `load_collection` implementation in to the `BaseLoader`. The idea here is to encapsulate the part of `load_collection` that varies. This allows there to be a single implementation that just submits `load` to the executor for each source.
- Remove duplicate loads in `load_collection` for sources that are listed more than once in the input (or at least that have the same key).
- Edit CHANGELOG


griptape/loaders/csv_loader.py
```diff
@@ -13,27 +15,34 @@
...
+        for chunk in chunks:
+            artifacts.append(chunk)
 
         return artifacts
+
+    def load_collection(
+        self, sources: Sequence[bytes | str], *args, **kwargs
+    ) -> dict[str, ErrorArtifact | list[CsvRowArtifact]]:
+        return cast(
+            dict[str, Union[ErrorArtifact, list[CsvRowArtifact]]], super().load_collection(sources, *args, **kwargs)
```
collindutter, line 47: Can we use pipe `|` syntax instead of `Union` here?

griptape/loaders/email_loader.py
```diff
@@ -88,3 +78,8 @@ def _retrieve_email(self, query: EmailQuery) -> ListArtifact | ErrorArtifact:
 
     def _count_messages(self, message_numbers: bytes):
         return len(list(filter(None, message_numbers.decode().split(" "))))
+
+    def load_collection(
+        self, sources: Sequence[EmailQuery], *args, **kwargs
+    ) -> dict[str, ListArtifact | ErrorArtifact]:
+        return cast(dict[str, Union[ListArtifact, ErrorArtifact]], super().load_collection(sources, *args, **kwargs))
```
collindutter, line 85: Pipe syntax

griptape/loaders/text_loader.py
```diff
@@ -33,13 +33,20 @@ class TextLoader(BaseTextLoader):
...
-        return utils.execute_futures_dict(
-            {
-                utils.str_to_hash(str(source)): self.futures_executor.submit(self._text_to_artifacts, source)
-                for source in sources
-            }
+    def load_collection(
+        self, sources: Sequence[bytes | str], *args, **kwargs
+    ) -> dict[str, ErrorArtifact | list[TextArtifact]]:
+        return cast(
+            dict[str, Union[ErrorArtifact, list[TextArtifact]]], super().load_collection(sources, *args, **kwargs)
```
collindutter, line 51: Pipe syntax

griptape/loaders/csv_loader.py
```diff
@@ -13,27 +15,34 @@
...
-        with open(filename, encoding="utf-8") as csv_file:
-            reader = csv.DictReader(csv_file, delimiter=self.delimiter)
-            chunks = [CsvRowArtifact(row) for row in reader]
+        if isinstance(source, bytes):
+            try:
+                source = source.decode(encoding=self.encoding)
+            except UnicodeDecodeError:
+                return ErrorArtifact(f"Failed to decode bytes to string using encoding: {self.encoding}")
+        elif not isinstance(source, str):
+            return ErrorArtifact(f"Expected source to be bytes or string, got {type(source)}")
```
collindutter, line 30: I think we can remove this type check, let's leave it up to the type checker.

griptape/loaders/text_loader.py
```diff
@@ -33,13 +33,20 @@ class TextLoader(BaseTextLoader):
...
 
-    def load(self, source: str | Path, *args, **kwargs) -> list[TextArtifact]:
+    def load(self, source: bytes | str, *args, **kwargs) -> ErrorArtifact | list[TextArtifact]:
+        if isinstance(source, bytes):
+            try:
+                source = source.decode(encoding=self.encoding)
+            except UnicodeDecodeError:
+                return ErrorArtifact(f"Failed to decode bytes to string using encoding: {self.encoding}")
+        elif not isinstance(source, str):
+            return ErrorArtifact(f"Expected source to be bytes or string, got {type(source)}")
```
collindutter, line 43: Same here, I think we can remove.

griptape/tools/file_manager/tool.py
```diff
@@ -24,7 +25,7 @@ class FileManager(BaseTool):
     """
 
     workdir: str = field(default=Factory(lambda: os.getcwd()), kw_only=True)
-    default_loader: BaseLoader = field(default=Factory(lambda: FileLoader()))
+    default_loader: BaseLoader = field(default=None, kw_only=True)
```
collindutter, line 28: `BaseLoader` should be `Optional[BaseLoader]`

griptape/utils/futures.py
```diff
@@ -1,5 +1,6 @@
 from concurrent import futures
-from typing import TypeVar
+from typing import Callable, TypeVar
+from collections.abc import Sequence
```
collindutter, line 3: Leftover change?

griptape/loaders/base_loader.py
```diff
@@ -18,8 +20,23 @@ class BaseLoader(ABC):
...
+        # to avoid duplicate work.
+        sources_by_key = {self.to_key(source): source for source in sources}
+        return execute_futures_dict(
+            {
+                key: self.futures_executor.submit(self.load, source, *args, **kwargs)
+                for key, source in sources_by_key.items()
+            }
+        )
+
+    def to_key(self, source: Any, *args, **kwargs) -> str:
```
collindutter, line 36: Should we rename to `to_hash`?

collindutter (changes requested):
Nice work!

griptape/loaders/csv_loader.py
dylanholmes: pyright didn't like it and I couldn't quickly figure out the reason why it complains about using pipe in the casts. I don't think its worth the time to investigate.

griptape/loaders/csv_loader.py
dylanholmes: I only made this change thanks to the type checker. For some reason that I don't fully understand, when I accept `bytes` as a parameter, it thinks that source can be `bytes | bytearray | memoryview`. If you don't like the `not isinstance(source, str)` I could change this to be:

```py
if isinstance(source, bytes):
    try:
        source = source.decode(encoding=self.encoding)
    except UnicodeDecodeError:
        return ErrorArtifact(f"Failed to decode bytes to string using encoding: {self.encoding}")
elif isinstance(source, bytearray | memoryview):
    return ErrorArtifact(f"Expected source to be bytes or string, got {type(source)}")
```

griptape/loaders/text_loader.py
dylanholmes: Replied above.

griptape/loaders/email_loader.py
dylanholmes: Replied above.

griptape/loaders/text_loader.py
dylanholmes: Replied above.

griptape/loaders/base_loader.py
dylanholmes: We could, but it doesn't necessarily need to be a hash. It just need to be the same value as the key in the dict returned by `load_collection`. Let me know if you feel strongly and I'll change it

griptape/loaders/csv_loader.py
dylanholmes: I just went ahead with my alternative suggestion above. (Note that I also needed to use `Union` instead of the pipe syntax to appease pyright's concerns about python version 3.10)

griptape/tools/file_manager/tool.py
dylanholmes: Done

griptape/utils/futures.py
dylanholmes: Removed

dylanholmes:
Fun, the latest unit tests fail in github actions, but not locally.

griptape/loaders/csv_loader.py
dylanholmes: Ok, I should have been using `Union` either because it makes no sense to use generics in an expression evaluated at runtime (duh), though it did work locally which is kind of weird. Anyway, it should be fixed now.

griptape/loaders/csv_loader.py
```diff
@@ -13,27 +15,34 @@
...
-        with open(filename, encoding="utf-8") as csv_file:
-            reader = csv.DictReader(csv_file, delimiter=self.delimiter)
-            chunks = [CsvRowArtifact(row) for row in reader]
+        if isinstance(source, bytes):
+            try:
+                source = source.decode(encoding=self.encoding)
+            except UnicodeDecodeError:
+                return ErrorArtifact(f"Failed to decode bytes to string using encoding: {self.encoding}")
+        elif isinstance(source, bytearray) or isinstance(source, memoryview):
+            return ErrorArtifact(f"Unsupported source type: {type(source)}")
```
collindutter, line 29: Wouldn't a type checker also prevent this case since we've got the type as `bytes | str`?

griptape/loaders/text_loader.py
```diff
@@ -33,13 +33,20 @@ class TextLoader(BaseTextLoader):
...
 
-    def load(self, source: str | Path, *args, **kwargs) -> list[TextArtifact]:
+    def load(self, source: bytes | str, *args, **kwargs) -> ErrorArtifact | list[TextArtifact]:
+        if isinstance(source, bytes):
+            try:
+                source = source.decode(encoding=self.encoding)
+            except UnicodeDecodeError:
+                return ErrorArtifact(f"Failed to decode bytes to string using encoding: {self.encoding}")
+        elif isinstance(source, bytearray) or isinstance(source, memoryview):
+            return ErrorArtifact(f"Unsupported source type: {type(source)}")
```
collindutter, line 43: Same comment as other file.

griptape/tools/file_manager/tool.py
```diff
@@ -93,7 +94,13 @@ def load_files_from_disk(self, params: dict) -> ListArtifact | ErrorArtifact:
             full_path = Path(os.path.join(self.workdir, path))
             extension = path.split(".")[-1]
             loader = self.loaders.get(extension) or self.default_loader
-            result = loader.load(full_path)
+            with open(full_path, "rb") as file:
+                content = file.read()
+
+            if loader:
```
collindutter, line 100: Flip conditional and use explicit `is None` check

collindutter (changes requested):

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Improved RAG performance in `VectorQueryEngine`.
 - **BREAKING**: Secret fields (ex: api_key) removed from serialized Drivers.
+- **BREAKING**: Remove `FileLoader`.
+- **BREAKING**: `CsvLoader` no longer accepts `str` file paths as a source. It will now accept the content of the CSV file as a `str` or `bytes` object.
+- **BREAKING**: `PdfLoader` no longer accepts `str` file content, `Path` file paths or `IO` objects as sources. Instead, it will only accept the content of the PDF file as a `bytes` object.
+- **BREAKING**: `TextLoader` no longer accepts `Path` file paths as a source. It will now accept the content of the text file as a `str` or `bytes` object.
```
vasinov, line 18: Another breaking change is that `FileManager.default_loader` is now `None`.

griptape/loaders/base_loader.py
```diff
@@ -18,8 +20,23 @@ class BaseLoader(ABC):
     def load(self, source: Any, *args, **kwargs) -> BaseArtifact | Sequence[BaseArtifact]:
         ...
 
-    @abstractmethod
     def load_collection(
-        self, sources: list[Any], *args, **kwargs
+        self, sources: Sequence[Any], *args, **kwargs
```
vasinov, line 24: We don't use sequences anywhere in the framework. Should we not introduce it here? Also, if we do introduce it, should it be `seq` instead?

vasinov (changes requested):
Looks great overall!

griptape/loaders/base_loader.py
collindutter: @vasinov it's already used as a return type for `BaseLoader.load`. This came as part of the pyright fixes we made recently. Curious if it's required in this particular scenario though.

griptape/loaders/base_loader.py
collindutter: Nah its fine

griptape/loaders/csv_loader.py
dylanholmes: The type checker is the one who decided this was necessary! I'm digging through the docs trying to find out why.

griptape/loaders/base_loader.py
dylanholmes: I see that list is used often, but `Sequence` is used in `ListArtifact`, `BaseLoader.load` (as @collindutter mentioned), and also (probably by mistake) in `PdfLoader.load_collection`. It seemed inconsistent, and if I had to pick one myself, I would choose `Sequence` because accepting a more general type as input gives the most freedom to the client.

(Given that `load_collection` returns a dict would never need to specifically accept a `list`, technically we don't even need an ordered collection, it just needs to be Iterable, but I think anything more general than `Sequence` is getting a little crazy.)

I'd be fine using `list` instead, but if we do go that route, then was a specific reason for `Sequence` being used in `ListArtifact`?

griptape/loaders/csv_loader.py
dylanholmes: Here is the source of the problem:

> ...the `typing` module docs state that type checkers should treat `bytearray` and `memoryview` as implicit subtypes of `bytes`.

From https://github.com/python/mypy/issues/12643#issuecomment-1105580582

(Side note, I noticed from that thread that I can apply slightly more concise syntax in the form of a tuple when passin the type argument - `isinstance(source, (bytearray, memoryview))`)

griptape/loaders/base_loader.py
collindutter: The introduction of Sequence was to resolve pyright covariance issues. If we can get away with using `list` I'd say do that. But if pyright complains, `Sequence` is probably the easiest solution.

CHANGELOG.md
dylanholmes: Added

griptape/loaders/base_loader.py
dylanholmes: Reverted `Sequence` back to `list`

collindutter (approved):

vasinov (approved):

PR #741, opened 2024-04-10
Remove boto3 from types

collindutter:
Fixes serialization crash when `boto3` is not installed via extras. ~Side effect of this is that we cannot use `boto3` type hints (boto.Session) in any serializable fields.~

### DIFF
--- a/griptape/schemas/base_schema.py
+++ b/griptape/schemas/base_schema.py
@@ -100,13 +100,15 @@ class BaseSchema(Schema):
         Args:
             attrs_cls: An attrs class.
         """
+        from griptape.utils.import_utils import import_optional_dependency, is_dependency_installed
 
         # These modules are required to avoid `NameError`s when resolving types.
         from griptape.drivers import BaseConversationMemoryDriver, BasePromptDriver, BasePromptModelDriver
         from griptape.structures import Structure
         from griptape.utils import PromptStack
         from griptape.tokenizers.base_tokenizer import BaseTokenizer
-        import boto3
+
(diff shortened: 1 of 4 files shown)


vachillo (dismissed):

griptape/schemas/base_schema.py
```diff
@@ -100,13 +100,15 @@ def _resolve_types(cls, attrs_cls: type) -> None:
...
+        from griptape.utils.import_utils import import_optional_dependency, is_dependency_installed
 
         # These modules are required to avoid `NameError`s when resolving types.
         from griptape.drivers import BaseConversationMemoryDriver, BasePromptDriver, BasePromptModelDriver
         from griptape.structures import Structure
         from griptape.utils import PromptStack
         from griptape.tokenizers.base_tokenizer import BaseTokenizer
-        import boto3
+
+        boto3 = import_optional_dependency("boto3") if is_dependency_installed("boto3") else {}
```
dylanholmes, line 111: If this is ever needed in more than one place, then might be nice to have something like `try_import_optional_dependency()`.

Btw, what happens if you default to `None` instead of `{}`?

dylanholmes (dismissed):

griptape/schemas/base_schema.py
collindutter: Honestly not totally sure, I think it was working fine with both. But I updated to a more appropriate value `Any` since that's an actual python type.

andrewfrench (approved):

PR #743, opened 2024-04-11
Add AmazonS3FileManagerDriver

dylanholmes:
Changes:
- Add AmazonS3FileManagerDriver
- Edit LocalFileManagerDriver to have consistent error artifact output
- Edit BaseFileManagerDriver with more exception to error artifact translations for common scenarios
- Improve test coverage of LocalFileManagerDriver

Also manually tested using the driver against a real s3 bucket, in a similar style to those added in the unit tests.

### DIFF
--- a/griptape/drivers/file_manager/amazon_s3_file_manager_driver.py
+++ b/griptape/drivers/file_manager/amazon_s3_file_manager_driver.py
@@ -43,6 +43,30 @@ class AmazonS3FileManagerDriver(BaseFileManagerDriver):
                 raise FileNotFoundError
         return files_and_dirs
 
+    def try_load_file(self, path: str) -> bytes:
+        botocore = import_optional_dependency("botocore")
+        full_key = self._to_full_key(path)
+
+        if self._is_a_directory(full_key):
+            raise IsADirectoryError
+
+        try:
+            response = self.s3_client.get_object(Bucket=self.bucket, Key=full_key)
+            return response["Body"].read()
+        except botocore.exceptions.ClientError as e:
+            if e.response["Error"]["Code"] in {"NoSuchKey", "404"}:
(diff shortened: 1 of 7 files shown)


dylanholmes:
The unit tests are currently failing due to an unrelated unit test -`TestAwsIotCoreEventListenerDriver.test_try_publish_event`. That particular unit test works locally, but it seems to be the only test that is using moto3's decorators and I'm wondering if they don't work in the github environment.

I would have expected this to have been caught in the original PR, not sure why I'm just seeing this behavior now.

griptape/drivers/file_manager/amazon_s3_file_manager_driver.py
```diff
@@ -0,0 +1,123 @@
...
+            for obj in page.get("CommonPrefixes", []):
+                prefix = obj.get("Prefix")
+                dir = prefix[len(full_key) :].rstrip("/")
+                files_and_dirs.append(dir)
+
+            for obj in page.get("Contents", []):
+                key = obj.get("Key")
+                file = key[len(full_key) :]
+                files_and_dirs.append(file)
+        return files_and_dirs
```
collindutter, line 85: Nit: private/protected methods should go after public ones.

collindutter (commented):
Great work!!

griptape/drivers/file_manager/amazon_s3_file_manager_driver.py
dylanholmes: Thanks for letting me know the convention 👍

dylanholmes: Fixed

collindutter (approved):

PR #757, opened 2024-04-25
Default `off_prompt=False` on `BaseTool`

collindutter:


docs/examples/multiple-agent-shared-memory.md
```diff
@@ -56,7 +56,7 @@ mongo_driver = AzureMongoDbVectorStoreDriver(
 
 loader = Agent(
     tools=[
-        WebScraper()
+        WebScraper(off_prompt=True),
```
vasinov, line 59: Should be `False`.

docs/griptape-framework/structures/tasks.md
```diff
@@ -103,7 +103,7 @@ agent = Agent()
 agent.add_task(
     ToolkitTask(
         "Load https://www.griptape.ai, summarize it, and store it in a file called griptape.txt", 
-        tools=[WebScraper(), FileManager(), TaskMemoryClient(off_prompt=False)]
+        tools=[WebScraper(off_prompt=True), FileManager(off_prompt=True), TaskMemoryClient(off_prompt=True)]
```
vasinov, line 106: Should be `False` for `TaskMemoryClient`.

vasinov (changes requested):
Bigger change than we thought, huh? :)

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
 * **Security requirements**: many organizations don't want data to leave their cloud for regulatory and security reasons.
-* **Long textual content**: when textual content returned by tools can't fit in the token limit, it's often useful to perform operations on it in a separate process, not in the main LLM.
-* **Non-textual content**: tools can generate images, videos, PDFs, and other non-textual content that can be stored in memory and acted upon later by other tools.
+* **Long textual content**: when textual content returned by Tools can't fit in the token limit, it's often useful to perform operations on it in a separate process, not in the main LLM.
+* **Non-textual content**: Tools can generate images, videos, PDFs, and other non-textual content that can be stored in Task Memory and acted upon later by other Tools.
+
+
+## Off Prompt
+
+You can enable or disable Task Memory on a Tool with the `off_prompt` parameter. When `off_prompt` is set to `True`, the Tool will store its output in Task Memory. When `off_prompt` is set to `False`, the Tool will return its output directly to the LLM.
```
shhlife, line 13: If it's mandatory to determine whether or not TaskMemory should be true or false, should we say "You must enable or disable..."?

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
 ## Overview
-Task Memory augments activity inputs and outputs with storage capabilities. It's a way to keep any data generated by tools off prompt but still allow LLMs to operate on that data remotely. This is useful in the following scenarios:
+
+Task Memory is a powerful feature of Griptape that allows you to control where the data returned by Tools is stored. This is useful in the following scenarios:
+
 
 * **Security requirements**: many organizations don't want data to leave their cloud for regulatory and security reasons.
-* **Long textual content**: when textual content returned by tools can't fit in the token limit, it's often useful to perform operations on it in a separate process, not in the main LLM.
-* **Non-textual content**: tools can generate images, videos, PDFs, and other non-textual content that can be stored in memory and acted upon later by other tools.
+* **Long textual content**: when textual content returned by Tools can't fit in the token limit, it's often useful to perform operations on it in a separate process, not in the main LLM.
```
vasinov, line 7: Should probably not say "process" since it's still the same system process. Can't think of a better word though right now.

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
+                             "6be74c5128024c0588eb9bee1fdb9aa5"
+[04/26/24 13:07:16] ERROR    Subtask ecbb788d9830491ab72a8a2bbef5fb0a
+                             Invalid action JSON: Or({Literal("name", description=""): 'Calculator', Literal("path", description="Can be used for computing simple
+                             numerical or algebraic calculations in Python"): 'calculate', Literal("input", description=""): {'values': Schema({Literal("expression",
+                             description="Arithmetic expression parsable in pure Python. Single line only. Don't use variables. Don't use any imports or external
+                             libraries"): <class 'str'>})}, Literal("tag", description="Unique tag name for action execution."): <class 'str'>}) did not validate {'name':
+                             'Memory', 'path': 'get', 'input': {'memory_name': 'TaskMemory', 'artifact_namespace': '6be74c5128024c0588eb9bee1fdb9aa5'}, 'tag':
+                             'get_sqrt_result'}
+                             Key 'name' error:
+                             'Calculator' does not match 'Memory'
```
vasinov, line 77: Should we truncate this as well and just say "errors are generated" or something?

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
+...Output truncated for brevity...
+```
+
+When we set `off_prompt` to `True`, the Agent does not function as expected, even generating an error. This is because because the Calculator output is being stored in Task Memory but the Agent has no way to access it. To fix this, we need the `TaskMemoryClient`.
```
vasinov, line 81: Double "because."

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
+                    INFO     Subtask 2618fc23194947018ce7d3facb33e5b9
+                             Response: Successfully saved memory artifacts to disk
+[04/26/24 14:46:04] INFO     ToolkitTask 1ed8480fa2dc42d7ac56165491869efc
+                             Output: The summary of the webpage has been successfully saved to the file "san_francisco_summary.txt" in the "summaries" directory.
+
 ```
+
+Note that in this example GPT-4 _never_ saw the contents of the page, only that it was stored in Task Memory.
+
+!!! info
```
shhlife, line 307: This didn't display properly in the preview

vasinov (changes requested):

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
 ## Overview
-Task Memory augments activity inputs and outputs with storage capabilities. It's a way to keep any data generated by tools off prompt but still allow LLMs to operate on that data remotely. This is useful in the following scenarios:
+
+Task Memory is a powerful feature of Griptape that allows you to control where the data returned by Tools is stored. This is useful in the following scenarios:
```
emjay07, line 3: How does "the data returned by Tools" relate to the data returned by the LLM? It may be useful to add context in relation to the model as that is what most new users will be familiar with.

docs/griptape-framework/structures/task-memory.md
emjay07: suggestions: operation, thread, or action?

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
-* **Non-textual content**: tools can generate images, videos, PDFs, and other non-textual content that can be stored in memory and acted upon later by other tools.
+* **Long textual content**: when textual content returned by Tools can't fit in the token limit, it's often useful to perform operations on it in a separate process, not in the main LLM.
+* **Non-textual content**: Tools can generate images, videos, PDFs, and other non-textual content that can be stored in Task Memory and acted upon later by other Tools.
+
+
+## Off Prompt
+
+You can enable or disable Task Memory on a Tool with the `off_prompt` parameter. When `off_prompt` is set to `True`, the Tool will store its output in Task Memory. When `off_prompt` is set to `False`, the Tool will return its output directly to the LLM.
+
+By default, all Tools have `off_prompt` as a required parameter, forcing you to consider whether you'd like to use Task Memory or not. Lets look at a simple example where `off_prompt` is set to `False`:
```
emjay07, line 15: nit: I would put the info about the default up in the previous paragraph as it puts all the base information a customer needs up front.

> You can enable or disable Task Memory on a Tool with the `off_prompt` parameter. By default, all Tools have `off_prompt` as a required parameter, forcing you to consider whether you'd like to use Task Memory or not. When `off_prompt` is set to `True`...

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
+[04/26/24 13:06:48] INFO     Subtask a88c0feeaef6493796a9148ed68c9caf
+                             Thought: To find the square root of 12345, I can use the Calculator action with the expression "12345 ** 0.5".
+                             Actions: [{"name": "Calculator", "path": "calculate", "input": {"values": {"expression": "12345 ** 0.5"}}, "tag": "sqrt_12345"}]
+                    INFO     Subtask a88c0feeaef6493796a9148ed68c9caf
+                             Response: 111.1080555135405
+[04/26/24 13:06:49] INFO     ToolkitTask 36b9dea13b9d479fb752014f41dca54c
+                             Output: The square root of 12345 is approximately 111.108.
+```
+
+Since the result of the Calculator Tool is neither sensitive nor too large, we can set `off_prompt` to `False` and not use Task Memory.
```
emjay07, line 41: I love these examples. Much more clear on how it impacts the end user.

docs/griptape-framework/structures/task-memory.md
emjay07: We should probably leave the actual error message the customer would see in the example, but fine with truncating any of the middle.

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
+                             the query path, providing the memory_name and artifact_namespace as input.
+                             Actions: [{"tag": "retrieve_sqrt", "name": "TaskMemoryClient", "path": "query", "input": {"values": {"memory_name": "TaskMemory",
+                             "artifact_namespace": "7554b69e1d414a469b8882e2266dcea1", "query": "What is the result of the calculation?"}}}]
+[04/26/24 13:13:16] INFO     Subtask 32b9163a15644212be60b8fba07bd23b
+                             Response: The result of the calculation is 111.1080555135405.
+[04/26/24 13:13:17] INFO     ToolkitTask 5b46f9ef677c4b31906b48aba3f45e2c
+                             Output: The square root of 12345 is approximately 111.108.
+```
+
+Note that on the `TaskMemoryClient` we've set `off_prompt` to `False` so that the results of the query can be returned directly to the LLM. If we had kept it as `True`, the results would have been stored back Task Memory which would've put us back to square one.
```
emjay07, line 122: nit: I'd put this comment up above the example as it was my first question when reading the example and was then confused until I got here.

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
+This is because the content of the webpage is too large to fit in the LLM's input token limit. We can fix this by storing the content in Task Memory, and then querying it with the `TaskMemoryClient`.
+
+```python
+from griptape.structures import Agent
+from griptape.tools import WebScraper, TaskMemoryClient
+
+agent = Agent(
+    tools=[
+        WebScraper(off_prompt=True),
+        TaskMemoryClient(off_prompt=False),
```
emjay07, line 158: We may have covered this already in a separate conversation, but what is the value of `TaskMemoryClient` having `off_prompt` at all? would it every be true?

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
+```
+
+## Sensitive Data
 
-prompt_driver = OpenAiChatPromptDriver(model="gpt-3.5-turbo")
+Because Task Memory splits up the storage and retrieval of data, you can use different models for each step.
+
+Here is an example where we use GPT-4 to orchestrate the Tools and store the data in Task Memory, and a self-hosted Llama-2 model to summarize the raw content.
+
+```python
```
emjay07, line 195: This example may be good to copy into our Recipes section of our docs as well. I could see this being useful for people at a top-level example of the power of Griptape vs. needing to find it buried in our docs.

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
+                             Response: Successfully saved memory artifacts to disk
+[04/26/24 14:46:04] INFO     ToolkitTask 1ed8480fa2dc42d7ac56165491869efc
+                             Output: The summary of the webpage has been successfully saved to the file "san_francisco_summary.txt" in the "summaries" directory.
+
 ```
+
+Note that in this example GPT-4 _never_ saw the contents of the page, only that it was stored in Task Memory.
+
+!!! info
+    In this instance of using `FileManager` to save a file to disk, it does not matter what the value of `off_prompt` is set to. This Tool Activity returns an `InfoArtifact` which Griptape will not store in Task Memory. Griptape will only store `TextArtifact`'s, `BlobArtifact`'s, and `ListArtifact`'s in Task Memory.
```
emjay07, line 308: This may be good to call out with a section header like `## When using FileManager` or `## Notes on InfoArtifacts`.

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,320 @@
...
+
+## Tools That Use Task Memory
+
+As seen in the previous example, certain Tools are designed to read directly from Task Memory. This means that you can use these Tools to interact with the data stored in Task Memory without needing to pass it through the LLM.
```
emjay07, line 312: Does this mean a user does not need to specify `off_prompt=true` when using these tools?

docs/examples/multiple-agent-shared-memory.md
collindutter: That's the nifty part about this example, it's a separate Agent using the same shared Task/Meta Memory instances. So this is correct.

docs/griptape-framework/structures/tasks.md
collindutter: `FileManager` will read directly from Task Memory

docs/griptape-framework/structures/task-memory.md
collindutter: I think it's important to show a full error message, I already truncated quite a bit of other output.

docs/griptape-framework/structures/task-memory.md
collindutter: Yes, the Sensitive Data example has it set to `true`. Will add more details.

emjay07 (commented):
The `TaskMemory` docs look great and have some awesome examples. 

I am questioning the over-arching change of removing the `off_prompt` default for the rest of this PR. I feel like we should default to *something* as removing the default makes the barrier to getting started w/ Griptape that much more complicated for customers.

README.md
```diff
@@ -94,9 +94,9 @@ from griptape.tools import WebScraper, FileManager, TaskMemoryClient
 agent = Agent(
     input_template="Load {{ args[0] }}, summarize it, and store it in a file called {{ args[1] }}.",
     tools=[
-        WebScraper(),
-        FileManager(),
-        TaskMemoryClient(off_prompt=True)
+        WebScraper(off_prompt=True),
+        TaskMemoryClient(off_prompt=True),
+        FileManager()
```
emjay07, line 99: FileManager need `off_prompt=False` here to be 1:1 with the original.

docs/griptape-framework/drivers/embedding-drivers.md
```diff
@@ -177,7 +177,7 @@ from griptape.drivers import (
 from griptape.config import StructureConfig
 
 agent = Agent(
-    tools=[WebScraper(), TaskMemoryClient(off_prompt=False)],
+    tools=[WebScraper(off_prompt=True), TaskMemoryClient(off_prompt=False)],
```
emjay07, line 180: for these examples, do we still need `off_prompt=False` for the `TaskMemoryClient` if the default is updated?

docs/griptape-framework/structures/task-memory.md
```diff
@@ -1,89 +1,345 @@
...
+                             Response: Successfully saved memory artifacts to disk
+[04/30/24 16:37:27] INFO     ToolkitTask 3d3c5f5a98a44f32ad9533621c036604
+                             Output: The number of copies of Elden Ring sold has been successfully saved to the file "elden_ring_sales.txt" in the "sales_data" directory.
 ```
+
+## Tools That Can Read From Task Memory
+
+As seen in the previous example, certain Tools are designed to read directly from Task Memory. This means that you can use these Tools to interact with the data stored in Task Memory without needing to pass it through the LLM.
+
+Unless you're handing off data to another Tool capable of reading from Task Memory, you should most likely set `off_prompt` to `False` on these Tools otherwise you may end up [Task Memory Looping](#task-memory-looping).
```
emjay07, line 279: nit: reword slightly to lean into the default being False now.

> Unless you're handing off data to another Tool capable of reading from Task Memory, you should most likely keep `off_prompt` set to `False` on these Tools

docs/griptape-framework/structures/tasks.md
emjay07, line 106: Should `TaskMemoryClient` have `off_prompt=False`? Seems like this would get stuck in a loop.

docs/griptape-tools/official-tools/aws-s3-client.md
```diff
@@ -62,4 +63,4 @@ agent.run("List all my S3 buckets.")
                              3. Bucket Name: 'example-bucket-3', Creation Date: 
                              '2022-01-03T00:00:00Z'                             
                              Please note that the creation dates are in UTC.  
-```
\ No newline at end of file
+```
```
emjay07, line 66: nit: what is going on with this change? some weird whitespace?

collindutter: I think some weird EOL stuff. Just ignore

docs/griptape-framework/structures/tasks.md
collindutter: `FileManager` can read from Task Memory. So it goes `WebScraper` -> Task Memory `TaskMemoryClient` -> Task Memory -> FileManager

docs/griptape-framework/structures/task-memory.md
collindutter: Removed, we already have a section that calls out Task Memory Looping. And this behavior is now harder to accidentally stumble into.

docs/griptape-framework/drivers/embedding-drivers.md
collindutter: Not technically, but I figured it couldn't hurt to show examples of being explicit.

README.md
collindutter: Do you mean `off_prompt=True`? `FileManager()` is now identical to `FileManager(off_prompt=False)`

Though technically `FileManager` returns an `InfoArtifact` which is not stored in Task Memory, so it doesn't matter. I figured omitting the parameter was slightly better than including one that didn't do anything.

vachillo (approved):
awesome. it might be good to fast-fail if there is no `TaskMemoryClient` when anything has `off_prompt=True` set. but we can fast-follow on that

README.md
emjay07: Got it, I must've misread something in my original pass.

PR #764, opened 2024-04-29
Add structure run tool

zachgiordano:
* Add `GriptapeCloudStructureRunClient` tool for invoking Griptape Cloud Structure Run APIs
* Refactor the Griptape Cloud tools to use a `BaseGriptapeCloudClient`


griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,101 @@
...
+
+            status = content.get("status")
+
+            wait_time = 1
+            wait_attempts = 0
+            while status in ["QUEUED", "RUNNING"] and wait_attempts < self.structure_run_wait_time_attempts:
+                # wait
+                time.sleep(wait_time)
+                wait_time = wait_time * self.structure_run_wait_time_multiplier
+                wait_attempts += 1
```
collindutter, line 72: Lets use [tenacity](https://tenacity.readthedocs.io/en/latest/) like we do in [other places](https://github.com/griptape-ai/griptape/blob/dev/griptape/mixins/exponential_backoff_mixin.py#L17).

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,101 @@
...
+        }
+    )
+    def submit_structure_run(self, params: dict) -> TextArtifact | ErrorArtifact:
+        from requests import post, exceptions
+
+        args: list[str] = params["values"]["args"]
+        url = urljoin(self.base_url.strip("/"), f"/api/structures/{self.structure_id}/runs")
+
+        try:
+            response = post(url, json={"args": args}, headers=self.headers)
```
collindutter, line 42: Let's add `response.raise_for_status()` to all requests.

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+
+    description: Optional[str] = field(default=None, kw_only=True)
+    structure_id: str = field(kw_only=True)
+    structure_run_wait_time_attempts: int = field(default=20, kw_only=True)
+    structure_run_wait_time_multiplier: float = field(default=1.5, kw_only=True)
+
+    @activity(
+        config={
+            "description": "Can be used to submit a run to a structure with the following description: {{ _self._get_structure_description() }}",
```
collindutter, line 29: Capitalize `Run` and `Structure`, may have subtle implications for the LLM.

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,101 @@
...
+        try:
+            response = post(url, json={"args": args}, headers=self.headers)
+
+            return TextArtifact(response.json())
+        except exceptions.RequestException as err:
+            return ErrorArtifact(str(err))
+
+    @activity(
+        config={
+            "description": "Can be used to get the result of a structure run",
```
collindutter, line 50: Capitalize.

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,101 @@
...
+
+        try:
+            response = get(url, headers=self.headers)
+            content = response.json()
+
+            status = content.get("status")
+
+            wait_time = 1
+            wait_attempts = 0
+            while status in ["QUEUED", "RUNNING"] and wait_attempts < self.structure_run_wait_time_attempts:
```
collindutter, line 68: Nit: a tuple is probably a better data structure for `["QUEUED", "RUNNING"]` since it's a known size and immutable.

collindutter (changes requested):
Nice work and thanks for the refactors!

collindutter (commented):
I wonder if we should just bake in the whole submit/polling logic into a single method. Can't think of any benefits to having the LLM be the one to poll.

zachgiordano:
@collindutter @vachillo Thanks for the comments! Attempted to address everything. Lmk how it looks

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,87 @@
...
+        else:
+            from requests import get
+
+            url = urljoin(self.base_url.strip("/"), f"/api/structures/{self.structure_id}/")
+
+            response = get(url, headers=self.headers).json()
+            if "description" in response:
+                return response["description"]
+            else:
+                raise ValueError(f'Error getting Structure description: {response["message"]}')
```
collindutter, line 87: Sorry didn't catch this in the first pass, this is probably better suited as a python `@property`

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,87 @@
...
+        if status in ("QUEUED", "RUNNING"):
+            raise Exception("Structure Run is still in progress")
+
+        if status != "SUCCEEDED":
+            return ErrorArtifact(content)
+
+        if "output" in content:
+            return TextArtifact(content["output"])
+        else:
+            return ErrorArtifact("No output found in response")
```
collindutter, line 73: No `output` doesn't necessarily mean it errored. We should return an `InfoArtifact` instead.

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,87 @@
...
+            return ErrorArtifact(str(err))
+
+    def _get_structure_run_result(self, structure_run_id: str) -> TextArtifact | ErrorArtifact:
+        url = urljoin(self.base_url.strip("/"), f"/api/structure-runs/{structure_run_id}")
+
+        for attempt in self.retrying():
+            with attempt:
+                return self._get_structure_run_result_attempt(url)
+        else:
+            return ErrorArtifact("Failed to get Run result.")
```
collindutter, line 53: Let's specify that it failed after retrying for X attempts.

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,87 @@
...
+        from requests import get, Response
+
+        response: Response = get(structure_run_url, headers=self.headers)
+        response.raise_for_status()
+        content = response.json()
+
+        status = content.get("status")
+
+        if status in ("QUEUED", "RUNNING"):
+            raise Exception("Structure Run is still in progress")
```
collindutter, line 65: It feels wrong to raise an exception when nothing is actually going wrong. Is there any way to use tenacity with expected retries?

collindutter (changes requested):

collindutter (commented):
Also please create a docs page for this Tool.

griptape/tools/griptape_cloud_structure_run_client/tool.py
zachgiordano: Discussed offline. Since this use case is intended for polling rather than exponential backoff, opted to go with a simple polling wait interval rather than involve tenacity at this time.

zachgiordano:
@collindutter good call, added a docs page and refactored based on comments!

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,99 @@
...
+            if "description" in response:
+                self._description = response["description"]
+            else:
+                raise ValueError(f'Error getting Structure description: {response["message"]}')
+
+        return self._description
+
+    @description.setter
+    def description(self, value: str) -> None:
+        self._description = value
```
collindutter, line 99: Properties should come after class fields

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,99 @@
...
+            return ErrorArtifact(str(err))
+
+    def _get_structure_run_result(self, structure_run_id: str) -> InfoArtifact | TextArtifact | ErrorArtifact:
+        url = urljoin(self.base_url.strip("/"), f"/api/structure-runs/{structure_run_id}")
+
+        result = self._get_structure_run_result_attempt(url)
+        status = result.get("status")
+
+        wait_attempts = 0
+        while status in ["QUEUED", "RUNNING"] and wait_attempts < self.structure_run_max_wait_time_attempts:
```
collindutter, line 55: Nit: use tuple over list.

griptape/tools/griptape_cloud_structure_run_client/tool.py
```diff
@@ -0,0 +1,99 @@
...
+            return self._get_structure_run_result(response_json["structure_run_id"])
+
+        except (exceptions.RequestException, HTTPError) as err:
+            return ErrorArtifact(str(err))
+
+    def _get_structure_run_result(self, structure_run_id: str) -> InfoArtifact | TextArtifact | ErrorArtifact:
+        url = urljoin(self.base_url.strip("/"), f"/api/structure-runs/{structure_run_id}")
+
+        result = self._get_structure_run_result_attempt(url)
+        status = result.get("status")
```
collindutter, line 52: This has the potential to be `None`. We should use a regular dictionary lookup or handle the `None` case if we ever expect it to be so.

collindutter (changes requested):
Couple more minor things.

collindutter (approved):
Nice 😎

PR #765, opened 2024-04-30
Refactor how configs work

collindutter:
- Simplify use of `StructureConfig` classes by moving Driver config to top level.
- Make it easier to configure custom Task Memory without needing Config classes.
- Remove `task_memory` config from Config classes.

----
📚 Documentation preview 📚: https://griptape--765.org.readthedocs.build//765/


dylanholmes (dismissed):
Sweet simplicity!

I think this great change overall.

Approving, though it would be nice if you could elaborate on the decision to remove `task_memory` from the config classes.

collindutter:
> it would be nice if you could elaborate on the decision to remove task_memory from the config classes.

Originally they were added here to make configuring Task Memory easier. Though in reality it didn't really make it much easier and just over-complicated the Config classes. Instead, this PR makes it easier to configure Task Memory _without_ the need of a Config class. More details on how to do this will come in https://github.com/griptape-ai/griptape/pull/757

griptape/config/openai_structure_config.py
```diff
@@ -20,37 +11,24 @@
...
-                vector_store_driver=LocalVectorStoreDriver(
-                    embedding_driver=OpenAiEmbeddingDriver(model="text-embedding-3-small")
-                ),
-            )
-        ),
-        kw_only=True,
+class OpenAiStructureConfig(StructureConfig):
+    prompt_driver: OpenAiChatPromptDriver = field(
+        default=Factory(lambda: OpenAiChatPromptDriver(model="gpt-4o")), metadata={"serializable": True}
+    )
```
dylanholmes, line 17: Should this (and similar fields) have `kw_only=True`?

dylanholmes (changes requested):

griptape/config/openai_structure_config.py
collindutter: Yes good catch

vachillo (dismissed):

vachillo (approved):

PR #779, opened 2024-05-14
Support case when model returns empty Actions list

andrewfrench:
This PR addresses a reported issue when parsing a list of Actions while initializing an ActionsSubtask.

In some cases, Models include an empty list of Actions in their responses. We currently raise a SchemaError this happens:

```
[05/11/24 12:58:44] INFO     ToolkitTask 9f94e60e40d04516b9cce029eab11a25                                                                                     
                             Input: hi my name is jason                                                                                                       
[05/11/24 12:58:47] ERROR    Subtask 9f94e60e40d04516b9cce029eab11a25                                                                                         
                             Invalid action JSON: Array item count 0 is less than minimum count of 1.                                                         
                    INFO     Subtask 3352ba0de32041969bbb7d9781a8c236                                                                                         
                             Thought: The user has introduced himself as Jason. There is no specific question or request, so I will acknowledge the           
                             introduction and ask how I can assist him.                                                                                       
                             Actions: []                                                                                                                      
                             Answer: Hi Jason! How can I assist you today?                                                                                    
                    INFO     Subtask 3352ba0de32041969bbb7d9781a8c236                                                                                         
                             Response: Action JSON validation error: Array item count 0 is less than minimum count of 1.                                      
```

If the model does not return an Actions list at all, we do not attempt to parse actions and do not raise an exception. In both cases the model is electing to take no actions. If "no actions" is a valid response, allowing one case and disallowing another is inconsistent behavior.


vasinov (dismissed):

griptape/tasks/actions_subtask.py
```diff
@@ -198,69 +198,73 @@ def __init_from_prompt(self, value: str) -> None:
...
 
-        if len(actions_matches) > 0:
-            try:
-                data = actions_matches[-1]
-                actions_list: list = json.loads(data, strict=False)
+        self.__parse_actions(actions_matches)
 
+        # If there are no actions to take but an answer is provided, set the answer as the output.
+        if len(self.actions) == 0 and self.output is None and len(answer_matches) > 0:
+            self.output = TextArtifact(answer_matches[-1])
```
andrewfrench, line 205: We're currently testing which branch to take based on whether or not `answer_matches` has elements or not. If an empty list is provided, there's a match and we try to parse the actions. If, while parsing, we decide there actually aren't any actions, we continue execution **without any actions or output**. This is interpreted by `ToolkitTask` as a deviation from the ReAct prompt and the `Subtask`'s _raw_ input is set as its output.

This change updates the logic to first try parsing actions. If there are no `actions_matches`, or if an empty set is provided by the model, this results in an empty `self.actions` list. If the model chooses to take no actions, we set the output to the answer the same way we would if no Actions response was included by the model in the current release.

griptape/tasks/actions_subtask.py
```diff
@@ -198,69 +198,73 @@ def __init_from_prompt(self, value: str) -> None:
...
+
+    def __parse_actions(self, actions_matches: list[str]) -> None:
```
vasinov, line 207: Thanks for refactoring this out. Long overdue!

vasinov (approved):

PR #780, opened 2024-05-14
Add gpt-4o and update OpenAiStructureConfig default prompt_driver

zachgiordano:
- Add `gpt-4o` in `OpenAiTokenizer`
- Update `OpenAiStructureConfig` default prompt_driver to `gpt-4o`
- Update openai structure config test

https://platform.openai.com/docs/models/gpt-4o


vasinov (changes requested):
I think `OpenAiTokenizer.count_tokens` might need to be updated as well.

andrewfrench (approved):
🚀

PR #781, opened 2024-05-15
Audio transcription support

andrewfrench:
- Introduces an `AudioLoader`, responsible for producing an `AudioArtifact` from binary audio data
- Add `AudioTranscriptionTask`, accepting an `AudioArtifact` and responding with a `TextArtifact` containing the transcribed text
- Add `AudioTranscriptionClient` allowing an Agent to generate a transcription from audio in memory or on the filesystem
- Add `OpenAiAudioTranscriptionDriver` to support integration with OpenAI's speech-to-text models
- Update structure configs to reference either `OpenAiAudioTranscriptionDriver` or `DummyAudioTranscriptionDriver`
- Add tests and documentation for added constructs

----
📚 Documentation preview 📚: https://griptape--781.org.readthedocs.build//781/


griptape/engines/audio/audio_transcription_engine.py
collindutter, line 1: Outside the scope of this PR but I'm starting to doubt our pattern of creating Engines that are just thin wrappers over Drivers. Engines should augment 1 or more Drivers for a specific use-case that a single Driver cannot do on its' own.

What do you think?
CC @vasinov

andrewfrench: I strongly agree, this layer is way too thin, but I also did not want to create a weird single audio-to-text modality that breaks our established pattern. Optimistically, they're a layer where we can add additional functionality. For now, they're mostly an additional initialization step.

griptape/drivers/audio_transcription/base_audio_transcription_driver.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def before_run(self) -> None:
+        if self.structure:
+            self.structure.publish_event(StartAudioTranscriptionEvent())
+
+    def after_run(self) -> None:
+        if self.structure:
+            self.structure.publish_event(FinishAudioTranscriptionEvent())
+
+    def run_transcription(self, audio: AudioArtifact, prompts: Optional[list[str]] = None) -> TextArtifact:
```
collindutter, line 29: Just for consistency with other Drivers should we rename this to `run/try_run`?

griptape/tools/transcription_client/tool.py
```diff
@@ -0,0 +1,55 @@
...
+from griptape.artifacts import ErrorArtifact, AudioArtifact, TextArtifact
+from griptape.engines import AudioTranscriptionEngine
+from griptape.loaders.audio_loader import AudioLoader
+from griptape.tools import BaseTool
+from griptape.utils import load_artifact_from_memory
+from griptape.utils.decorators import activity
+
+
+@define
+class TranscriptionClient(BaseTool):
```
collindutter, line 17: Maybe rename to `AudioTranscriptionClient`?

griptape/drivers/audio_transcription/base_audio_transcription_driver.py
```diff
@@ -0,0 +1,42 @@
...
+from griptape.artifacts import TextArtifact, AudioArtifact
+from griptape.events import StartAudioTranscriptionEvent, FinishAudioTranscriptionEvent
+from griptape.mixins import ExponentialBackoffMixin, SerializableMixin
+
+if TYPE_CHECKING:
+    from griptape.structures import Structure
+
+
+@define
+class BaseAudioTranscriptionDriver(SerializableMixin, ExponentialBackoffMixin, ABC):
```
collindutter, line 17: Should we name these Drivers `BaseSpeechToTextDriver` for consistency with the inverse Drivers?

collindutter (commented):
Nice work, just naming comments

griptape/drivers/audio_transcription/base_audio_transcription_driver.py
andrewfrench: I'm on the fence about this one. I think I steered a little too far when setting the pattern of drivers based on their artifact interfaces and perhaps too far from describing _what they actually do_ (text in audio out, image in text out, text+image in image out, etc). I think we should probably group drivers by function rather than form in these cases for specificity.

andrewfrench: I think I've steered a bit too far in the direction of naming drivers based on their artifact interfaces. I think the specificity of this name is helpful, what do you think about adding a similarly specific name to the `BaseTextToSpeechDriver`? `BaseSpeechGenerationDriver`?

collindutter: Yeah thinking about it more I think this current convention is the more "correct" one.

Down to do a rename of `BaseTextToSpeechDriver` (though maybe in a separate PR), what do you think about `BaseAudioGenerationDriver`? This may extend beyond speech in the future?

collindutter (approved):

collindutter (approved):
Great work!

collindutter (dismissed):

collindutter (approved):

PR #785, opened 2024-05-15
Release v0.25.1

andrewfrench:
### Fixed
- Honor `namespace` in `RedisVectorStoreDriver.query()`.
- Correctly set the `meta`, `score`, and `vector` fields of query result returned from `RedisVectorStoreDriver.query()`.
- Standardize behavior between omitted and empty actions list when initializing `ActionsSubtask`.

### Added
- Optional event batching on Event Listener Drivers.
- `id` field to all events.

### Changed
- Default behavior of Event Listener Drivers to batch events.
- Default behavior of OpenAiStructureConfig to utilize `gpt-4o` for prompt_driver.


vasinov (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Correctly set the `meta`, `score`, and `vector` fields of query result returned from `RedisVectorStoreDriver.query()`.
+- Standardize behavior between omitted and empty actions list when initializing `ActionsSubtask`.
+
+### Added
+- Optional event batching on Event Listener Drivers.
+- `id` field to all events.
+
+### Changed
+- Default behavior of Event Listener Drivers to batch events.
+- Default behavior of OpenAiStructureConfig to utilize `gpt-4o` for prompt_driver.
```
emjay07, line 22: if our default has changed, will this be breaking to customers?

andrewfrench: No, developers should be able to use this in exactly the same way as the previous `gpt-4` default. We updated the default because the new model is faster, better, and less expensive.

vachillo (approved):

PR #791, opened 2024-05-17
Feature/native tool calling

collindutter:
This PR adds the ability for prompt drivers to use native function calling over the Griptape provided approach (regex).

To do this, the following things have been changed:
- Added `PrompStack.TOOL_CALL_ROLE`, and `PromptStack.TOOL_RESULT_ROLE`. Depending on the the LLM provider, these may get translated into `user` or `assistant` roles.
- Added `ActionsArtifact` for tracking function calls made by the LLM.
- Added `ActionChunkArtifact` for tracking partial function calls made by the LLM when streaming.
- Moved `ActionsSubtask.Action` to `ActionsArtifact.Action` so that Prompt Drivers could more easily interact with Actions.
- Changed `PromptStack.content: str` to `PromptStack.content: str | BaseArtifact`. Right now the type of content depends on the role (Tool roles use `ActionsArtifact`), but this will be changed in the next PR that breaks up PromptStack.Input into distinct classes (`PromptStackTextInput`, `PromptStackActionsInput`, `PromptStackImageInput` etc.)
- Changed `ActionsSubtask.input: str` to `ActionsSubtask.input: str | ActionsArtifact.Action`. If a string is provided, the Task will parse it into an `ActionsSubtask.Action` using regex (current behavior). If an `ActionsSubtask.Action` is provided, it will use this (new behavior).
- Added `ActionsSubtask.single_action` flag for use with `ToolTask` to denote that the incoming action string is not a list.  The parsing logic changes slightly when this flag is set to true.
- Added `BasePromptDriver.use_native_tools` to toggle between old regex tool usage and native tool usage.
- Updated `OpenAiChatPromptDriver` to use new Prompt Stack tool inputs
- Updated `AnthropicChatPromptDriver` to use new Prompt Stack tool inputs.
- Updated `toolkit_task/system.j2` prompt to conditionally render CoT wording based on `use_native_tools`.
- Updated `tool_task/system.j2` prompt to conditionally render action schema instructions based on `use_native_tools`.


griptape/artifacts/action_chunk_artifact.py
```diff
@@ -0,0 +1,49 @@
...
+from typing import Optional
+
+from attr import define, field
+
+from griptape.artifacts import BaseArtifact, TextArtifact
+from griptape.mixins import SerializableMixin
+
+
+@define
+class ActionChunkArtifact(TextArtifact, SerializableMixin):
```
collindutter, line 12: Really don't like this class, but it's required for when we're streaming tool usage. Should we change `ActionsArtifact.Action` into `ActionArtifact`? Then we can yield `ActionArtifact`s much like we yield `TextArtifact`s during streaming.

dylanholmes: Are you saying that you would yield an `ActionArtifact` after getting all the `ActionChunkArtifact` for that action, instead of yielding the `ActionChunkArtifact`s themselves? (Or would you yield both?)

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -64,6 +75,8 @@ class OpenAiChatPromptDriver(BasePromptDriver):
         ),
         kw_only=True,
     )
+    tool_choice: str = field(default="auto", kw_only=True, metadata={"serializable": False})
```
dylanholmes, line 78: There's no doc comment for this. What other values would you supply other than `"auto"`?

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -79,7 +92,25 @@ def try_run(self, prompt_stack: PromptStack) -> TextArtifact:
...
-            return TextArtifact(value=parsed_result.choices[0].message.content.strip())
+            message = parsed_result.choices[0].message
+            tool_calls: list[ChatCompletionMessageToolCall] = message.tool_calls
+
+            if tool_calls:
+                actions = [
+                    ActionsArtifact.Action(
+                        tag=tool_call.id,
+                        name=tool_call.function.name.split("-")[0],
+                        path=tool_call.function.name.split("-")[1],
```
dylanholmes, line 103: Why do you put these split these and put them into `name` and `path` rather than just putting `function.name` directly into `name`?

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -79,7 +92,25 @@ def try_run(self, prompt_stack: PromptStack) -> TextArtifact:
...
 
         if len(parsed_result.choices) == 1:
-            return TextArtifact(value=parsed_result.choices[0].message.content.strip())
+            message = parsed_result.choices[0].message
+            tool_calls: list[ChatCompletionMessageToolCall] = message.tool_calls
+
+            if tool_calls:
+                actions = [
+                    ActionsArtifact.Action(
+                        tag=tool_call.id,
```
dylanholmes, line 101: Can multiple `Action`s have the same `tag`? If so, should it be called `id` instead of `tag`? I might just generally be missing the reasoning behind each of the fields in `Action`

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -79,7 +92,25 @@ def try_run(self, prompt_stack: PromptStack) -> TextArtifact:
...
+                        name=tool_call.function.name.split("-")[0],
+                        path=tool_call.function.name.split("-")[1],
+                        input=json.loads(tool_call.function.arguments),
+                    )
+                    for tool_call in tool_calls
+                ]
+
+                return ActionsArtifact(actions=actions)
```
dylanholmes, line 109: Should we be passing the message like in the anthropic prompt driver (`value=message.content.strip()`)?

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -89,13 +120,35 @@ def try_stream(self, prompt_stack: PromptStack) -> Iterator[TextArtifact]:
...
-            else:
-                raise Exception("Completion with more than one choice is not supported yet.")
+                tool_call_deltas = delta.tool_calls
+
+                if tool_call_deltas:
+                    if len(tool_call_deltas) == 1:
+                        tool_call_delta = tool_call_deltas[0]
+
+                        if tool_call_delta.id is not None:
+                            name, path = tool_call_delta.function.name.split("-")
```
dylanholmes, line 130: Did you mean to check `tool_call_delta.function.name` (rather than `tool_call_delta.id`)?

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -89,13 +120,35 @@ def try_stream(self, prompt_stack: PromptStack) -> Iterator[TextArtifact]:
...
+                            index=tool_call_delta.index,
+                            tag=tool_call_delta.id,
+                            name=name,
+                            path=path,
+                            partial_input=tool_call_delta.function.arguments,
+                        )
+                    else:
+                        raise Exception("Streaming completion with more than one tool call delta is not supported yet.")
+                else:
+                    content_delta = delta.content or ""
```
dylanholmes, line 147: Is it ever possible for both `delta.content` and `delta.tool_calls` to be present / non-None?

griptape/drivers/prompt/anthropic_prompt_driver.py
```diff
@@ -31,47 +41,151 @@ class AnthropicPromptDriver(BasePromptDriver):
...
+                content_block = chunk.content_block
+                content_block_type = content_block.type
 
-    def _prompt_stack_to_model_input(self, prompt_stack: PromptStack) -> dict:
+                if content_block_type == "text":
+                    yield TextArtifact(value=chunk.content_block.text)
+                elif content_block_type == "tool_use":
+                    name, path = content_block.name.split("-")
+                    yield ActionChunkArtifact(
+                        value=f"{name}-{path}", index=chunk.index, tag=content_block.id, name=name, path=path
```
dylanholmes, line 81: Its a little confusing to sometimes set `value` to the "function name" (e.g. `name-path`) and sometimes set it to `partial_input`. It feels like we want two different types of chunks.

Or maybe just add some note in the doc string about how to interpret `value` in the context of `ActionChunkArtifact`.

griptape/drivers/prompt/anthropic_prompt_driver.py
```diff
@@ -31,47 +41,151 @@ class AnthropicPromptDriver(BasePromptDriver):
     )
     top_p: float = field(default=0.999, kw_only=True, metadata={"serializable": True})
     top_k: int = field(default=250, kw_only=True, metadata={"serializable": True})
+    tool_choice: dict = field(default=Factory(lambda: {"type": "auto"}), kw_only=True, metadata={"serializable": False})
+    use_native_tools: bool = field(default=True, kw_only=True, metadata={"serializable": True})
```
dylanholmes, line 45: doc strings for these would be nice

dylanholmes (commented):
Almost got my head around everything. I think I'll need one more review though

collindutter:
Closing in favor of https://github.com/griptape-ai/griptape/pull/867

PR #794, opened 2024-05-20
Add codecov

collindutter:


.github/workflows/code-checks.yml
```diff
@@ -46,3 +46,23 @@ jobs:
         uses: ./.github/actions/init-environment
       - name: Run linter
         run: poetry run ruff check griptape/
+  coverage:
+      runs-on: ubuntu-latest
+      strategy:
+        fail-fast: false
+        matrix:
+          python-version: ["3.9"]
```
vachillo, line 54: should we do `3.11`?

vachillo (approved):

vasinov (approved):

.github/workflows/code-checks.yml
collindutter: Other workflows we've followed the pattern of running on the lowest supported version. Not saying that's necessarily correct but 🤷

collindutter: Updated to `3.12`

vachillo (approved):

PR #797, opened 2024-05-21
Makefile target for big install command

andrewfrench:
Our Poetry dependency installation command is pretty long. This captures it in the Makefile as `make install`.


dylanholmes (approved):
Nice work, I like this idea

collindutter:
Nice call. IMO we should do this for all common commands. I.e. `make code-checks/lint`, `make test`, etc.

Would make it easier to ensure the checks being run in pre-commit, GitHub actions, and manually are all the same.

PR #806, opened 2024-05-24
Refactor/artifact chunks

collindutter:
Splitting out some of the work from https://github.com/griptape-ai/griptape/pull/791.

This PR introduces several new Artifact types, and makes the distinction between "full" Artifacts and "chunk" Artifacts. While the difference between a `TextArtifact` and `TextChunkArtifact` is negligible, as we get into more advanced modalities types like function calls, images, and audio, the difference is significant.

The "full" classes have a `from_chunks` method to assemble a "full" Artifact from a list of partial chunks.


dylanholmes (approved):
Nice work! Thanks for splitting this change into its own PR!

Only minor comments

PR #807, opened 2024-05-24
Add `overrides` parameter on `BaseStructureConfig`, remove `merge_config`

vachillo:
add `overrides` to `BaseStructureConfig` to allow overriding basic fields without having to redefine them

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `AzureOpenAiStructureConfig` for providing Structures with all Azure OpenAI Driver configuration.
 - `AzureOpenAiVisionImageQueryDriver` to support queries on images using Azure's OpenAI Vision models.
+- Parameter `overrides` to `BaseStructureConfig` to support overriding values of Structure config defaults.
 
 ### Changed
 - **BREAKING**: Updated OpenAI-based image query drivers to remove Vision from the name.
+- **BREAKING**: Removed method `merge_config` from `BaseStructureConfig`.
 - Default the value of `azure_deployment` on all Azure Drivers to the model the Driver is using.
 - Field `azure_ad_token` on all Azure Drivers is no longer serializable.
 - Default standard OpenAI and Azure OpenAI image query model to `gpt-4o`.
--- a/griptape/config/azure_openai_structure_config.py
+++ b/griptape/config/azure_openai_structure_config.py
@@ -40,7 +40,9 @@ class AzureOpenAiStructureConfig(StructureConfig):
     api_key: Optional[str] = field(kw_only=True, default=None, metadata={"serializable": False})
     prompt_driver: BasePromptDriver = field(
         default=Factory(
-            lambda self: AzureOpenAiChatPromptDriver(
+            lambda self: self._factory(
+                AzureOpenAiChatPromptDriver,
+                "prompt_driver",
                 model="gpt-4o",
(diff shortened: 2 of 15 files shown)


griptape/config/azure_openai_structure_config.py
```diff
@@ -96,7 +103,15 @@ class AzureOpenAiStructureConfig(StructureConfig):
...
+                LocalVectorStoreDriver, "vector_store_driver", embedding_driver=self.embedding_driver
+            ),
+            takes_self=True,
+        ),
         metadata={"serializable": True},
         kw_only=True,
     )
+
+    def __attrs_post_init__(self):
+        self.vector_store_driver = LocalVectorStoreDriver(embedding_driver=self.embedding_driver)
```
collindutter, line 117: Can we do `Factory(lambda self: self.embedding_driver)` here?

vachillo: there might have been an issue where the embedding driver was not getting initialized before the Factory method was being called, resulting in the vector store driver having a `DummyEmbeddingDriver`. still trying to reproduce

CHANGELOG.md
```diff
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `AzureOpenAiStructureConfig` for providing Structures with all Azure OpenAI Driver configuration.
 - `AzureOpenAiVisionImageQueryDriver` to support queries on images using Azure's OpenAI Vision models.
+- Parameter `overrides` to `BaseStructureConfig` to support overriding values of Structure config defaults.
 
 ### Changed
 - **BREAKING**: Updated OpenAI-based image query drivers to remove Vision from the name.
+- **BREAKING**: Removed method `merge_config` from `BaseStructureConfig`.
```
collindutter, line 16: Maybe worth mentioning "...in favor of `overrides`"

griptape/config/base_structure_config.py
```diff
@@ -29,9 +28,17 @@ class BaseStructureConfig(BaseConfig, ABC):
...
-        base_config = self.to_dict()
-        merged_config = dict_merge(base_config, config)
+    def _factory(self, cls: type, overrides_namespace: str, **params):
+        """Factory method to create a new instance of the class with the given parameters
 
-        return BaseStructureConfig.from_dict(merged_config)
+        Args:
+            :param cls: the class to instantiate
+            :param overrides_namespace: the namespace to look for values in the overrides dictionary
+            :param params: the parameters to pass to the class
```
collindutter, line 39: Nit the `:param ...:` syntax is not necessary

griptape/config/base_structure_config.py
```diff
@@ -29,9 +28,17 @@ class BaseStructureConfig(BaseConfig, ABC):
...
+
+        Returns:
+            :return: a new instance of the class
```
collindutter, line 42: Same with `:return:`. Search `Returns` for examples.

griptape/config/azure_openai_structure_config.py
```diff
@@ -40,7 +40,9 @@ class AzureOpenAiStructureConfig(StructureConfig):
     api_key: Optional[str] = field(kw_only=True, default=None, metadata={"serializable": False})
     prompt_driver: BasePromptDriver = field(
         default=Factory(
```
collindutter, line 42: Do we need to use attr's `Factory` now that we're doing our own? Can we somehow extend/modify the existing `Factory`?

collindutter (changes requested):

PR #809, opened 2024-05-28
Add Workflow.set_tasks_from_layers()

dylanholmes:
Changes:
- Add `Workflow.set_tasks_from_layers()`, see unit tests for example usage.

### DIFF
--- a/griptape/structures/workflow.py
+++ b/griptape/structures/workflow.py
@@ -89,6 +89,35 @@ class Workflow(Structure):
 
         return task
 
+    def set_tasks_from_layers(self, layers: list[BaseTask | list[BaseTask]]) -> None:
+        if len(layers) < 1:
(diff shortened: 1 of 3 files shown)


griptape/structures/workflow.py
```diff
@@ -89,6 +89,35 @@ def insert_task(
 
         return task
 
+    def set_tasks_from_layers(self, layers: list[BaseTask | list[BaseTask]]) -> None:
```
collindutter, line 92: Can we just override `add_tasks` instead of introducing a new method?

collindutter (commented):
Can you please update the docs to use whatever syntax we land on here?

collindutter:
Also, what happened to your syntax proposed offline? I think I like the flexibility of something like that over this that only supports a relatively specific set of topologies.

dylanholmes:
Discarding this in favor of a PR that modifies the api for BaseTask.

PR #811, opened 2024-05-31
Refactor tasks to use ListArtifact as input

collindutter:
Several Tasks have non-`BaseArtifact` `input` fields. For instance, `ImageQueryTask.input` has a type of:

```
tuple[TextArtifact, list[ImageArtifact]]
```

The complexity of this type makes it very difficult to serialize the inputs of these Tasks in downstream abstractions like Events. Additionally, it make the Task API inconsistent across Task types.

This PR changes the input type to `ListArtifact`, making it consistent with other Task types. Note that the `setter` for the field is mostly unchanged, only the `getter` has changed.

### DIFF
--- a/griptape/tasks/image_query_task.py
+++ b/griptape/tasks/image_query_task.py
@@ -4,7 +4,7 @@ from typing import Callable
 
 from attrs import define, field
 
-from griptape.artifacts import ImageArtifact, TextArtifact
+from griptape.artifacts import ImageArtifact, ListArtifact, TextArtifact
 from griptape.engines import ImageQueryEngine
 from griptape.tasks import BaseTask
 from griptape.utils import J2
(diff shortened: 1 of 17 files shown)


griptape/tasks/image_query_task.py
```diff
@@ -49,8 +47,11 @@ def input(self) -> tuple[TextArtifact, list[ImageArtifact]]:
     @input.setter
     def input(
         self,
-        value: tuple[TextArtifact, list[ImageArtifact]]
-        | Callable[[BaseTask], tuple[TextArtifact, list[ImageArtifact]]],
+        value: (
+            tuple[str, list[ImageArtifact]]
+            | tuple[TextArtifact, list[ImageArtifact]]
+            | Callable[[BaseTask], ListArtifact]
```
andrewfrench, line 53: My one consideration was whether or not it's useful to accept a `ListArtifact` here directly, similarly to how the `Callable` now just returns a `ListArtifact`. I don't think I've run into a situation where this would have made anything more convenient, and these input type unions are already unwieldy enough, but curious what you think.

griptape/tasks/image_query_task.py
```diff
@@ -68,8 +69,9 @@ def image_query_engine(self, value: ImageQueryEngine) -> None:
         self._image_query_engine = value
 
     def run(self) -> TextArtifact:
-        query, image_artifacts = self.input
+        query = self.input.value[0]
+        image_artifacts = [input for input in self.input.value[1:] if isinstance(input, ImageArtifact)]
```
andrewfrench, line 73: This quietly drops non-image inputs, should we raise an exception if `self.input.value[1:]` aren't all `ImageArtifact`s like in the in/outpainting tasks?

andrewfrench (approved):
Two tiny comments, very cool to be able to support this change and not worry about breaking interfaces for users!

collindutter:
@andrewfrench thanks for the review, just fixed both of your comments.

andrewfrench (approved):

PR #814, opened 2024-05-31
Include extra fields

collindutter:


vachillo (commented):
this will fail if an unknown field is passed to `from_dict`. is that expected?

collindutter:
> this will fail if an unknown field is passed to from_dict. is that expected?

I think so, it would be the same behavior if you instantiated the object yourself with an unknown field.

collindutter:
@vachillo can we close https://github.com/griptape-ai/griptape/pull/807 in favor of this PR? Or is there still missing functionality?

PR #822, opened 2024-06-04
Add task relationship APIs

dylanholmes:
See CHANGLOG.md for the majority of changes and docs + unit tests for example usages.

One change not listed in the change log is the removal of `ActionSubTask.add_child()` and `ActionSubTask.add_parent()` due to effectively no change in behavior. The behavior of  `ActionSubTask.add_child()` and `ActionSubTask.add_parent()` has been moved up the hierarchy into `BaseTask` and changed to only define relationships asymmetrically in order to be consistent with the approach of passing parent or child ids into `BaseTask()`. This change in behavior required one update to `ToolKitTask.add_subtask` to invoke both `add_child` and `add_parent` since `Structure.resolve_relationships()` only effects tasks that have been added to `Structure` directly (which does not include sub tasks.

----
📚 Documentation preview 📚: https://griptape--822.org.readthedocs.build//822/

### DIFF
--- a/griptape/tasks/actions_subtask.py
+++ b/griptape/tasks/actions_subtask.py
@@ -172,24 +172,6 @@ class ActionsSubtask(BaseTextInputTask):
     def actions_to_json(self) -> str:
         return json.dumps(self.actions_to_dicts())
 
-    def add_child(self, child: ActionsSubtask) -> ActionsSubtask:
-        if child.id not in self.child_ids:
-            self.child_ids.append(child.id)
-
-        if self.id not in child.parent_ids:
-            child.parent_ids.append(self.id)
-
-        return child
-
-    def add_parent(self, parent: ActionsSubtask) -> ActionsSubtask:
-        if parent.id not in self.parent_ids:
-            self.parent_ids.append(parent.id)
-
-        if self.id not in parent.child_ids:
-            parent.child_ids.append(self.id)
-
-        return parent
-
     def __init_from_prompt(self, value: str) -> None:
         thought_matches = re.findall(self.THOUGHT_PATTERN, value, re.MULTILINE)
         actions_matches = re.findall(self.ACTIONS_PATTERN, value, re.DOTALL)
--- a/griptape/tasks/base_task.py
+++ b/griptape/tasks/base_task.py
@@ -60,7 +60,7 @@ class BaseTask(ABC):
     def __str__(self) -> str:
         return str(self.output.value)
 
-    def add_parents(self, *parents: str | BaseTask) -> None:
+    def add_parents(self, parents: list[str | BaseTask]) -> None:
         for parent in parents:
             self.add_parent(parent)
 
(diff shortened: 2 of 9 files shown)


griptape/tasks/base_task.py
```diff
@@ -60,6 +60,26 @@ def meta_memories(self) -> list[BaseMetaEntry]:
     def __str__(self) -> str:
         return str(self.output.value)
 
+    def add_parents(self, *parents: str | BaseTask) -> None:
```
collindutter, line 63: I think this should be `parents: list[str] | list[BaseTask]`. That way it won't break if we add  additional positional arguments in the future.

griptape/tasks/base_task.py
```diff
@@ -60,6 +60,26 @@ def meta_memories(self) -> list[BaseMetaEntry]:
...
+        for parent in parents:
+            self.add_parent(parent)
+
+    def add_parent(self, parent: str | BaseTask) -> None:
+        parent_id = parent if isinstance(parent, str) else parent.id
+
+        if parent_id not in self.parent_ids:
+            self.parent_ids.append(parent_id)
+
+    def add_children(self, *children: str | BaseTask) -> None:
```
collindutter, line 73: Same

griptape/tasks/actions_subtask.py
```diff
@@ -172,24 +172,6 @@ def actions_to_dicts(self) -> list[dict]:
...
-
-    def add_parent(self, parent: ActionsSubtask) -> ActionsSubtask:
-        if parent.id not in self.parent_ids:
-            self.parent_ids.append(parent.id)
-
-        if self.id not in parent.child_ids:
-            parent.child_ids.append(self.id)
-
-        return parent
-
```
collindutter, line 192: Is this necessary to remove? I understand it's for consistency, but isn't that what overriding methods is good for? Changing behavior in subclasses?

collindutter (commented):
Great work! Can you also add an example that shows how a user might use the new imperative methods?

griptape/tasks/base_task.py
dylanholmes: I agree I like an explicit list better (though I'd prefer `list[str | BaseTask]`).

I originally chose the variable args syntax for consistency with [Structure.add_tasks()](https://github.com/griptape-ai/griptape/blob/1042211e22e99425ea8867fb9c506bf913f5387a/griptape/structures/structure.py#L190).

griptape/tasks/actions_subtask.py
dylanholmes: Its not necessary to remove. I just thought it makes the logic easier to reason about.

I agree that overriding is for changing behavior in subclasses, but it can be abused. It seems unintuitive to change mutate objects passed in as arguments in a subclass only.

As an extreme example, imagine a class that has a method called `sort` which returns a `list`. It would be pretty strange if the base class returned a new list in sorted order, but a subclass sorted in-place and returned a reference to the same list.

That said, this case is not nearly that extreme, and I don't have the strongest opinion. I just gave the example as the intuition behind my position.

Do you still want me to revert this change?

collindutter: I'm sold, thanks for explaining!

collindutter (approved):
Great work! So much better.

PR #832, opened 2024-06-05
Updated `HuggingFacePipelinePromptDriver` to use chat features of `transformers.TextGenerationPipeline`.

collindutter:
----
📚 Documentation preview 📚: https://griptape--832.org.readthedocs.build//832/

### DIFF
--- a/griptape/drivers/prompt/huggingface_pipeline_prompt_driver.py
+++ b/griptape/drivers/prompt/huggingface_pipeline_prompt_driver.py
@@ -1,11 +1,16 @@
+from __future__ import annotations
 from collections.abc import Iterator
 
+from typing import TYPE_CHECKING
 from attrs import Factory, define, field
 
 from griptape.artifacts import TextArtifact
 from griptape.drivers import BasePromptDriver
 from griptape.tokenizers import HuggingFaceTokenizer
 from griptape.utils import PromptStack, import_optional_dependency
+
+if TYPE_CHECKING:
+    from transformers import TextGenerationPipeline
 
 
 @define
@@ -14,12 +19,8 @@ class HuggingFacePipelinePromptDriver(BasePromptDriver):
     Attributes:
         params: Custom model run parameters.
         model: Hugging Face Hub model name.
-        tokenizer: Custom `HuggingFaceTokenizer`.
 
     """
-
-    SUPPORTED_TASKS = ["text2text-generation", "text-generation"]
-    DEFAULT_PARAMS = {"return_full_text": False, "num_return_sequences": 1}
 
     max_tokens: int = field(default=250, kw_only=True, metadata={"serializable": True})
     model: str = field(kw_only=True, metadata={"serializable": True})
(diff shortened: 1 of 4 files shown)


griptape/drivers/prompt/huggingface_pipeline_prompt_driver.py
```diff
@@ -34,28 +35,38 @@ class HuggingFacePipelinePromptDriver(BasePromptDriver):
         ),
         kw_only=True,
     )
+    pipe: TextGenerationPipeline = field(
+        default=Factory(
+            lambda self: import_optional_dependency("transformers").pipeline(
+                "text-generation", model=self.model, max_new_tokens=self.max_tokens, tokenizer=self.tokenizer.tokenizer
```
dylanholmes, line 41: Do you know what will happen if you pass model that doesn't support he `text-generation` task?

griptape/drivers/prompt/huggingface_pipeline_prompt_driver.py
```diff
@@ -34,28 +35,38 @@ class HuggingFacePipelinePromptDriver(BasePromptDriver):
...
-            tokenizer=self.tokenizer.tokenizer,
-            model=self.model,
-            max_new_tokens=self.tokenizer.count_output_tokens_left(prompt),
+        result = self.pipe(
+            messages,
+            max_new_tokens=self.max_tokens,
+            eos_token_id=[
+                self.tokenizer.tokenizer.eos_token_id,
+                *[self.pipe.tokenizer.convert_tokens_to_ids(token) for token in self.tokenizer.stop_sequences],
+            ],
```
dylanholmes, line 56: I'm having trouble finding the docs related to this one, could you share your sources?

I would have expect `eos_token_id` to be set automatically, since the pipeline has access to the model and tokenizer. The one case where I'd expect this to be required is for model specific differences. Did you test this will multiple models?

> Note that I did not see `eos_token_id` specified in the hugging face example usage of [TinyLlama/TinyLlama-1.1B-Chat-v1.0](https://huggingface.co/TinyLlama/TinyLlama-1.1B-Chat-v1.0)

dylanholmes (approved):
Another nicely simplified prompt driver example!

Just minor comments that I'll leave to your discretion.

griptape/drivers/prompt/huggingface_pipeline_prompt_driver.py
collindutter: It'll probably break. But part of this refactor allows users to pass in their own `pipe` if they'd like to use a different task type.

griptape/drivers/prompt/huggingface_pipeline_prompt_driver.py
collindutter: I wasn't too sure about this TBH. I got it from [this article](https://huggingface.co/blog/llama3#using-%F0%9F%A4%97-transformers). Seems like if you override `eos_token_id` you need to provide the original `self.tokenizer.tokenizer.eos_token_id`. And then we want to include our own stop tokens on top of that.

I really can't find much documentation on how to properly set stop sequences.

collindutter: Found the [documentation](https://huggingface.co/docs/transformers/v4.41.3/en/main_classes/text_generation#generation) for the parameters and updated to use `stop_strings` which seems like what we want.

dylanholmes (approved):

PR #837, opened 2024-06-06
Reduce scope of Amazon SageMaker Prompt Driver to only Jumpstart

collindutter:
By reducing the scope of our integration with Amazon SageMaker to just Jumpstart, we can reduce the request/response variance, getting rid of the need for Prompt Model Drivers.


docs/griptape-framework/drivers/prompt-drivers.md
```diff
@@ -347,87 +347,35 @@ agent = Agent(
...
 
 agent = Agent(
     config=StructureConfig(
-        prompt_driver=AmazonSageMakerPromptDriver(
-            endpoint=os.environ["SAGEMAKER_FALCON_ENDPOINT_NAME"],
-            model=os.environ["SAGEMAKER_FALCON_INFERENCE_COMPONENT_NAME"],
-            prompt_model_driver=SageMakerFalconPromptModelDriver(),
+        prompt_driver=AmazonSageMakerJumpstartPromptDriver(
+            endpoint=os.environ["SAGEMAKER_LLAMA_3_INSTRUCT_ENDPOINT_NAME"],
+            model="meta-llama/Meta-Llama-3-8B-Instruct",
```
dylanholmes, line 375: We should add an `inference_component` parameter so that you can invoke endpoints with multiple models.

> Inference component name had been passed via the `model` parameter in `SageMakerLlamaPromptModelDriver`, but that was probably not the best name anyway since it isn't really related to the model id.

The "newer" SageMaker Studio JumpStart creates multiple model endpoints by default. Actually, I haven't even found a way to create single model endpoints in the newer studio.

dylanholmes (changes requested):
Love the direction!

docs/griptape-framework/drivers/prompt-drivers.md
collindutter: Updated

collindutter:
@dylanholmes also updated PR to include embedding drivers.

dylanholmes (approved):

collindutter:
Ignoring integration test, will be fixed in separate PR.

PR #839, opened 2024-06-06
Improve error when knowledgebase is missing a description

collindutter:


griptape/tools/griptape_cloud_knowledge_base_client/tool.py
```diff
@@ -52,4 +52,6 @@ def _get_knowledge_base_description(self) -> str:
             if "description" in response:
                 return response["description"]
             else:
-                raise ValueError(f'Error getting Knowledge Base description: {response["message"]}')
+                raise ValueError(
+                    "No description found for Knowledge Base. Please set a description, or manually set the `GriptapeCloudKnowledgeBaseClient.description` attribute."
```
vachillo, line 56: maybe throw the id in the message?

vachillo (approved):

griptape/tools/griptape_cloud_knowledge_base_client/tool.py
collindutter: Does that come back in the response?

vachillo: not sure, but cant you pull it from `self. knowledge_base_id`?

collindutter: Oh duh

vachillo (approved):

PR #840, opened 2024-06-06
Refactor/tokenizers

collindutter:
### Added
- `BaseTokenizer.prompt_stack_to_string()` to convert a Prompt Stack to a string.
- `BaseTokenizer.prompt_stack_input_to_string()` to convert a Prompt Stack Input to a ChatML-style message dictionary.
### Changed
- **BREAKING**: Removed `BasePromptDriver.count_tokens()`.
- **BREAKING**: Removed `BasePromptDriver.max_output_tokens()`.
- **BREAKING**: Moved `BasePromptDriver.prompt_stack_to_string()` to `BaseTokenizer`.
- **BREAKING**: Moved/renamed `PromptStack.add_to_conversation_memory` to `BaseConversationMemory.add_to_prompt_stack`.
- `BaseTokenizer.count_tokens()` can now approximately token counts given a Prompt Stack.
- Updated Prompt Drivers to use `BasePromptDriver.max_tokens` instead of using `BasePromptDriver.max_output_tokens()`.
- **BREAKING**: Moved `griptape.constants.RESPONSE_STOP_SEQUENCE` to `ToolkitTask`.
- `ToolkitTask.RESPONSE_STOP_SEQUENCE` is now only added when using `ToolkitTask`.


griptape/tokenizers/base_tokenizer.py
```diff
@@ -1,50 +1,98 @@
...
+            A dictionary with the role and content of the input.
+        """
+        content = prompt_input.content
+
+        if prompt_input.is_system():
+            return {"role": "system", "content": content}
+        elif prompt_input.is_assistant():
+            return {"role": "assistant", "content": content}
+        else:
+            return {"role": "user", "content": content}
```
andrewfrench, line 89: Should this be a method on `PromptInput`?

griptape/tokenizers/base_tokenizer.py
```diff
@@ -1,50 +1,98 @@
...
+        else:
+            return self.try_count_tokens(text)
+
     @abstractmethod
-    def count_tokens(self, text: str | list[dict]) -> int: ...
+    def try_count_tokens(self, text: Any) -> int: ...
+
+    def prompt_stack_to_string(self, prompt_stack: PromptStack) -> str:
+        """Converts a Prompt Stack to a string for token counting or model input.
+        This base implementation will not be very accurate, and should be overridden by subclasses with model-specific tokens.
```
andrewfrench, line 51: Originally [same question](https://github.com/griptape-ai/griptape/pull/840/files#r1633879326) here, but I sense that the model-specific method in its own tokenizers means this should be owned by the tokenizer.

griptape/tokenizers/base_tokenizer.py
```diff
@@ -1,50 +1,98 @@
...
+            return {"role": "assistant", "content": content}
+        else:
+            return {"role": "user", "content": content}
 
     def _default_max_input_tokens(self) -> int:
         tokens = next((v for k, v in self.MODEL_PREFIXES_TO_MAX_INPUT_TOKENS.items() if self.model.startswith(k)), None)
 
         if tokens is None:
-            raise ValueError(f"Unknown model default max input tokens: {self.model}")
+            return self.DEFAULT_MAX_INPUT_TOKENS
```
andrewfrench, line 95: Seems reasonable to fall back to the default here. Do you think it'd be worth it to throw out a warning if this happens? It seems like a big enough difference in expectations between the known values defined in the mapping and a default 'best guess' value.

griptape/tokenizers/base_tokenizer.py
collindutter: I don't think so, each Tokenizer will have a slightly different implementation of this. In an upcoming PR i'm changing this to an abstract method.

griptape/tokenizers/base_tokenizer.py
collindutter: Honestly we could probably get rid of this method. All of our Prompt Drivers now support message-style APIs.

griptape/tokenizers/base_tokenizer.py
collindutter: Yeah warning is a good idea.

griptape/tokenizers/base_tokenizer.py
collindutter: Ah but now I remember why I didn't remove in this PR: I didn't want to remove the token counts in the events which relies on this method. My next PR changes this functionality so I will remove this method.

dylanholmes (commented):
Just a couple of questions

griptape/drivers/prompt/huggingface_pipeline_prompt_driver.py
```diff
@@ -68,3 +64,21 @@ def try_run(self, prompt_stack: PromptStack) -> TextArtifact:
...
+
+    def _prompt_stack_input_to_message(self, prompt_input: PromptStack.Input) -> dict:
+        return {"role": prompt_input.role, "content": prompt_input.content}
+
+    def __prompt_stack_to_tokens(self, prompt_stack: PromptStack) -> list[int]:
+        tokens = self.tokenizer.tokenizer.apply_chat_template(
+            [self._prompt_stack_input_to_message(i) for i in prompt_stack.inputs],
+            add_generation_prompt=True,
+            tokenize=True,
+        )
```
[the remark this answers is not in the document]
collindutter: In the next PR I will be using the raw tokens, this is mainly a set-up.

griptape/tasks/prompt_task.py
```diff
@@ -34,7 +34,7 @@ def prompt_stack(self) -> PromptStack:
 
         if memory:
             # inserting at index 1 to place memory right after system prompt
-            stack.add_conversation_memory(memory, 1)
+            memory.add_to_prompt_stack(stack, 1)
```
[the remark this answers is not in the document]
collindutter: Identical behavior, but the responsibility is better suited for Conversation Memory rather than Prompt Stack. Prompt Stack is intended to be a relatively lightweight abstraction and should not have use-case specific functionality like this in it.

dylanholmes (approved):
Nice work!

dylanholmes (approved):

PR #842, opened 2024-06-06
Add new Workflow/Task helper methods

collindutter:


griptape/structures/workflow.py
```diff
@@ -127,9 +127,8 @@ def context(self, task: BaseTask) -> dict[str, Any]:
 
         context.update(
             {
-                "parent_outputs": {
-                    parent.id: parent.output.to_text() if parent.output else "" for parent in task.parents
-                },
+                "parent_outputs": task.parent_outputs,
+                "parents_output": task.parents_output,
```
collindutter, line 131: Reeeeally don't like this naming but I don't have any other ideas.

dylanholmes: We definitely need to find some alternative for this.

How about changing `parents_output` to `parent_outputs_text`?

Not wild about this either, but it'd be less confusing/error prone.

docs/griptape-framework/structures/workflows.md
```diff
@@ -7,7 +7,8 @@ A [Workflow](../../reference/griptape/structures/workflow.md) is a non-sequentia
 
 Workflows have access to the following [context](../../reference/griptape/structures/workflow.md#griptape.structures.workflow.Workflow.context) variables in addition to the [base context](./tasks.md#context):
 
-* `parent_outputs`: outputs into the current task referenceable by parent task IDs.
+* `parent_outputs`: dictionary containing mapping of parent ids to their outputs.
+* `parents_output`: string containing the concatenated outputs of all parent tasks.
```
dylanholmes, line 11: Do we still want to be able to refer to parent outputs by index? If so, want to add something like a `parent_outputs_list` as a part of this change too?

dylanholmes (changes requested):
Nice job!

docs/griptape-framework/structures/workflows.md
andrewfrench: We can still index parents and to find outputs directly, right? Something like `parents[2].output`.

griptape/structures/workflow.py
andrewfrench: If `parent_outputs` is a dict of parent ids -> output could we name it something like `outputs_by_parent_id`? A little wordy but I agree we need some distinction here.

collindutter: I'd like to avoid renaming `parent_outputs` since that will be a breaking change on existing workflows. I do like `parent_outputs_text` but it's a bit misleading towards `parent_outputs` which is also text.

docs/griptape-framework/structures/workflows.md
collindutter: Yes ^

andrewfrench (dismissed):

docs/griptape-framework/structures/workflows.md
dylanholmes: Awesome, I'll try to remember this the next time it comes up!

dylanholmes (dismissed):

dylanholmes (approved):

PR #849, opened 2024-06-10
Fix streaming not working when using deprecated Structure.stream field

collindutter:


collindutter:
That's what this does: https://github.com/griptape-ai/griptape/blob/812013945170f53d654b40ab96bebfcad44d8a2c/griptape/structures/structure.py#L149-L156

vachillo (approved):

PR #851, opened 2024-06-11
Add observable

dylanholmes:
Changes:
- Add `BaseObservabilityDriver` as the base class for all observability drivers.
- Add `DummyObservabilityDriver` as a no-op observability driver.
- Add `OpenTelemetryObservabilityDriver` for sending observability data to an open telemetry collector or vendor.
- Add `GriptapeCloudObservabilityDriver` for sending observability data to Griptape Cloud.
- Add `Observability` for controlling which observability driver to use.
- Add `@observable` for selecting which functions/methods to provide observability for.
- Add docs at `structures/observability` and `drivers/observability-drivers`


griptape/utils/observability/griptape_cloud_observability.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self):
+        self.trace_provider = None
+        self.root_span_context_manager = None
+        self.exit_handler = None
+
+    def start(self):
+        # 1. Configure griptape exporter
+        resource = Resource(attributes={"service.name": "unknown"}, schema_url=None)
+        self.trace_provider = TracerProvider(resource=resource)
+        trace.set_tracer_provider(self.trace_provider)
```
dylanholmes, line 24: Remove this line.

griptape/utils/observability/griptape_cloud_observability.py
collindutter, line 1: We've already got a `utils.decorators`, maybe we should throw in there? Or `utils.decorators.observability`?

griptape/utils/observability/griptape_cloud_observability.py
```diff
@@ -0,0 +1,59 @@
...
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+
+class GriptapeCloudObservability:
+    def __init__(self):
+        self.trace_provider = None
+        self.root_span_context_manager = None
+        self.exit_handler = None
```
collindutter, line 18: Is there a reason this isn't an attrs class?

griptape/utils/observability/griptape_cloud_observability.py
```diff
@@ -0,0 +1,59 @@
...
+from griptape.utils.observability.griptape_instrumentor import GriptapeInstrumentor
+from opentelemetry import trace
+from opentelemetry.trace import Status, StatusCode
+from opentelemetry.instrumentation.threading import ThreadingInstrumentor
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+
+class GriptapeCloudObservability:
```
collindutter, line 14: Nit: `Observability` isn't really a noun like the other abstractions in the framework.

griptape/utils/observability/griptape_cloud_observability.py
```diff
@@ -0,0 +1,59 @@
...
+        self.trace_provider = None
+        self.root_span_context_manager = None
+        self.exit_handler = None
+
+    def start(self):
+        # 1. Configure griptape exporter
+        resource = Resource(attributes={"service.name": "unknown"}, schema_url=None)
+        self.trace_provider = TracerProvider(resource=resource)
+        trace.set_tracer_provider(self.trace_provider)
+        self.trace_provider.add_span_processor(BatchSpanProcessor(GriptapeCloudSpanExporter()))
```
collindutter, line 25: What if users want to swap out their own exporter?

griptape/utils/observability/griptape_cloud_observability.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # 2. Instrument library
+        GriptapeInstrumentor().instrument(tracer_provider=self.trace_provider)
+        ThreadingInstrumentor().instrument()
+
+        # 3. Create a root span, named "main"
+        tracer = trace.get_tracer("griptape", tracer_provider=self.trace_provider)
+        self.root_span_context_manager = tracer.start_as_current_span("main")  # pyright: ignore
```
collindutter, line 33: Why the pyright ignore? If necessary, be more specific i.e. https://github.com/griptape-ai/griptape/blob/888d50ecd5cab443fade30b8786c97d73ecf2e7a/griptape/tools/google_drive/tool.py#L236

griptape/utils/observability/griptape_cloud_observability.py
```diff
@@ -0,0 +1,59 @@
...
+        self.root_span_context_manager.__enter__()
+        trace.get_current_span().set_status(Status(StatusCode.OK))
+
+        def exit_handler():
+            if self.root_span_context_manager:
+                self.root_span_context_manager.__exit__(None, None, None)
+            if self.trace_provider:
+                self.trace_provider.shutdown()
```
collindutter, line 41: Should avoid nested functions.

griptape/utils/observability/griptape_cloud_observability.py
```diff
@@ -0,0 +1,59 @@
...
+
+        ThreadingInstrumentor().uninstrument()
+        GriptapeInstrumentor().uninstrument()
+
+
+if os.environ.get("GT_CLOUD_OBSERVABILITY") == "True":
+    if "GT_CLOUD_STRUCTURE_RUN_ID" not in os.environ:
+        raise ValueError("GT_CLOUD_OBSERVABILITY requires GT_CLOUD_STRUCTURE_RUN_ID to be set")
+    griptape_cloud_observability = GriptapeCloudObservability()
+    griptape_cloud_observability.start()
```
collindutter, line 59: I don't love the observability being so tightly coupled to Griptape Cloud. Griptape Cloud should just be one destination users can opt into.

griptape/utils/observability/griptape_cloud_span_exporter.py
```diff
@@ -0,0 +1,49 @@
...
+from urllib.parse import urljoin
+from typing_extensions import override
+from attrs import Factory, define, field
+from opentelemetry.sdk.trace import ReadableSpan
+from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+
+
+@define
+class GriptapeCloudSpanExporter(SpanExporter):
```
collindutter, line 12: This class seems suitable to make into a Driver -- even if it's not a traditional one. For instance, what if I wanted to connect to [Prometheus](https://opentelemetry.io/docs/languages/js/exporters/#prometheus) instead?

griptape/utils/observability/griptape_cloud_span_exporter.py
```diff
@@ -0,0 +1,49 @@
...
+    base_url: str = field(
+        default=Factory(lambda: os.getenv("GT_CLOUD_BASE_URL", "https://cloud.griptape.ai")), kw_only=True
+    )
+    api_key: str = field(default=Factory(lambda: os.getenv("GT_CLOUD_API_KEY")), kw_only=True)
+    headers: dict = field(
+        default=Factory(lambda self: {"Authorization": f"Bearer {self.api_key}"}, takes_self=True), kw_only=True
+    )
+    structure_run_id: str = field(default=Factory(lambda: os.getenv("GT_CLOUD_STRUCTURE_RUN_ID")), kw_only=True)
+    _underlying: SpanExporter = field(
```
collindutter, line 21: Is "underlying" an open telemetry term? Can we name it something like `exporter` instead?

griptape/utils/observability/griptape_cloud_span_exporter.py
```diff
@@ -0,0 +1,49 @@
+import os
+from collections.abc import Sequence
+from urllib.parse import urljoin
+from typing_extensions import override
```
collindutter, line 4: We don't use `typing_extensions` anywhere else in the framework, I think this may be provided from a dependency.

griptape/utils/observability/observable.py
collindutter, line 1: Is there a reason we're introducing a new dependency `wrapt` vs using functools like we do [here](https://github.com/griptape-ai/griptape/blob/ff6ddfd79614d97ea03d7e1c1dc1965f469d3ca2/griptape/utils/decorators.py#L18)?

pyproject.toml
```diff
@@ -28,6 +28,12 @@ stringcase = "^1.2.0"
 docker = "^7.1.0"
 sqlalchemy = "~=1.0"
 requests = "^2"
+wrapt = "^1.16.0"
+opentelemetry-sdk = "^1.25.0"
+opentelemetry-api = "^1.25.0"
+opentelemetry-instrumentation = "^0.46b0"
+opentelemetry-instrumentation-threading = "^0.46b0"
+opentelemetry-exporter-otlp = "^1.25.0"
```
collindutter, line 36: I think we need to figure out a way to get these as optional dependencies. Feels like a lot of baggage out of box.

griptape/utils/observability/griptape_cloud_observability.py
dylanholmes: Changed abstractions, which also now use attrs.

griptape/utils/observability/griptape_cloud_observability.py
dylanholmes: Technically it is a noun ([dictionary](https://www.merriam-webster.com/dictionary/observable#:~:text=%3A%20capable%20of%20being%20observed%20%3A%20discernible,noun)), but I'd also understand if you just generally dislike the naming choice. Do you have an alternative you'd prefer?

griptape/utils/observability/griptape_cloud_observability.py
dylanholmes: Refactored into a driver to allow switching implementations.

griptape/utils/observability/griptape_cloud_observability.py
dylanholmes: Made the pyright error more specific. It seems like the libraries isn't providing the correct type hints because pyright is complaining about not calling `start_as_current_span` with the right args, but they are right.

griptape/utils/observability/griptape_cloud_observability.py
dylanholmes: Switched to using a context manager to avoid needing this. There is still a potential use case not handled without having an exit handler, but we can solve that separately without putting it in here.

griptape/utils/observability/griptape_cloud_observability.py
dylanholmes: Remove the module level initialization. As of the current draft, the only way to enable it is in code like this:
```py
from griptape.observability import Observability
...
with Observability(driver=...):
    # normal griptape app code here
```

We can make a more seamless/sugary version in a later PR if needed or wanted.

griptape/utils/observability/griptape_cloud_span_exporter.py
dylanholmes: It comes from some past application of the decorator pattern. Mad naming more consistent in new abstractions.

griptape/utils/observability/griptape_cloud_span_exporter.py
dylanholmes: Removed, though it is nice to have an `@override` method to catch stupid naming typos that result in creating a new method instead of overriding an existing method.

griptape/utils/observability/observable.py
dylanholmes: Note that I replaced used of the wrapt  library with native python (though it was not easy). Now however, I need to re-engineer some of its magic to appease pyright (which was content with the wrapt version). I originally decided to use that library after skimming through https://github.com/GrahamDumpleton/wrapt/blob/develop/blog/01-how-you-implemented-your-python-decorator-is-wrong.md . (Plus we technically have it as an indirect dependency because the open telemetry libraries depend on it).

pyproject.toml
dylanholmes: Yeah, I'm planning on adding an extra, will happen in a subsequent draft (this is nice reminder!)

griptape/structures/agent.py
```diff
@@ -46,6 +47,7 @@ def add_tasks(self, *tasks: BaseTask) -> list[BaseTask]:
             raise ValueError("Agents can only have one task.")
         return super().add_tasks(*tasks)
 
+    @observable
```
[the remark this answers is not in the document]
dylanholmes: The choice is up to us!

Here's my reasoning: in some places we have `run`/`try_run` pairs, where the `try_run` is executed with retries. In these places I definitely think we should mark with `@observable`. Now the problem with doing this is that it instills an expectation as to what types of spans we expect to see in our traces. Therefore I decided to lean towards consistency and just always mark both `run` and `try_run` type pairs where ever they occur.

If you have an alternative perspective, let me know!

griptape/utils/observability/griptape_cloud_span_exporter.py
dylanholmes: Added drivers!

docs/griptape-framework/structures/observability.md
```diff
@@ -0,0 +1,56 @@
+## Overview
+
+The [Observability](../../reference/griptape/observability/observability.md) context manager sends telemetry (metrics and traces) for all functions and methods annotated with the `@observable` decorator to a destination of your choice. This is useful for monitoring and debugging your application.
+
+Observability is completely optional. To opt in, provide an `ObservabilityDriver` to the `Observability` context manager, for example:
```
collindutter, line 5: What happens if I don't provide a Driver?

collindutter, line 5: Link `ObservabilityDriver` to `drivers/observability-drivers.md`

docs/griptape-framework/structures/observability.md
```diff
@@ -0,0 +1,56 @@
...
+
+```python title="PYTEST_IGNORE"
+from griptape.drivers import BaseObservabilityDriver
+from griptape.rules import Rule
+from griptape.structures import Agent
+from griptape.observability import Observability
+
+driver: BaseObservabilityDriver = ... # Your choice of driver
```
collindutter, line 13: Please include a real Driver instantiation here.

docs/griptape-framework/structures/observability.md
```diff
@@ -0,0 +1,56 @@
...
+
+with Observability(driver=driver):
+    # Your code here
+    # Important! Only code within this block will be instrumented
```
collindutter, line 17: Please add real code example.

docs/griptape-framework/structures/observability.md
collindutter, line 5: Link `Observability` to reference docs.

docs/griptape-framework/structures/observability.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+For available drivers (and destinations), see [Observability Drivers](../drivers/observability-drivers.md).
```
collindutter, line 20: Capitalize drivers.

docs/griptape-framework/structures/observability.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Tracing Custom Code
+
+All functions and methods annotated with the `@observable` decorator will be traced when invoked within the context of the [Observability](../../reference/griptape/observability/observability.md) context manager, including functions and methods defined outside of the griptape framework. Thus to trace custom code, you just need to add the `@observable` decorator to your function or method, then invoke it within the [Observability](../../reference/griptape/observability/observability.md) context manager.
```
collindutter, line 24: Capitalize `griptape`

griptape/drivers/observability/base_observability_driver.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class BaseObservabilityDriver(ABC):
```
collindutter, line 5: Should be an attrs class (even if we're not using any attrs features atm)

griptape/drivers/observability/base_observability_driver.py
```diff
@@ -0,0 +1,13 @@
...
+    def __enter__(self) -> None:
+        pass
+
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
+        pass
```
collindutter, line 10: What is the point of these methods? Should they be marked as `abstractmethod`?

collindutter, line 10: `pass` should be replaced with `...`

griptape/drivers/observability/base_observability_driver.py
```diff
@@ -0,0 +1,13 @@
...
+
+    @abstractmethod
+    def invoke_observable(self, func, instance, args, kwargs, decorator_args, decorator_kwargs) -> Any: ...
```
collindutter, line 13: Missing type hints on parameters.

griptape/drivers/observability/base_observability_driver.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class BaseObservabilityDriver(ABC):
+    def __enter__(self) -> None:
+        pass
+
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
```
collindutter, line 9: Missing type hints on parameters.

griptape/drivers/observability/base_observability_driver.py
collindutter, line 13: This is quite a few positional arguments. Is this just the nature of the beast?

griptape/drivers/observability/dummy_observability_driver.py
```diff
@@ -0,0 +1,6 @@
+from griptape.drivers.observability.base_observability_driver import BaseObservabilityDriver
+
+
+class DummyObservabilityDriver(BaseObservabilityDriver):
+    def invoke_observable(self, func, instance, args, kwargs, decorator_args, decorator_kwargs):
```
collindutter, line 5: Parameter type hints.

griptape/drivers/observability/dummy_observability_driver.py
```diff
@@ -0,0 +1,6 @@
...
+        return func(*args, **kwargs)
```
collindutter, line 6: Should raise a `DummyException` like the other Drivers.

griptape/drivers/observability/open_telemetry_observability_driver.py
```diff
@@ -0,0 +1,67 @@
...
+    def __attrs_post_init__(self):
+        self.trace_provider.add_span_processor(self.span_processor)
+        self._tracer = trace.get_tracer(self.service_name, tracer_provider=self.trace_provider)
+
+    def __enter__(self) -> None:
+        ThreadingInstrumentor().instrument()
+        self._root_span_context_manager = self._tracer.start_as_current_span("main")  # pyright: ignore[reportCallIssue]
+        self._root_span_context_manager.__enter__()
+
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
```
collindutter, line 34: Parameter type hints.

griptape/observability/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .observability import Observability
+
+__all__ = ["Observability"]
```
collindutter, line 3: Wondering if this should live in `griptape.utils` instead of introducing a new module. What do you think?

griptape/observability/observability.py
```diff
@@ -0,0 +1,39 @@
...
+from griptape.drivers.observability.base_observability_driver import BaseObservabilityDriver
+from griptape.drivers.observability.dummy_observability_driver import DummyObservabilityDriver
+
+_dummy_observability_driver = DummyObservabilityDriver()
+_global_observability_driver: Optional[BaseObservabilityDriver] = None
+
+
+@define
+class Observability:
+    driver: BaseObservabilityDriver = field(kw_only=True)
```
collindutter, line 12: Rename `driver` to `observability_driver` for consistency with other uses of Drivers.

griptape/observability/observability.py
```diff
@@ -0,0 +1,39 @@
...
+        global _global_observability_driver
+        return _global_observability_driver
+
+    @staticmethod
+    def set_global_driver(driver: Optional[BaseObservabilityDriver]):
+        global _global_observability_driver
+        _global_observability_driver = driver
+
+    @staticmethod
+    def invoke_observable(func, instance, args, kwargs, decorator_args, decorator_kwargs):
```
collindutter, line 25: Missing parameter type hints.

griptape/observability/observability.py
```diff
@@ -0,0 +1,39 @@
...
+        return driver.invoke_observable(func, instance, args, kwargs, decorator_args, decorator_kwargs)
+
+    def __enter__(self):
+        if Observability.get_global_driver() is not None:
+            raise ValueError("Observability driver already set.")
+        Observability.set_global_driver(self.driver)
+        self.driver.__enter__()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
```
collindutter, line 36: Missing parameter type hints.

CHANGELOG.md
```diff
@@ -22,6 +22,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `BaseTask.parents_outputs` to get the textual output of all parent tasks. 
 - `BaseTask.parents_output_text` to get a concatenated string of all parent tasks' outputs.
 - `parents_output_text` to Workflow context.
+- `BaseObservabilityDriver` as the base class for all observability drivers.
+- `DummyObservabilityDriver` as a no-op observability driver.
+- `OpenTelemetryObservabilityDriver` for sending observability data to an open telemetry collector or vendor.
+- `GriptapeCloudObservabilityDriver` for sending observability data to Griptape Cloud.
+- `Observability` for enabling observability and configuring which observability driver to use.
```
collindutter, line 29: Might be worth pointing out that this is a context manager.

collindutter (changes requested):
Fantastic work! Love that you were able stay within the Driver pattern, and keep the dependencies optional!

docs/griptape-framework/drivers/observability-drivers.md
```diff
@@ -0,0 +1,186 @@
...
+* `__enter__()` sets up the driver.
+* `__exit__()` tears down the driver.
+* `invoke_observable()` wraps all functions and methods marked with the `@observable` decorator. At a bare minimum, implementations invoke the wrapped function and return its result (a no-op). This enables drivers to generate telemetry related to the invocation's call arguments, return values, exceptions, latency, etc.
+
+## Griptape Cloud
+
+The Griptape Cloud Observability Driver instruments `@observable` functions and methods with metrics and traces for use with the Griptape Cloud.
+
+!!! note
+    This Driver is required when using the Griptape Cloud Managed Structures feature. For local development, you can use the [Skatepark Emulator](https://github.com/griptape-ai/griptape-cli?tab=readme-ov-file#skatepark-emulator).
```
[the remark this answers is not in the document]
dylanholmes: Oh my, you caught me, I skimmed and copy/pasted. It is not required.

griptape/observability/__init__.py
dylanholmes: We totally could, especially considering the amount of code in here is trivial. I originally added a new top level module because it felt like a top level feature (like memory) that could use the attention. None of the observability drivers are intended to be used without this module, and so it really is the primary entry point for the feature. That said, I don't really have much of a strong opinion.

Are you still on the fence?

griptape/drivers/observability/dummy_observability_driver.py
dylanholmes: We don't actually need a `DummyObservabilityDriver`, that's just me abusing the conventions.

This was actually intended as a `NoOpObservabilityDriver`, for (1) use in unit tests and (2) illustrate a bare minimum example of subclassing `BaseObservabilityDriver`. If you don't think (2) is important, I can move this into a unit test utility.

griptape/drivers/observability/base_observability_driver.py
dylanholmes: I agree, I'm not really a fan. I could use kwargs or a dataclass (e.g. `ObservableInvocation` or `ObservableCall`) instead. I know those options aren't very common in the framework, but I am a fan of readability. You wanna see how it looks?

griptape/drivers/observability/base_observability_driver.py
dylanholmes: They are for subclasses to perform startup/shutdown type activities (as is needed by open telemetry). They need to be defined in the base class because the `Observability` context manager needs to invoke them.

I figured it isn't always going to be necessary, so it seems nice to have the default implementation (no-op) be implemented in the base class rather than requiring subclasses to implement a no-op (which would be needed if marking them as `abstractmethod`).

I don't expect there to be many subclasses, so it's getting close to bikeshedding. Want me to change it?

docs/griptape-framework/structures/observability.md
dylanholmes: `Observability.observability_driver` is a required parameter for the `Observability` context manager, so an exception will be raised.

Maybe I should reword to not mention the driver at all. Something like: `To opt in, wrap your application code with the `Observability` context manager, for example:`

Does that raise less questions?

griptape/utils/observability/observable.py
dylanholmes: `wrapt` is removed!

pyproject.toml
dylanholmes: Done

CHANGELOG.md
```diff
@@ -22,6 +22,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `@observable` for selecting which functions/methods to provide observability for.
```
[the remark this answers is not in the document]
dylanholmes: ✅

docs/griptape-framework/drivers/observability-drivers.md
```diff
@@ -0,0 +1,186 @@
+# Observability Drivers
+
+Observability Drivers are used by [Observability](../structures/observability.md) to send telemetry (metrics and traces) related to the execution of an LLM application. The telemetry can be used to monitor the application and to diagnose and troubleshoot issues. All Observability Drivers implement the following methods:
+
+* `__enter__()` sets up the driver.
+* `__exit__()` tears down the driver.
+* `invoke_observable()` wraps all functions and methods marked with the `@observable` decorator. At a bare minimum, implementations invoke the wrapped function and return its result (a no-op). This enables drivers to generate telemetry related to the invocation's call arguments, return values, exceptions, latency, etc.
```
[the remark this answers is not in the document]
dylanholmes: ✅

docs/griptape-framework/drivers/observability-drivers.md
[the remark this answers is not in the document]
dylanholmes: ✅

docs/griptape-framework/structures/observability.md
dylanholmes: Changed it as proposed

dylanholmes: Removed reference to ObservabilityDriver, so no longer applicable.

docs/griptape-framework/structures/observability.md
dylanholmes: Added real code

docs/griptape-framework/structures/observability.md
dylanholmes: Done!

docs/griptape-framework/structures/observability.md
dylanholmes: Done!

griptape/drivers/observability/base_observability_driver.py
dylanholmes: Done!

griptape/drivers/observability/base_observability_driver.py
dylanholmes: Added!

griptape/drivers/observability/dummy_observability_driver.py
dylanholmes: Added!

griptape/drivers/observability/dummy_observability_driver.py
dylanholmes: I forgot one other reason for this, it's the default implementation of the `@observable` decorator, which applies when not using the observability feature.

Since this is not just for illustration + unit testing, I'll leave the class where it is, but I'll rename it to `NoOpObservabilityDriver` as to not abuse conventions.

Leaving this comment unresolved in case you have follow up comments or questions.

griptape/drivers/observability/open_telemetry_observability_driver.py
dylanholmes: Added!

griptape/observability/observability.py
dylanholmes: Renamed!

griptape/observability/observability.py
dylanholmes: Added!

CHANGELOG.md
dylanholmes: Pointed out!

griptape/observability/observability.py
dylanholmes: Added!

docs/griptape-framework/structures/observability.md
dylanholmes: Included!

docs/griptape-framework/structures/observability.md
dylanholmes: Capitalized!

griptape/drivers/observability/base_observability_driver.py
collindutter: Makes sense. If we're keeping it as non-abstract then maybe `pass` _is_ correct over `...`.

griptape/drivers/observability/base_observability_driver.py
collindutter: Yeah let's create an inner data class (still using attrs) similar to how we do [here](https://github.com/griptape-ai/griptape/blob/0f196f5b47e8bb8308b4d7feba7a44f6c78d6db4/griptape/utils/prompt_stack.py#L15).

griptape/drivers/observability/dummy_observability_driver.py
collindutter: If it was just for unit-testing purposes, i'd say move to `tests.mocks`. If it's the default implementation can we just move the logic to the base class and get rid of this entirely?

griptape/observability/__init__.py
collindutter: Hmm that's a good point. I'd like to be conscious of having one-off modules, but I also don't want to bloat `griptape.utils` What do you think @vasinov?

vasinov: I agree with Dylan's reasoning here. It feels like it should be its own thing, particularly because it has its own drivers.

griptape/utils/decorators.py
```diff
@@ -26,3 +30,49 @@ def wrapper(self, *args, **kwargs):
...
+
+
+T = TypeVar("T", bound=Callable)
+
+
+def observable(*args: T | Any, **kwargs: Any) -> T:
+    return cast(T, Observable(*args, **kwargs))
+
+
+class Observable:
```
vasinov, line 42: I would pull this out into a separate file under `griptape.observability` for consistency.

vasinov (changes requested):
Looks great overall!

griptape/drivers/observability/base_observability_driver.py
dylanholmes: Added `Observable.Call`

griptape/utils/decorators.py
dylanholmes: I moved the decorator into a separate file under `griptape.common`. Moving it under `griptape.observability` introduced a particularly difficult-to-mitigate circular dependency. I chose `griptape.common` over `griptape.utils` since it is being introduced in https://github.com/griptape-ai/griptape/pull/861 and seems like a better place for common "core" things.

vasinov (approved):

PR #853, opened 2024-06-11
Fix `ActionSubtask` output value

vachillo:


collindutter:
Can you please update the change log's `Fixed` section with `Raw Tool output being lost when being executed by ActionsSubtask.`

collindutter (approved):

PR #854, opened 2024-06-11
Fix `WebSearch.search` output

vachillo:
- fixed `TextArtifacts` contained in `ListArtifact` returned by `WebSearch.search` to properly formatted stringified JSON


griptape/tools/web_search/tool.py
```diff
@@ -32,7 +33,7 @@ def search(self, props: dict) -> ListArtifact | ErrorArtifact:
         query = props["values"]["query"]
 
         try:
-            return ListArtifact([TextArtifact(str(result)) for result in self._search_google(query)])
+            return ListArtifact([TextArtifact(json.dumps(result)) for result in self._search_google(query)])
```
dylanholmes, line 36: Would you be able to give an example of a problematic scenario that this fixes?

vachillo: currently i am trying to get a list of URLs. the tool returns a string version of some JSON that contains `url` as a key. i cannot get access to the `url` key unless i can `loads` the json output from this tool.

collindutter (approved):

collindutter (dismissed):

collindutter (approved):

PR #856, opened 2024-06-12
Fix missing parent error handling in Workflow task insertion

cjkindel:
Added tests for missing child & parent too


tests/unit/structures/test_workflow.py
```diff
@@ -331,6 +331,32 @@ def test_run_topology_1_imperative_insert(self):
...
+    def test_run_topology_1_missing_child(self):
+        task1 = PromptTask("test1", id="task1")
+        task2 = PromptTask("test2", id="task2")
+        task3 = PromptTask("test3", id="task3")
+        task4 = PromptTask("test4", id="task4")
+        workflow = Workflow(prompt_driver=MockPromptDriver())
+
+        # task4 never added to workflow
+        workflow + task1
+        workflow.insert_tasks(task1, [task2, task3], task4)
```
dylanholmes, line 355: Might be nice if `insert_tasks` handled errors the same way for missing parents and children. (Just pick between `Workflow.insert_tasks()` and  `Workflow.run()`).

A natural place to validate might be `Structure.resolve_relationships()` (kind of nice because its also being used by the `StructureVisualizer` too). We seem to be leaning towards messy creation, then validate before the run, so I might lean toward that option.

Another option could be to update `insert_tasks` to just add parents/children to the workflow if they aren't already. Actually, that's not necessarily orthogonal, it could be done in addition to the validation.

What do you think?

dylanholmes (commented):
Nice work!

tests/unit/structures/test_workflow.py
cjkindel: Validating children in the same way sounds good but it may not be possible as they aren't likely to already be in self.tasks since they are added after the parents are added (chicken/egg). I believe resolve_relationships will already [catch an error like this](https://github.com/griptape-ai/griptape/blob/b634598f8b10783c81ec514738ddfd1355419e07/griptape/structures/structure.py#L219) but that only catches in [before_run](https://github.com/griptape-ai/griptape/blob/b634598f8b10783c81ec514738ddfd1355419e07/griptape/structures/structure.py#L239) not on task insertion.

dylanholmes: Can we update `insert_tasks` to just add parents/children to the workflow if they aren't already?

cjkindel: I'd rather follow the Python pattern "easier to ask forgiveness than permission", it you try to insert a task in an invalid way, we raise an exception. Instead of making an assumption on behalf of the author

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Performance issue in `OpenAiChatPromptDriver` when extracting unused rate-limiting headers.
 - Streaming not working when using deprecated `Structure.stream` field.
 - Raw Tool output being lost when being executed by ActionsSubtask.
+- Enumerate Structure parent tasks and handle non-existent parent.
```
collindutter, line 63: We should specify that this fix was done in `Workflow.insert_tasks`. And were were already enumerating parent tasks, but we were using `index` which was using object equality as opposed to id equality.

tests/unit/structures/test_workflow.py
collindutter, line 1: This fix came from `self.tasks.index` using object equality instead of id equality. Meaning this would break:

```python
        task1 = PromptTask("test1", id="task1")
        task2 = PromptTask("test2", id="task2")
        task3 = PromptTask("test3", id="task3")
        workflow = Workflow(prompt_driver=MockPromptDriver())

        # task1 never added to workflow
        workflow + task4
        with pytest.raises(ValueError):
            workflow.insert_tasks(PromptTask("test1", id="task1"), [task2], task4)
```

But I'm actually not sure what the intended behavior is here...

collindutter: @vasinov @andrewfrench what do you think about this? Are two tasks with the same id equal? If so, maybe this should be fixed at the Task level with an `__eq__` override.

dylanholmes: I think this example should raise an exception to avoid edge cases. If you accidentally create two tasks with the same id, with different relationships, then allowing this could get quite confusing.

tests/unit/structures/test_workflow.py
dylanholmes: My comment wasn't motivated by catching the error, but about convenient user experience.

It doesn't make sense to insert tasks that aren't part of a workflow, so it's not that weird to guess that the parent and child tasks should be added to the workflow.
```python
workflow.add_task(a)
workflow.add_task(d)
workflow.insert_tasks(a, [b, c], d)
```
is unnecessarily verbose. You will never want to supply a parent or child that is not part of the workflow, not to mention the whole point of insert_tasks is convenience anyway.

It's be much nicer to just write:
```python
workflow.insert_tasks(a, [b, c], d)
```

That said, such a change should probably go in a separate PR anyway.

collindutter: @dylanholmes would you mind creating an issue for this?

tests/unit/structures/test_workflow.py
collindutter: @cjkindel can you please revert the logic to use `index` and raise an exception if it's not found? Also please add a test for the example above.

collindutter (approved):
Did not know about `try/else`, neat.

PR #857, opened 2024-06-12
Re-order Workflow tasks on every task execution wave

cjkindel:


griptape/structures/workflow.py
```diff
@@ -92,11 +92,11 @@ def insert_task(
 
     def try_run(self, *args) -> Workflow:
         self._execution_args = args
-        ordered_tasks = self.order_tasks()
         exit_loop = False
 
         while not self.is_finished() and not exit_loop:
             futures_list = {}
+            ordered_tasks = self.order_tasks()
```
dylanholmes, line 99: Just curious, what problem did you encounter?

collindutter: Adding Tasks to the Workflow while it's running. It's less about re-ordering the Tasks on each wave and more about re-fetching them.

cjkindel: This is actually @vachillo's finding, so I don't have the code. But I understand it was dynamic task/DAG changes during a workflow's execution that necessitates the task order being re-indexed. They may have more details

dylanholmes: sounds exciting

collindutter (approved):

PR #859, opened 2024-06-12
Updated `GriptapeCloudKnowledgeBaseClient.query()` to optionally return raw query responses from the Knowledge Base

cjkindel:


griptape/tools/griptape_cloud_knowledge_base_client/tool.py
```diff
@@ -40,6 +40,24 @@ def query(self, params: dict) -> TextArtifact | ErrorArtifact:
...
+            response_body = response.json()
+
+            artifacts: list[BaseArtifact] = []
+            for query_result in response_body.get("result", []):
+                artifacts.append(BaseArtifact.from_json(query_result.meta["artifact"]))
+
+            return ListArtifact(artifacts)
+        except exceptions.RequestException as err:
+            return ErrorArtifact(str(err))
+
```
collindutter, line 60: This should be part of the schema for `query`, not a new Activity.

cjkindel: Will remain a draft until internal KnowledgeBase API conversation concludes. But updated to be part of `query` for now

collindutter:
@cjkindel are you planning on doing the Vector Store Driver work on this PR? Otherwise I think we should close since there should be no changes to `GriptapeCloudKnowledgeBaseClient`

PR #861, opened 2024-06-13
Refactor prompt stack

collindutter:


griptape/common/prompt_stack/contents/text_prompt_stack_content.py
```diff
@@ -0,0 +1,20 @@
...
+from griptape.common import BasePromptStackContent, BaseDeltaPromptStackContent, TextDeltaPromptStackContent
+
+
+@define
+class TextPromptStackContent(BasePromptStackContent):
+    artifact: TextArtifact = field(metadata={"serializable": True})
+
+    @classmethod
+    def from_deltas(cls, deltas: Sequence[BaseDeltaPromptStackContent]) -> TextPromptStackContent:
+        text_deltas = [delta for delta in deltas if isinstance(delta, TextDeltaPromptStackContent)]
```
vachillo, line 16: should this silently fail if its not the right type?

griptape/common/prompt_stack/messages/prompt_stack_message.py
```diff
@@ -0,0 +1,41 @@
...
+
+from .base_prompt_stack_message import BasePromptStackMessage
+
+
+@define
+class PromptStackMessage(BasePromptStackMessage):
+    def __init__(self, content: str | list[BasePromptStackContent], **kwargs: Any):
+        if isinstance(content, str):
+            content = [TextPromptStackContent(TextArtifact(value=content))]
+        self.__attrs_init__(content, **kwargs)  # pyright: ignore[reportAttributeAccessIssue]
```
vachillo, line 18: could use an attrs `converter` here

griptape/drivers/prompt/huggingface_pipeline_prompt_driver.py
```diff
@@ -40,43 +48,49 @@ class HuggingFacePipelinePromptDriver(BasePromptDriver):
...
     def prompt_stack_to_string(self, prompt_stack: PromptStack) -> str:
         return self.tokenizer.tokenizer.decode(self.__prompt_stack_to_tokens(prompt_stack))
 
-    def _prompt_stack_input_to_message(self, prompt_input: PromptStack.Input) -> dict:
-        return {"role": prompt_input.role, "content": prompt_input.content}
+    def _prompt_stack_to_messages(self, prompt_stack: PromptStack) -> list[dict]:
+        messages = []
+        for i in prompt_stack.messages:
+            if len(i.content) == 1:
+                messages.append({"role": i.role, "content": TextPromptStackContent(i.to_text_artifact())})
```
dylanholmes, line 85: Can you replace `TextPromptStackContent(i.to_text_artifact())` with just `i` (or equivalently `i.content`)?

I understand that this works because of `BasePromptStackContent. __str__`, but is that equivalent to just `i`, since  `PromptStackMessage` also has `__str__` (which calls `self.to_text_artifact().to_text()`)?

dylanholmes (commented):
Love it

griptape/common/prompt_stack/messages/delta_prompt_stack_message.py
```diff
@@ -0,0 +1,15 @@
...
+from attrs import define, field
+
+from griptape.common.prompt_stack.contents.text_delta_prompt_stack_content import TextDeltaPromptStackContent
+
+
+from .base_prompt_stack_message import BasePromptStackMessage
+
+
+@define
+class DeltaPromptStackMessage(BasePromptStackMessage):
```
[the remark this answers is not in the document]
collindutter: Yeah I went back and forth on this. I didn't love putting the `Delta`/`Chunk`/etc at the end because it breaks the naming convention of its parent class. `PartialPromptStackMessage` works but is inconsistent with the usage of `Delta` in the Prompt Stack Content classes. So I could update those to `Partial`, but `Delta` feels better there.

What do you think?

griptape/common/prompt_stack/messages/prompt_stack_message.py
```diff
@@ -0,0 +1,41 @@
...
+
+    content: list[BasePromptStackContent] = field(metadata={"serializable": True})
```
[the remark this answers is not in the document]
collindutter: Went back and forth on this a million times. In this PR, yeah it's not much more. But the upcoming function calling PR, for instance, has an `ActionResultPromptStackContent` that contains an artifact _and_ some metadata about the action.

Additionally, to quote something @vasinov said:
> I think the reason we wanted to keep it separate is because artifacts are a pretty clean abstraction for data passing. In the PromptStack inputs really are elements that currently somewhat match artifacts but we don't know if that's going to hold true in the future. It would also be much easier to add API-specific elements if an API requires it.

griptape/common/prompt_stack/messages/base_prompt_stack_message.py
```diff
@@ -0,0 +1,46 @@
...
+from griptape.common import BasePromptStackContent, BaseDeltaPromptStackContent
+from griptape.mixins import SerializableMixin
+
+
+@define
+class BasePromptStackMessage(ABC, SerializableMixin):
+    @define
+    class Usage(SerializableMixin):
+        input_tokens: Optional[float] = field(kw_only=True, default=None, metadata={"serializable": True})
+        output_tokens: Optional[float] = field(kw_only=True, default=None, metadata={"serializable": True})
```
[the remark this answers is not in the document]
collindutter: One of the prompt drivers (I can't remember which one) returned the token usage as floats. They were still truncated to integers, but it seems like partial tokens could be a possibility in the future.

griptape/drivers/prompt/huggingface_pipeline_prompt_driver.py
collindutter: Cleaned up a bit, let me know what you think.

griptape/common/prompt_stack/contents/text_prompt_stack_content.py
collindutter: Not sure I follow -- this will not silently fail as it currently is.

griptape/common/prompt_stack/messages/prompt_stack_message.py
collindutter: I don't think there's a good way to use a converter and get proper type hints. With a converter, you'll get yelled at for trying to pass in a string. And updating the field's type hints leads down a path of `Union`s everywhere.

dylanholmes (changes requested):
Minor comments and a couple of topics for discussion

griptape/common/prompt_stack/prompt_stack.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import annotations
+from attrs import define, field
+
+from griptape.artifacts import TextArtifact, BaseArtifact, ListArtifact, ImageArtifact
+from griptape.mixins import SerializableMixin
+from griptape.common import PromptStackMessage, TextPromptStackContent, BasePromptStackContent, ImagePromptStackContent
+
+
+@define
+class PromptStack(SerializableMixin):
```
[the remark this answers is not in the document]
collindutter: I like the shortening of Content/Message. Also in favor of renaming Prompt Stack since it's grown to be much more than just a list of string prompts. Let's try `Messages` and see how it sits.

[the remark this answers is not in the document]
dylanholmes: Why `Stack`?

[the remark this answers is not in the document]
dylanholmes: I was trying to understand the origins of `Stack` in the context of a prompt stack. I found this article: [Unveiling the Mastery of Prompt Stacking](https://www.linkedin.com/pulse/unveiling-mastery-prompt-stacking-your-roadmap-precision-mike-meehan), which defines Prompt Stacking as "meticulously crafting a series of prompts to guide the model step by step towards a desired outcome." with the following basic example:
- Basic Prompt: “Translate the following text to French: {text}”
- Stacked Prompt: “You are a proficient bilingual assistant. Translate the following text to French: {text}”

So it sounds like prompt stacking is more or less synonymous with prompt engineering. 

This makes me realize that stack is definitely a common terminology wrt LLMs, but that it doesn't seem to directly align with our `PromptStack`, since it would probably be tacking place within our prompt templates (I suppose an individual "prompt" roughly equals a statement or instruction or rule within the template). `MessageStack` does work, since it is "stacking" messages just as much as "stacking" prompts does, though the computer science pupil in me is screaming "but it's not a stack!" (it's ok we can just ignore him).

So long story short, I'm fine with `MessageStack`, though I'd still prefer something that no one would ever need to ask "why" about.

docs/griptape-framework/misc/events.md
```diff
@@ -244,9 +244,9 @@ from griptape.events import BaseEvent, StartPromptEvent, EventListener
 
 def handler(event: BaseEvent):
     if isinstance(event, StartPromptEvent):
-        print("Prompt Stack Inputs:")
-        for input in event.prompt_stack.inputs:
-            print(f"{input.role}: {input.content}")
+        print("Message Stack PromptStack:")
```
dylanholmes, line 247: "Prompt Stack Messages:"

docs/griptape-framework/misc/events.md
```diff
@@ -259,7 +259,7 @@ agent.run("Write me a poem.")
 ```
 ```
 ...
-Prompt Stack Inputs:
+Message Stack Messages:
```
dylanholmes, line 262: "Prompt Stack Messages"

dylanholmes (changes requested):

docs/griptape-framework/misc/events.md
collindutter: Done

docs/griptape-framework/misc/events.md
collindutter: Done

dylanholmes (approved):
💯

dylanholmes (approved):
🥮

PR #862, opened 2024-06-13
Add structure property `output`

vachillo:


griptape/structures/structure.py
```diff
@@ -127,6 +127,10 @@ def input_task(self) -> Optional[BaseTask]:
     def output_task(self) -> Optional[BaseTask]:
         return self.tasks[-1] if self.tasks else None
 
+    @property
+    def output(self) -> Optional[BaseArtifact]:
+        return self.output_task.output if self.output_task else None
```
collindutter, line 132: Use explicit `is not None` check

collindutter (commented):
Looks good, just changelog pls. Document both the addition of `Structure.output`, and the fix on conversation memory.

collindutter (approved):

PR #864, opened 2024-06-13
Set allowlist in certain file manager examples

collindutter:
`gpt-4o` appears to get confused when using the `FileManager`, and is occasionally hallucinating invalid json. Until native function calling is ready, we're updating the examples to remove all non-relevant Activities from `FileManager`

----
📚 Documentation preview 📚: https://griptape--864.org.readthedocs.build//864/


vasinov (approved):

collindutter:
Should be resolved with native function calling.

PR #879, opened 2024-06-19
Don't include "v" in tag name

collindutter:
Allows RTD to [automatically generate](https://docs.readthedocs.io/en/stable/versions.html#versions) a `stable` version from semvar tag.


vachillo (approved):
v good

zachgiordano (approved):

PR #882, opened 2024-06-20
Update `AmazonBedrockStructureConfig` with helper params

vachillo:


griptape/config/amazon_bedrock_structure_config.py
```diff
@@ -8,49 +13,78 @@
...
 class AmazonBedrockStructureConfig(StructureConfig):
+    region: Optional[str] = field(
+        default=Factory(lambda: os.environ.get("AWS_DEFAULT_REGION")), metadata={"serializable": True}, kw_only=True
+    )
+    access_key_id: Optional[str] = field(
+        default=Factory(lambda: os.environ.get("AWS_ACCESS_KEY_ID")), metadata={"serializable": False}, kw_only=True
+    )
+    secret_access_key: Optional[str] = field(
+        default=Factory(lambda: os.environ.get("AWS_SECRET_ACCESS_KEY")), metadata={"serializable": False}, kw_only=True
+    )
```
collindutter, line 36: I don't think we need these. All Drivers that use boto only accept a `session`. Users can configure that however they like (region, keys, role, etc).

collindutter (approved):

collindutter (approved):

PR #891, opened 2024-06-24
`meta` parameter added to TextArtifact

cjkindel:


griptape/artifacts/text_artifact.py
```diff
@@ -13,6 +13,7 @@ class TextArtifact(BaseArtifact):
     value: str = field(converter=str, metadata={"serializable": True})
     encoding: str = field(default="utf-8", kw_only=True)
     encoding_error_handler: str = field(default="strict", kw_only=True)
+    meta: dict[str, str] = field(factory=dict, kw_only=True, metadata={"serializable": True})
```
vasinov, line 16: Is there a reason for values to be so restrictive? Should it be an `Any` instead?

cjkindel: Agreed, updated

vasinov (approved):

PR #900, opened 2024-06-25
Bug Fix: Cohere prompts with no history

emjay07:
- Updating the CoherePromptDriver to properly account for empty history
- Updating the docs to `command-r` model, which is recommended for Cohere

----
📚 Documentation preview 📚: https://griptape--900.org.readthedocs.build//900/

### DIFF
--- a/griptape/drivers/prompt/cohere_prompt_driver.py
+++ b/griptape/drivers/prompt/cohere_prompt_driver.py
@@ -52,12 +52,18 @@ class CoherePromptDriver(BasePromptDriver):
     def _base_params(self, prompt_stack: PromptStack) -> dict:
         user_message = prompt_stack.inputs[-1].content
 
-        history_messages = [self._prompt_stack_input_to_message(input) for input in prompt_stack.inputs[:-1]]
+        history_messages = [
+            self._prompt_stack_input_to_message(input) for input in prompt_stack.inputs[:-1] if input.content
+        ]
 
-        return {
+        params = {
             "message": user_message,
-            "chat_history": history_messages,
             "temperature": self.temperature,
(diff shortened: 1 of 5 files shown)


griptape/drivers/prompt/cohere_prompt_driver.py
```diff
@@ -52,12 +52,18 @@ def _prompt_stack_input_to_message(self, prompt_input: PromptStack.Input) -> dic
     def _base_params(self, prompt_stack: PromptStack) -> dict:
         user_message = prompt_stack.inputs[-1].content
 
-        history_messages = [self._prompt_stack_input_to_message(input) for input in prompt_stack.inputs[:-1]]
+        history_messages = [
+            self._prompt_stack_input_to_message(input) for input in prompt_stack.inputs[:-1] if input.content
```
vasinov, line 56: Can you compare to `None` explicitly with `input.content is not None`?

vasinov (changes requested):

griptape/drivers/prompt/cohere_prompt_driver.py
emjay07: Sadly, no. it's a problem with it being an empty string. So I can check for None or an empty string, but that's what the boolean check is doing.

```
'chat_history': [{'role': 'SYSTEM', 'text': ''}]}
```

vasinov: Ah, gotcha, nevermind then.

vasinov (approved):

vasinov (approved):

vasinov (approved):

Issue #904, opened 2024-06-26
Add support for branching in `Workflow`s


vachillo:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
I am trying to build a workflow where I want to conditionally run different `Task`s based on the parent's output.

**Describe the solution you'd like**
I want to be able to conditionally branch to different `Task`s depending on the output of the parent.

**Describe alternatives you've considered**

**Additional context**

PR #905, opened 2024-06-26
Add `BooleanArtifact`

vachillo:
artifact that is explicitly `True` or `False`.

i want to output a boolean from a `Task` and do something with that information later, and checking the string value of a `TextArtifact` is clunky.

incremental update to support #904

### DIFF
--- a/griptape/artifacts/boolean_artifact.py
+++ b/griptape/artifacts/boolean_artifact.py
@@ -1,0 +1,25 @@
+from __future__ import annotations
+from typing import Any
+from attrs import define, field
+from griptape.artifacts import BaseArtifact
+
+
+@define
+class BooleanArtifact(BaseArtifact):
+    value: bool = field(converter=bool, metadata={"serializable": True})
+    meta: dict[str, Any] = field(factory=dict, kw_only=True, metadata={"serializable": True})
+
+    @classmethod
+    def parse_bool(cls, value: str) -> BooleanArtifact:
+        """
+        Convert a string literal to a BooleanArtifact. The string must be either "true" or "false" with any casing.
+        """
+        if value is not None:
+            if value.lower() == "true":
+                return BooleanArtifact(True)
+            elif value.lower() == "false":
+                return BooleanArtifact(False)
+        raise ValueError(f"Cannot convert string literal '{value}' to BooleanArtifact")
+
+    def __add__(self, other: BaseArtifact) -> BooleanArtifact:
+        raise ValueError("Cannot add BooleanArtifact with other artifacts")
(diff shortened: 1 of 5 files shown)


griptape/artifacts/boolean_artifact.py
```diff
@@ -0,0 +1,25 @@
...
+    def parse_bool(cls, value: str) -> BooleanArtifact:
+        """
+        Convert a string literal to a BooleanArtifact. The string must be either "true" or "false" with any casing.
+        """
+        if value is not None:
+            if value.lower() == "true":
+                return BooleanArtifact(True)
+            elif value.lower() == "false":
+                return BooleanArtifact(False)
+        raise ValueError(f"Cannot convert string literal '{value}' to BooleanArtifact")
```
vachillo, line 22: should this return `None` or raise an exception?

griptape/artifacts/boolean_artifact.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import annotations
+from typing import Any
+from attrs import define, field
+from griptape.artifacts import BaseArtifact
+
+
+@define
+class BooleanArtifact(BaseArtifact):
+    value: bool = field(converter=bool, metadata={"serializable": True})
```
vachillo, line 9: passing in a value on init will get coerced into a `bool`, whereas using `parse_bool` will check for the literal true or false. should that check be done on init instead?

griptape/artifacts/boolean_artifact.py
dylanholmes: I vote for raise an exception.

If the user's code assumes that the parse always succeeds and then it fails, returning None will (most likely) cause an exception to be raised by code that consumes this output. Then it will be the user's responsibility to trace it back to here, so we might as well do it for them.

griptape/artifacts/boolean_artifact.py
dylanholmes: I think having these two methods of creating a `BooleanArtifact` seems reasonable. The default (init) way being a little more flexible, but still offering a stricter option for strings specifically.

If you don't offer the bool coercion option and the use case comes up, then what would that look like? like this: `BooleanArtifact(bool(x))`? I suppose that's easy enough, but `BooleanArtifact(x)` is a little better. Actually, the `parse_bool` wouldn't work for a literal `True` or `False`, so I definitely vote for keeping both creation options.

dylanholmes (changes requested):
Nice work!

I know there's not much to test, but wanna add unit tests?

griptape/artifacts/boolean_artifact.py
vachillo: the init method will always convert the input to a boolean. this can lead to suprising behavior sometimes where `BooleanArtifact("False").value is True`. `parse_bool` is to parse string literal values as a convenience. thinking about it now, seems more "pythonic" to keep the init as to whatever python interprets as a bool, and use `parse_bool` to parse string literal values as a convenience

dylanholmes (approved):

Issue #872, opened 2024-06-17
Allow passing k/v pairs to a custom J2 template


vachillo:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
I have custom J2 templates that i would like to use, but i have no way to pass custom k/v pairs without overriding the entire class

**Describe the solution you'd like**
I want to pass a dict to a Tool/Engine that use J2 templates that i can use to pass in k/v pairs to render to my custom template

**Describe alternatives you've considered**

**Additional context**

collindutter:
This should be implemented through the following:
1. Add `task: self` to `BaseTask.full_context`. 
2. Add `context: self.full_context` as arguments to all Task uses of `J2.render`.

collindutter:
@vachillo I know this is quite old but can you expand on the ask here? Can you provide a code snippet that shows what you're trying to do?

PR #907, opened 2024-06-27
Add context to BaseSummaryEngine

vachillo:


tests/unit/engines/summary/test_prompt_summary_engine.py
```diff
@@ -9,7 +9,7 @@
 class TestPromptSummaryEngine:
```
dylanholmes, line 9: Would be nice if there was at least one test that used the context in the template too (currently added tests seem to just supply the context init param, but never test that it effects the output).

griptape/engines/summary/base_summary_engine.py
```diff
@@ -1,12 +1,14 @@
...
-from attrs import define
+from typing import Any, Optional
+from attrs import define, field
 from griptape.artifacts import TextArtifact, ListArtifact
 from griptape.rules import Ruleset
 
 
 @define
 class BaseSummaryEngine(ABC):
+    context: dict[str, Any] = field(factory=dict, kw_only=True)
```
dylanholmes, line 10: I'm wondering if additional template parameters makes sense in the engine or not. Do you have a specific use case that would make it easier to deduce the requirements on possible solutions? Is the use case dependent on adding the template parameters to the engine? For example, do you need to be able to use the engine with the context outside of a task/structure?

If we do indeed want to supply the additional template parameters in the engine, I think it should at least be named something other than context, to make it less confusing with the existing `context` in [BaseTask.context](https://github.com/griptape-ai/griptape/blob/b93f209e458039fcd546e6dbb90a97869259ec2a/griptape/tasks/base_task.py#L162). This existing `context` gets consumed (though indirectly) by [TextSummaryTask](https://github.com/griptape-ai/griptape/blob/b93f209e458039fcd546e6dbb90a97869259ec2a/griptape/tasks/text_summary_task.py#L30C69-L30C74) (via [BaseTask.full_context](https://github.com/griptape-ai/griptape/blob/b93f209e458039fcd546e6dbb90a97869259ec2a/griptape/tasks/base_task.py#L162) and [BaseTextInputTask.input](https://github.com/griptape-ai/griptape/blob/b93f209e458039fcd546e6dbb90a97869259ec2a/griptape/tasks/base_text_input_task.py#L29)).

collindutter: What if we added a `context` field to the J2 class instead?

vachillo: it is meant to mimic the init parameter on [BaseTask](https://github.com/griptape-ai/griptape/blob/b93f209e458039fcd546e6dbb90a97869259ec2a/griptape/tasks/base_task.py#L35). yes the use case is custom template rendering. right now in the summary engine (or really _any_ class that uses templates), it is really hard to know what parameters are accessible when overriding a template. see #872. 

all of the existing uses of `full_context` are only for input rendering (which this class does not have), not system/user prompt rendering. i can follow the same `context` init and `full_context` property pattern if that would be better here?

collindutter: Is this PR intended to solve #872?

> context field on J2 might make sense too, but that would be a much bigger change.

Would it be a big change though? The change would be isolated to `J2`, and it would benefit all uses of `J2` rather than just this Engine.

vachillo:
closing to reimplement on J2

PR #909, opened 2024-06-27
fixing bug for upserting csv artifacts

emjay07:
- `CSVRowArtifact` inherits from `TextArtifact` and therefore needs to be supported in `upsert_text_artifact`
- This converts the value to a string if it's not a string before hashing

### DIFF
--- a/griptape/drivers/vector/base_vector_store_driver.py
+++ b/griptape/drivers/vector/base_vector_store_driver.py
@@ -54,7 +54,11 @@ class BaseVectorStoreDriver(SerializableMixin, ABC):
         **kwargs,
     ) -> str:
         meta = {} if meta is None else meta
-        vector_id = utils.str_to_hash(artifact.value) if vector_id is None else vector_id
+
+        # Properly handle CSVs
+        artifact_value_string = artifact.value if isinstance(artifact.value, str) else str(artifact.value)
+
+        vector_id = utils.str_to_hash(artifact_value_string) if vector_id is None else vector_id
 
         if self.does_entry_exist(vector_id, namespace):
             return vector_id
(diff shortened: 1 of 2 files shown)


griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -54,7 +54,11 @@ def upsert_text_artifact(
         **kwargs,
     ) -> str:
         meta = {} if meta is None else meta
-        vector_id = utils.str_to_hash(artifact.value) if vector_id is None else vector_id
+
+        # Properly handle CSVs
+        artifact_value_string = artifact.value if isinstance(artifact.value, str) else str(artifact.value)
+
+        vector_id = utils.str_to_hash(artifact_value_string) if vector_id is None else vector_id
```
vachillo, line 61: use `.to_string()` here instead.
```suggestion
        vector_id = utils.str_to_hash(artifact.to_string()) if vector_id is None else vector_id
```

vachillo (changes requested):

griptape/drivers/vector/base_vector_store_driver.py
vasinov: Isn't it `to_text()`?

vachillo: yep my bad

emjay07: Why this change? Python's standard form is to convert to strings with str()? A dict is a built in type and this method will properly handle other types like arrays vs. using a custom method.

vasinov: `BaseArtifact.to_text()` is for handling `value`s in more specific ways. For example, `CsvRowArtifact.to_text()` does proper CSV parsing and generates more friendly string values.

emjay07: gotcha. is there a reason we didn't just override `__str__()`?

will make this fix, but seems weird to not use built-in functionaltiy.

vasinov: Ha, we did in fact:

```python
def __str__(self) -> str:
    return self.to_text()
```

I can't remember the reason we have a separate `to_text()` method. But there was a reason.

vachillo (approved):

PR #913, opened 2024-06-28
Add `ControlFlowTask`

### LINKED ISSUE #904: Add support for branching in `Workflow`s

vachillo:
- add `ControlFlowTask`, `ControlFlowArtifact`, `TaskArtifact`
- add new task state `CANCELLED`
Closes #904


griptape/artifacts/boolean_artifact.py
```diff
@@ -10,11 +10,13 @@ class BooleanArtifact(BaseArtifact):
     meta: dict[str, Any] = field(factory=dict, kw_only=True, metadata={"serializable": True})
 
     @classmethod
-    def parse_bool(cls, value: Union[str, bool]) -> BooleanArtifact:
+    def parse_bool(cls, value: Union[str, bool, TextArtifact]) -> BooleanArtifact:
```
collindutter, line 13: Should be named `from_bool` for consistency with other similar methods.

griptape/artifacts/boolean_artifact.py
```diff
@@ -10,11 +10,13 @@ class BooleanArtifact(BaseArtifact):
...
         """
         Convert a string literal or bool to a BooleanArtifact. The string must be either "true" or "false" with any casing.
```
collindutter, line 15: Update docstring to mention text artifact.

griptape/tasks/base_control_flow_task.py
```diff
@@ -0,0 +1,35 @@
...
+    def after_run(self) -> None:
+        super().after_run()
+
+        self.structure.meta_memory.add_entry(
+            ControlFlowMetaEntry(
+                input_tasks=[parent.id for parent in self.parents],
+                output_tasks=[child.id for child in filter(lambda child: not child.is_finished(), self.children)],
+                output=self.output,
+            )
+        )
```
collindutter, line 24: Why are we adding to Meta Memory at all? What does the LLM need to know about the branch decisions?
Also, there has been talk of removing Meta Memory entirely, I'm not sure we should rely on it.

griptape/artifacts/task_artifact.py
```diff
@@ -0,0 +1,30 @@
...
+from typing import TYPE_CHECKING
+from griptape.artifacts import ControlFlowArtifact
+
+if TYPE_CHECKING:
+    from griptape.tasks import BaseTask
+    from griptape.artifacts import BaseArtifact
+
+
+@define
+class TaskArtifact(ControlFlowArtifact):
```
collindutter, line 12: Not sure how I feel about this artifact. And in a fully declarative Workflow it may be impossible to return a Task object.

Can we just operate with Task ids in a TextArtifact?

griptape/tasks/boolean_control_flow_task.py
```diff
@@ -0,0 +1,45 @@
...
+from attrs import field
+
+from griptape.artifacts import BooleanArtifact
+from griptape.tasks import BaseControlFlowTask
+
+if TYPE_CHECKING:
+    from griptape.tasks import BaseTask
+
+
+class BooleanControlFlowTask(BaseControlFlowTask):
```
collindutter, line 12: I'm not sure we need to reinvent boolean logic in its own Task. Users can implement this themselves in their `control_flow_fn`

griptape/tasks/choice_control_flow_task.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import Callable
+from attrs import define, field
+
+from griptape.artifacts import BaseArtifact, ErrorArtifact, TaskArtifact, ListArtifact
+from griptape.tasks import BaseTask
+from griptape.tasks import BaseControlFlowTask
+
+
+@define
+class ChoiceControlFlowTask(BaseControlFlowTask):
```
collindutter, line 11: Do we need a dedicated Task for this? What if we added `control_flow_fn` to `BaseTask`?

griptape/artifacts/boolean_artifact.py
vachillo: - a bool is one of the possible inputs, so `from_bool` isn't really accurate.
- i wanted to convey the fact that its going to actually evaluate the literal text in the string, rather than just coerce the input type to a bool type.

griptape/tasks/base_control_flow_task.py
vachillo: wanted to have some mechanism of storing the "decision" of the control flow.

griptape/artifacts/task_artifact.py
vachillo: this is just used to return the `input` of the task. and it will always return the tasks because it pulls directly from the structure

griptape/tasks/choice_control_flow_task.py
vachillo: future proof if there is other types of control flow tasks

collindutter:
Slight variation to this implementation: https://github.com/griptape-ai/griptape/compare/dev...feature/branch-task

Issue #643, opened 2024-02-12
Prompt Drivers Crash When Streaming Without A Structure


collindutter:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Describe the bug**
Prompt Drivers require `structure` to be set in order to publish `CompletionChunkEvents`.

**To Reproduce**
1. Create a Prompt Driver without a Structure.
2. Try streaming.
3. Observe a crash [here](https://github.com/griptape-ai/griptape/blob/dev/griptape/drivers/prompt/base_prompt_driver.py#L79).

**Expected behavior**
Prompt Drivers should not depend on a Structure just to stream. Perhaps they should take in an `event_listener: EventListener` instead of a `structure: Structure`. If a Structure is the one using the Prompt Driver, it can provide an `EventListener`. Otherwise we can default one.

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Additional context**
[Add any other context about the problem here.](https://github.com/griptape-ai/griptape/pull/591/files#r1484815191)

torabshaikh:
@collindutter Can I do it?

collindutter:
@torabshaikh sure! Thanks :)

collindutter:
Hey @torabshaikh, just checking in to see if you had any questions on this! Let us know if there's anything we can do to help.

torabshaikh:
@collindutter I have started working on it. I will surely ask you if I have any doubts.

torabshaikh:
Hi @collindutter , I have replaced `structure: Structure` with  `event_listener: EventListener` but there is an issue. `preprocess` method inside `PromptTask` takes a structure. And an structure can have multiple `event_listener`. This is breaking the flow since we can't assign a list of `EventListener` to a single `EventListener` .

collindutter:
@torabshaikh can you make it so the Prompt Driver takes `event_listeners: list[EventListener]` instead?

torabshaikh:
Hi @collindutter I just submitted #914 for this. Please review.

PR #914, opened 2024-06-28
Updated: BasePromptDriver to publish on a list of EventListner instead of structures

torabshaikh:
This PR fixes this - #643

### DIFF
--- a/griptape/drivers/prompt/base_prompt_driver.py
+++ b/griptape/drivers/prompt/base_prompt_driver.py
@@ -11,7 +11,7 @@ from griptape.tokenizers import BaseTokenizer
 from griptape.artifacts import TextArtifact
 
 if TYPE_CHECKING:
-    from griptape.structures import Structure
+    from griptape.events import EventListener
 
 
 @define(kw_only=True)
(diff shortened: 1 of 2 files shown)


collindutter:
Thanks for the PR @torabshaikh! I do think it's a good idea to decouple Prompt Drivers from Structures, but I don't love having to iterate over `EventListener`s in each place we apply this pattern.

Maybe we need an abstraction to hold `EventListener`s? `EventDispatcher`, `EventPublisher`, etc?

torabshaikh:
@collindutter Just to be clear, We need to implement a new class named `EventsMixin` in the `events` module. This class will have a list of `EventListener` and method like `add_event_listener`, `remove_event_listener` and `publish_event`. Does `EventsMixin` needs to iterate over the list and publish on all `event_listener` or it needs to have a single `event_listener`. Do we need to remove all of these three methods from `Structure` or just `publish_event`? Then all of the drivers that right now uses `Structure` to publish event  can use `EventsMixin`  to do that instead.

collindutter:
> We need to implement a new class named EventsMixin in the events module.

Should be added in the `mixins` module, but yes.

> This class will have a list of `EventListener` and method like `add_event_listener`, `remove_event_listener` and `publish_event`.

Yes, good catch I forgot about those other methods.

>  Does EventsMixin needs to iterate over the list and publish on all event_listener or it needs to have a single event_listener

It should iterate over `event_listeners`, just as the methods works in `Structure` today.

> Do we need to remove all of these three methods from Structure or just publish_event?

Yes, remove the methods and field from `Structure`, and add `EventsMixin` to Structure.

> Then all of the drivers that right now uses Structure to publish event can use EventsMixin to do that instead.

Yes.

Finally: [this method](https://github.com/griptape-ai/griptape/blob/add234a75e6ead979b0a018d56c3385c588dcf84/griptape/tasks/prompt_task.py#L67-L72) should change to set `self.prompt_driver.event_listeners` instead of `self.prompt_driver.structure`.

collindutter:
@torabshaikh thanks for starting this work! Some issues in Discord bumped up the urgency of this fix so I am finishing it up here: https://github.com/griptape-ai/griptape/pull/984.

PR #917, opened 2024-07-01
DynamoDB sort key

datashaman:
With the current code, the table items are get/put with the partition key only. Most users will already have a table, and it's likely to have a sort key, which leads them to creating a table just for the memory.

DynamoDB encourages the use of a single database with multiple types of data within.

This change adds the sort key and value to the get/put actions (if supplied), making it possible to use with an existing table with a sort key defined.

Technically, the partition and sort key values can be _string_, _number_ or _binary_. I can't imagine why anyone would use `float` or `bytes` as keys (despite DynamoDB supporting them), so I've declared it as `str | int`.

The partition key value _could_ also be declared as such for more flexibility (it has the same allowed types as the sort key value), but that's your call - I'd rather not change an existing declaration.


collindutter:
Excellent change, thanks! Could you please update the changelog with a message following [this convention](https://keepachangelog.com/en/1.1.0/)?

datashaman:
added a line in the docs as well, hope it's ok.

collindutter (approved):

collindutter:
Even better, thanks!

PR #919, opened 2024-07-01
Ensure Executor.shutdown() called after task submission in BaseEventListenerDriver.publish_event()

dylanholmes:
Problem: Missed one instance of context manager use for executors in `BaseEventListenerDriver.publish_event()` related to the existing changelog note:

> All `futures_executor` fields renamed to `futures_executor_fn` and now accept callables instead of futures; wrapped all future `submit` calls with the `with` block to address future executor shutdown issues.

This particular instance was causing some intermittent unit tests failures in a feature I'm working on.

Changes:
- Use the `Executor` context manager when submitting tasks in `BaseEventListenerDriver.publish_event()`.


collindutter (dismissed):

collindutter (approved):
Hah thanks for adding the unit test, I was going to ignore.

PR #930, opened 2024-07-02
`RagContext` and other `RagEngine` improvements

vasinov:
## Describe your changes

* Added `module_params` to `RagContext`; updated modules to use config overrides from `module_params`.
* Removed `namespace` and `alternative_queries` from `RagContext`.
* Renamed RAG `generation` stage and modules to `response`.
* Added `TextChunksResponseRagModule`.
* Multiple minor refactors.

## Checklist before requesting a review
- [x] I have performed a self-review of my code.
- [x] I have added thorough tests.
- [x] Is this a notable change? If so, please update the [changelog](https://github.com/griptape-ai/griptape/blob/dev/CHANGELOG.md).

----
📚 Documentation preview 📚: https://griptape--930.org.readthedocs.build//930/


griptape/engines/rag/modules/base_rag_module.py
```diff
@@ -17,3 +17,12 @@ def generate_query_prompt_stack(self, system_prompt: str, query: str) -> PromptS
         return PromptStack(
             messages=[Message(system_prompt, role=Message.SYSTEM_ROLE), Message(query, role=Message.USER_ROLE)]
         )
+
+    def context_param(self, context: RagContext, key: str) -> Optional[Any]:
```
collindutter, line 21: Let's rename to `get_context_param` for consistency with `set_context_param`

griptape/engines/rag/modules/base_rag_module.py
```diff
@@ -17,3 +17,12 @@ def generate_query_prompt_stack(self, system_prompt: str, query: str) -> PromptS
...
+        return context.module_params.get(self.name, {}).get(key)
+
+    def set_context_param(self, context: RagContext, key: str, value: Any) -> None:
+        if not isinstance(context.module_params.get(self.name), dict):
```
collindutter, line 25: Should this be `if self.name not in context.module_params`? Using `isinstance` feels inappropriate but maybe I'm missing something.

griptape/memory/task/storage/text_artifact_storage.py
```diff
@@ -36,12 +42,22 @@ def summarize(self, namespace: str) -> TextArtifact:
...
+    def query(self, namespace: str, query: str, metadata: Any = None) -> BaseArtifact:
         if self.rag_engine is None:
             raise ValueError("RAG engine is not set.")
 
         result = self.rag_engine.process(
-            RagContext(initial_query=query, namespace=namespace, metadata=None if metadata is None else str(metadata))
+            RagContext(
+                query=query,
+                module_params={
+                    self.retrieval_rag_module_name: {  # pyright: ignore
```
collindutter, line 53: What is the pyright issue? If it's required, only ignore the specific error, not all errors.

griptape/templates/engines/rag/modules/response/prompt/system.j2
```diff
@@ -0,0 +1,19 @@
...
+{{ before_system_prompt }}
+
+{% endif %}
+Always be truthful. Don't make up facts. You can answer questions by searching through text chunks. Use the following list of text chunks to respond. If there are no text chunks available or text chunks don't have relevant information respond with "I could not find an answer."
+
+{% if text_chunks and text_chunks|length > 0 %}
+{% for chunk in text_chunks %}
+Text chunk: """
+{{ chunk }}
+"""
```
collindutter, line 11: I know we do this in other engines too, but do the chunks belong in the system prompt? I always thought the system prompt was better for describing _how_ to do something ("answer questions by searching through text chunks"), and the user prompt for _what_ to do ("look through these specific chunks to answer this question").

vasinov: That's a good question. I think that's something we need to evaluate once we have evals. I suspect the result will depend on the underlying LLM. The nice thing about modular RAG is that we can easily try those different approaches by substituting response modules.

griptape/engines/rag/modules/base_rag_module.py
vasinov: I am testing for presence and for type being explicitly a `dict`.

collindutter (dismissed):

collindutter (approved):

PR #934, opened 2024-07-03
Update pyright, fix new pyright errors

collindutter:

## Describe your changes
Upgrade pyright and and fix type errors when passing non string data into `TextArtifact`. I.e.

```
/Users/collindutter/Documents/griptape/griptape/griptape/tasks/actions_subtask.py
  /Users/collindutter/Documents/griptape/griptape/griptape/tasks/actions_subtask.py:48:33 - error: Argument of type "str | None" cannot be assigned to parameter "value" of type "str" in function "__init__"
    Type "str | None" is incompatible with type "str"
      "None" is incompatible with "str" (reportArgumentType)
```

Ideally pyright would pick up the fact that we've got a `converter=str` on `TextArtifact`. Should we ignore these errors instead?
NA


vachillo:
[allegedly this should work](https://github.com/microsoft/pyright/issues/7713).

collindutter:
Good find with that issue. I think as long as `attrs` supports it, I'm fine with using the experimental features from pyright.

vachillo (approved):

PR #939, opened 2024-07-04
Add task to structure when possible

vachillo:

## Describe your changes
this adds the task to the structure that is associated with either task when using `add_parent/child` when passing a `BaseTask`

### DIFF
--- a/griptape/tasks/base_task.py
+++ b/griptape/tasks/base_task.py
@@ -76,9 +76,12 @@ class BaseTask(ABC):
 
     def add_parent(self, parent: str | BaseTask) -> None:
         parent_id = parent if isinstance(parent, str) else parent.id
+        parent_task = parent if isinstance(parent, BaseTask) else None
 
         if parent_id not in self.parent_ids:
             self.parent_ids.append(parent_id)
+
+        self._add_to_structure(parent_task)
 
     def add_children(self, children: list[str | BaseTask]) -> None:
         for child in children:
@@ -86,9 +89,12 @@ class BaseTask(ABC):
 
     def add_child(self, child: str | BaseTask) -> None:
         child_id = child if isinstance(child, str) else child.id
+        child_task = child if isinstance(child, BaseTask) else None
 
         if child_id not in self.child_ids:
             self.child_ids.append(child_id)
+
+        self._add_to_structure(child_task)
 
     def preprocess(self, structure: Structure) -> BaseTask:
         self.structure = structure
@@ -168,3 +174,15 @@ class BaseTask(ABC):
             return structure_context
         else:
             return {}
+
+    def _add_to_structure(self, task: Optional[BaseTask]):
+        if task is not None:
+            if self.structure is not None and task.structure is not None and self.structure != task.structure:
+                raise ValueError("Task is already associated with a different structure.")
+            elif self.structure:
+                if task not in self.structure.tasks:
+                    self.structure.add_task(task)  # pyright: ignore reportArgumentType
+            elif task.structure:
+                if self not in task.structure.tasks:
+                    self.preprocess(task.structure)
+                    task.structure.add_task(self)
--- a/griptape/tasks/base_task.py
+++ b/griptape/tasks/base_task.py
@@ -175,14 +175,13 @@ class BaseTask(ABC):
         else:
             return {}
 
-    def _add_to_structure(self, task: Optional[BaseTask]):
+    def _add_to_structure(self, task: Optional[BaseTask]) -> None:
         if task is not None:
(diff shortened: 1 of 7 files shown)


griptape/tasks/base_task.py
```diff
@@ -168,3 +177,14 @@ def full_context(self) -> dict[str, Any]:
             return structure_context
         else:
             return {}
+
+    def _attach_to_structure(self, task: BaseTask):
```
collindutter, line 181: What do you think about renaming to `_add_to_structure`? "attach" conveys more of a 1-to-1 relationship.

griptape/utils/__init__.py
```diff
@@ -23,6 +23,10 @@ def minify_json(value: str) -> str:
     return json.dumps(json.loads(value), separators=(",", ":"))
 
 
+def not_none(value):
+    return value is not None
+
+
```
collindutter, line 29: I don't think we should create a utility method for this.

collindutter (changes requested):

griptape/utils/__init__.py
vachillo: was hoping to sneak it past you

griptape/tasks/base_task.py
```diff
@@ -76,20 +76,26 @@ def add_parents(self, parents: list[str | BaseTask]) -> None:
 
     def add_parent(self, parent: str | BaseTask) -> None:
         parent_id = parent if isinstance(parent, str) else parent.id
+        parent_task = parent if isinstance(parent, BaseTask) else None
 
         if parent_id not in self.parent_ids:
             self.parent_ids.append(parent_id)
 
+        self._add_to_structure(parent_task)
```
collindutter, line 84: Won't this add a `None` Task if a parent id is provided?

griptape/tasks/base_task.py
```diff
@@ -168,3 +174,15 @@ def full_context(self) -> dict[str, Any]:
             return structure_context
         else:
             return {}
+
+    def _add_to_structure(self, task: Optional[BaseTask]):
```
collindutter, line 178: Missing return type

griptape/tasks/base_task.py
```diff
@@ -168,3 +174,15 @@ def full_context(self) -> dict[str, Any]:
...
+        if task is not None:
+            if self.structure is not None and task.structure is not None and self.structure != task.structure:
+                raise ValueError("Task is already associated with a different structure.")
+            elif self.structure:
```
collindutter, line 182: Use `is not None` check

griptape/tasks/base_task.py
```diff
@@ -168,3 +174,15 @@ def full_context(self) -> dict[str, Any]:
...
+                if task not in self.structure.tasks:
+                    self.structure.add_task(task)  # pyright: ignore reportArgumentType
```
collindutter, line 184: Why the pyright ignore?

griptape/tasks/base_task.py
```diff
@@ -168,3 +174,15 @@ def full_context(self) -> dict[str, Any]:
...
+            elif task.structure:
```
collindutter, line 185: Use explicit `is not None` check

griptape/tasks/base_task.py
```diff
@@ -168,3 +174,15 @@ def full_context(self) -> dict[str, Any]:
...
+                if self not in task.structure.tasks:
+                    self.preprocess(task.structure)
```
collindutter, line 187: Doesn't `Structure.add_task` call `preprocess`?

collindutter (changes requested):

griptape/tasks/base_task.py
vachillo: it says that `BaseTask | None` is not valid for `add_task`. even though theres a None check

griptape/tasks/base_task.py
vachillo: there was a reason this was here...seems like its not needed now.

griptape/tasks/base_task.py
vachillo: is there a linting rule that can be added for this?

griptape/tasks/base_task.py
vachillo: oh it passes now

tests/unit/structures/test_workflow.py
collindutter, line 1: Can we leave this test as is?

collindutter, line 1: I think better to leave the existing tests as is and add new tests for each of the Structures.

collindutter, line 1: Can you also add tests for Pipeline and Agent?

PR #940, opened 2024-07-05
Update ruff linter rules according to documentation

collindutter:

## Describe your changes
Enables more ruff linting rules as suggested by [the documentation](https://docs.astral.sh/ruff/linter/#rule-selection). Also disables some that interfere with some core functionality. We can evaluate removing these in the future.
NA.

### DIFF
--- a/griptape/drivers/file_manager/amazon_s3_file_manager_driver.py
+++ b/griptape/drivers/file_manager/amazon_s3_file_manager_driver.py
@@ -55,7 +55,7 @@ class AmazonS3FileManagerDriver(BaseFileManagerDriver):
             return response["Body"].read()
         except botocore.exceptions.ClientError as e:
             if e.response["Error"]["Code"] in {"NoSuchKey", "404"}:
-                raise FileNotFoundError
+                raise FileNotFoundError from e
             else:
                 raise e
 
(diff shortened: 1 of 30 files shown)


collindutter:
I'm on the fence about [SIM108](https://docs.astral.sh/ruff/rules/if-else-block-instead-of-if-exp/), but I think i'm leaning towards including it for consistency.

griptape/drivers/file_manager/amazon_s3_file_manager_driver.py
```diff
@@ -55,7 +55,7 @@ def try_load_file(self, path: str) -> bytes:
             return response["Body"].read()
         except botocore.exceptions.ClientError as e:
             if e.response["Error"]["Code"] in {"NoSuchKey", "404"}:
-                raise FileNotFoundError
+                raise FileNotFoundError from e
```
dylanholmes, line 58: TIL - I've actually never seen this syntax before, cool!

pyproject.toml
```diff
@@ -200,7 +200,22 @@ line-length = 120
 skip-magic-trailing-comma = true
 
 [tool.ruff.lint]
-select = ["E4", "E7", "E9", "F", "TID251"]
+select = [
+    "E", # pycodestyle
```
dylanholmes, line 204: Thanks for adding the comments!

dylanholmes (approved):
Love the consistency

PR #941, opened 2024-07-05
Format span and trace id as UUIDs in Griptape Cloud

dylanholmes:
Changes:
- Format span and trace id as UUIDs in Griptape Cloud

## NOTE: Merging into `release/cloud-unreleased` rather than dev!

### DIFF
--- a/griptape/drivers/observability/griptape_cloud_observability_driver.py
+++ b/griptape/drivers/observability/griptape_cloud_observability_driver.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+from typing import Optional
 import requests
 
 from uuid import UUID
@@ -80,7 +81,7 @@ class GriptapeCloudObservabilityDriver(OpenTelemetryObservabilityDriver):
                 "structure_run_id must be set either in the constructor or as an environment variable (GT_CLOUD_STRUCTURE_RUN_ID)."
             )
 
-    def get_span_id(self) -> str | None:
+    def get_span_id(self) -> Optional[str]:
         span = get_current_span()
         if span is INVALID_SPAN:
             return None
--- a/griptape/drivers/observability/griptape_cloud_observability_driver.py
+++ b/griptape/drivers/observability/griptape_cloud_observability_driver.py
@@ -1,11 +1,14 @@
+from __future__ import annotations
+
 import os
 import requests
 
+from uuid import UUID
 from collections.abc import Sequence
 
 from attrs import define, Factory, field
 from griptape.drivers.observability.open_telemetry_observability_driver import OpenTelemetryObservabilityDriver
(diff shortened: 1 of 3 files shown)


griptape/drivers/observability/griptape_cloud_observability_driver.py
```diff
@@ -48,9 +51,9 @@ def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
             url = urljoin(self.base_url.strip("/"), f"/api/structure-runs/{self.structure_run_id}/spans")
             payload = [
                 {
-                    "trace_id": format_trace_id(span.context.trace_id),
-                    "span_id": format_span_id(span.context.span_id),
-                    "parent_id": format_span_id(span.parent.span_id) if span.parent else None,
+                    "trace_id": str(UUID(int=span.context.trace_id)),
+                    "span_id": str(UUID(int=span.context.span_id)),
+                    "parent_id": str(UUID(int=span.parent.span_id)) if span.parent else None,
```
collindutter, line 56: This is a bit inconsistent with how we generate uuids in [other places](https://github.com/griptape-ai/griptape/blob/1cdc5b98b9b1af30aa087b566234ec3f8280bc36/griptape/structures/structure.py#L41). And what does the `UUID(int=...)` syntax do? Is it generating a uuid v5 using the span/trace id as the seed?

griptape/drivers/observability/griptape_cloud_observability_driver.py
```diff
@@ -76,3 +79,9 @@ def validate_run_id(self, _, structure_run_id: str):
             raise ValueError(
                 "structure_run_id must be set either in the constructor or as an environment variable (GT_CLOUD_STRUCTURE_RUN_ID)."
             )
+
+    def get_span_id(self) -> str | None:
+        span = get_current_span()
+        if span is INVALID_SPAN:
+            return None
+        return str(UUID(int=span.get_span_context().span_id))
```
collindutter, line 87: Same uuid comment here.

collindutter (changes requested):

griptape/drivers/observability/griptape_cloud_observability_driver.py
dylanholmes: It's not generating a uuid, its just using `str(UUID(int=...))` to format the integer id generated by open telemetry. The range of the integer values supported by UUID are the compatible with the range generated by open telemetry.

We are using this value instead of generating our own to make it easy to get the correct value of parent spans and on events. If we generated our own, then we'd need to keep track of a mapping between open telemetry id and our generated id, but since they are 1:1, there's no reason not to just use the one generated by open telemetry.

collindutter (approved):

collindutter (approved):

PR #942, opened 2024-07-08
`FootnotePromptResponseRagModule`, `TextLoaderRetrievalRagModule`, and more

vasinov:

## Describe your changes

- Added `FootnotePromptResponseRagModule` to support citations in RAG responses.
- Added `TextLoaderRetrievalRagModule` for realtime data loading in RAG engine.
- Added `Reference`.
- Integrated `Reference` in loaders.
- Refactors and bug fixes.

----
📚 Documentation preview 📚: https://griptape--942.org.readthedocs.build//942/


griptape/schemas/base_schema.py
```diff
@@ -59,7 +59,7 @@ def _get_field_for_type(cls, field_type: type) -> fields.Field | fields.Nested:
             if ABC in field_class.__bases__:
                 return fields.Nested(PolymorphicSchema(inner_class=field_class), allow_none=optional)
             else:
-                return fields.Nested(cls.from_attrs_cls(field_type), allow_none=optional)
+                return fields.Nested(cls.from_attrs_cls(field_class), allow_none=optional)
```
vasinov, line 62: This change represents three hours of my life on Saturday 🤪

CHANGELOG.md
```diff
@@ -52,6 +54,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Agent.input` for passing Artifacts as input.
 - Support for `PromptTask`s to take `TextArtifact`s, `ImageArtifact`s, and `ListArtifact`s as input.
 - Parameters `sort_key` and `sort_key_value` on `AmazonDynamoDbConversationMemoryDriver` for tables with sort keys.
+- `Citation` for supporting artifact citations in loaders and RAG engine modules.
```
collindutter, line 57: Do you mean `Reference`?

griptape/schemas/base_schema.py
collindutter: Yeah this is a fun area of the framework :) Can you expand a bit on what the issue was and why this fixes it?

griptape/artifacts/base_artifact.py
```diff
@@ -1,15 +1,20 @@
...
 
+if TYPE_CHECKING:
+    from griptape.common import Reference
 
-@define()
+
+@define
 class BaseArtifact(SerializableMixin, ABC):
     id: str = field(default=Factory(lambda: uuid.uuid4().hex), kw_only=True, metadata={"serializable": True})
+    reference: Optional[Reference] = field(default=None, kw_only=True, metadata={"serializable": True})
```
collindutter, line 16: `Reference`s are very specific to a particular RAG Module, does this belong on all Artifacts? What if we created a `ReferenceArtifact` with `Reference` as an inner class?

griptape/loaders/base_text_loader.py
```diff
@@ -43,12 +45,14 @@ def _text_to_artifacts(self, text: str) -> list[TextArtifact]:
 
         chunks = self.chunker.chunk(text) if self.chunker else [TextArtifact(text)]
 
-        if self.embedding_driver:
-            for chunk in chunks:
+        for chunk in chunks:
+            if self.embedding_driver:
```
collindutter, line 49: Why the swap? Isn't it cheaper to check if the embedding driver is set once vs every iteration?

griptape/loaders/base_text_loader.py
```diff
@@ -32,6 +30,10 @@ class BaseTextLoader(BaseLoader, ABC):
     )
     embedding_driver: Optional[BaseEmbeddingDriver] = field(default=None, kw_only=True)
     encoding: str = field(default="utf-8", kw_only=True)
+    reference: Optional[Reference] = field(default=None, kw_only=True)
```
collindutter, line 33: Should this be on `BaseLoader`? Or should `Reference` only live on `TextArtifact`s?

griptape/artifacts/base_artifact.py
dylanholmes, line 16: Just to make sure I understand, `reference` will only be populated for excerpts loaded directly from the source (or vector db).  Artifacts that are derived from source(s) will not have `reference` populated, but will instead contain footnotes directly in the artifact content. Right?

vasinov: I want reference to be uniformly adopted throughout the framework where it makes sense. For example, I want all loaders to support it.

griptape/loaders/base_text_loader.py
vasinov: It was hard for me to make sense of the code during the initial read and I realized it's because we had the same for loop twice. I just combined them into one. Conditional checks are very cheap performance-wise, so I think it's a reasonable tradeoff.

griptape/loaders/base_text_loader.py
vasinov: Yes but there is more to it. Will start a Slack thread :)

griptape/engines/rag/modules/response/prompt_response_rag_module.py
```diff
@@ -11,50 +12,52 @@
...
         return J2("engines/rag/modules/response/prompt/system.j2").render(
-            text_chunks=text_chunks,
-            before_system_prompt="\n\n".join(before_system_prompt),
-            after_system_prompt="\n\n".join(after_system_prompt),
+            text_chunks=[c.to_text() for c in artifacts],
+            before_system_prompt="\n\n".join(context.before_query),
+            after_system_prompt="\n\n".join(context.after_query),
         )
+
+    def references_from_artifacts(self, artifacts: list[TextArtifact]) -> list[Reference]:
```
dylanholmes, line 56: Does this method belong in this class? It isn't used by the class and doesn't use the class' state. Seems like it belongs in utils

griptape/engines/rag/modules/retrieval/text_loader_retrieval_rag_module.py
```diff
@@ -0,0 +1,41 @@
...
+        query_params["namespace"] = namespace
+
+        loader_output = self.loader.load(source)
+
+        if isinstance(loader_output, ErrorArtifact):
+            raise Exception(loader_output.to_text() if loader_output.exception is None else loader_output.exception)
+        else:
+            self.vector_store_driver.upsert_text_artifacts({namespace: loader_output})
+
+            return self.process_query_output_fn(self.vector_store_driver.query(context.query, **query_params))
```
dylanholmes, line 41: Is this a general pattern for Retrieval Modules? Retrieve from original source, upsert into vector db, then query the vector db and return results from that query?

griptape/loaders/base_text_loader.py
```diff
@@ -43,12 +45,14 @@ def _text_to_artifacts(self, text: str) -> list[TextArtifact]:
...
         chunks = self.chunker.chunk(text) if self.chunker else [TextArtifact(text)]
 
-        if self.embedding_driver:
-            for chunk in chunks:
+        for chunk in chunks:
+            if self.embedding_driver:
                 chunk.generate_embedding(self.embedding_driver)
 
-        for chunk in chunks:
+            chunk.reference = self.reference
```
dylanholmes, line 52: Out of scope for this PR, but In the future, would we ever want to add additional info to chunk references, like offset into document or something? Doesn't seem that difficult to do from what's added in this PR, but just curious if it had crossed your mind at all

dylanholmes (dismissed):
Awesome! Like the naming choice of `footnote` too.

Approved, only minor comments/questions

griptape/engines/rag/modules/response/prompt_response_rag_module.py
vasinov: Moving it to utils.

griptape/engines/rag/modules/retrieval/text_loader_retrieval_rag_module.py
vasinov: I think for some it will be. For example, `VectorStoreRetrievalRagModule` queries the vector store passed to it. But I imagine other retrieval modules, that are more akin to search engines, will utilize this more general pattern. We'll DRY it up then 😄.

griptape/loaders/base_text_loader.py
vasinov: Interesting! Yeah, I think that would be very cool, especially for some things that can be split up into logical sections more easily (e.g., markdown docs).

griptape/engines/rag/modules/response/prompt_response_rag_module.py
dylanholmes: Thanks for moving the implementation to utils, but how come you kept the method?

vasinov: Oops, I meant to call it from `RagContext` not `PromptResponseRagModule`. Fixed.

dylanholmes (approved):
🚀

collindutter (approved):

PR #951, opened 2024-07-09
Add ruff rule for complex code paths

collindutter:

## Describe your changes
Adds [C901](https://docs.astral.sh/ruff/rules/complex-structure/) for measuring complexity of functions. Ignoring certain functions that are very old and not worth the effort to refactor at this time.

### DIFF
--- a/griptape/chunkers/base_chunker.py
+++ b/griptape/chunkers/base_chunker.py
@@ -31,7 +31,7 @@ class BaseChunker(ABC):
 
         return [TextArtifact(c) for c in self._chunk_recursively(text)]
 
-    def _chunk_recursively(self, chunk: str, current_separator: Optional[ChunkSeparator] = None) -> list[str]:
+    def _chunk_recursively(self, chunk: str, current_separator: Optional[ChunkSeparator] = None) -> list[str]:  # noqa: C901
         token_count = self.tokenizer.count_tokens(chunk)
 
         if token_count <= self.max_tokens:
--- a/griptape/chunkers/base_chunker.py
+++ b/griptape/chunkers/base_chunker.py
@@ -31,7 +31,7 @@ class BaseChunker(ABC):
 
         return [TextArtifact(c) for c in self._chunk_recursively(text)]
 
-    def _chunk_recursively(self, chunk: str, current_separator: Optional[ChunkSeparator] = None) -> list[str]:  # noqa: C901
+    def _chunk_recursively(self, chunk: str, current_separator: Optional[ChunkSeparator] = None) -> list[str]:
         token_count = self.tokenizer.count_tokens(chunk)
 
         if token_count <= self.max_tokens:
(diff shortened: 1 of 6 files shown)


griptape/drivers/image_generation_model/bedrock_stable_diffusion_image_generation_model_driver.py
```diff
@@ -107,30 +107,21 @@ def _request_parameters(
...
-
-        if self.steps is not None:
-            request["steps"] = self.steps
-
-        if seed is not None:
-            request["seed"] = seed
+        request = {
+            "text_prompts": text_prompts,
+            "cfg_scale": self.cfg_scale,
+            **({"style_preset": self.style_preset} if self.style_preset is not None else {}),
```
cjkindel, line 113: This approach seems unnatural, are we sure we want a rule that requires this?

collindutter: We do use it in other places but maybe wrongly so. What do you think about the updated approach?

griptape/chunkers/base_chunker.py
```diff
@@ -31,7 +31,7 @@ def chunk(self, text: TextArtifact | str) -> list[TextArtifact]:
 
         return [TextArtifact(c) for c in self._chunk_recursively(text)]
 
-    def _chunk_recursively(self, chunk: str, current_separator: Optional[ChunkSeparator] = None) -> list[str]:
+    def _chunk_recursively(self, chunk: str, current_separator: Optional[ChunkSeparator] = None) -> list[str]:  # noqa: C901
```
dylanholmes, line 34: I'm not sure I like adding a rule that only leads to one violation being "fixed", while the other 3 violations are immediately ignored.

Is the goal to just get authors of PRs to think twice before adding complex functions? (but still allow them if the complexity is warranted?)

dylanholmes (commented):
Just one "why?" comment. I'll give a ship it given a reasonable response

griptape/chunkers/base_chunker.py
collindutter: > Is the goal to just get authors of PRs to think twice before adding complex functions?

Yes. Ignoring these is 100% lazyness on my end -- there was not an immediately easy refactor available. I'm more concerned about new code, but I can definitely spend the time fixing these now.

griptape/structures/workflow.py
```diff
@@ -64,37 +64,12 @@ def insert_task(
     ) -> BaseTask:
         task.preprocess(self)
 
-        for child_task in child_tasks:
-            # Link the new task to the child task
-            if child_task.id not in task.child_ids:
-                task.child_ids.append(child_task.id)
-            if task.id not in child_task.parent_ids:
-                child_task.parent_ids.append(task.id)
+        self.__link_task_to_children(task, child_tasks)
```
dylanholmes, line 67: Love the refactor ❤️

dylanholmes (approved):

dylanholmes (approved):

PR #968, opened 2024-07-12
Improve native tool name check

collindutter:

## Describe your changes
Validates that Tool name is alphanumeric, and Tool Activity name is alphanumeric with underscores.


collindutter:
Code cov will probably fail for not checking the activity name branch. This is somewhat non-trivial to test since right now you can't change an Activity's name -- it's set to the function name which is already enforced to be alphanumeric + underscores by python.

dylanholmes:
> Code cov will probably fail for not checking the activity name branch. This is somewhat non-trivial to test since right now you can't change an Activity's name -- it's set to the function name which is already enforced to be alphanumeric + underscores by python.

How about: Mock the class, then have `activity_name()` return something invalid?

collindutter:
@dylanholmes great call, updated.

dylanholmes (approved):
🙌

PR #969, opened 2024-07-12
Fix inserting conversation into  when no system content is present.

collindutter:

## Describe your changes
Conversation Memory is being incorrectly inserted into index 1 in the Prompt Stack when there is no system message.
NA

### DIFF
--- a/griptape/memory/structure/base_conversation_memory.py
+++ b/griptape/memory/structure/base_conversation_memory.py
@@ -88,9 +88,9 @@ class BaseConversationMemory(SerializableMixin, ABC):
 
         if num_runs_to_fit_in_prompt:
             memory_inputs = self.to_prompt_stack(num_runs_to_fit_in_prompt).messages
-            if index is not None:
-                prompt_stack.messages[index:index] = memory_inputs
-            else:
+            if index is None:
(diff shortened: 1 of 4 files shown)


griptape/memory/structure/base_conversation_memory.py
```diff
@@ -88,7 +88,7 @@ def add_to_prompt_stack(self, prompt_stack: PromptStack, index: Optional[int] =
 
         if num_runs_to_fit_in_prompt:
             memory_inputs = self.to_prompt_stack(num_runs_to_fit_in_prompt).messages
-            if index:
+            if index is not None:
```
collindutter, line 91: This is why we check is not None!

vasinov: Can you flip the if-else to be `if is None ... else ...` for better readability?

vasinov (changes requested):

griptape/memory/structure/base_conversation_memory.py
collindutter: One of these days I'll get it first try...

I wonder if there's a linter rule for this?

vasinov (approved):

PR #974, opened 2024-07-12
Add flake8-commas ruff rule

collindutter:

## Describe your changes
Add [flake8-commas](https://docs.astral.sh/ruff/rules/#flake8-commas-com) ruff rule

### DIFF
--- a/griptape/artifacts/base_artifact.py
+++ b/griptape/artifacts/base_artifact.py
@@ -19,7 +19,9 @@ class BaseArtifact(SerializableMixin, ABC):
     reference: Optional[Reference] = field(default=None, kw_only=True, metadata={"serializable": True})
     meta: dict[str, Any] = field(factory=dict, kw_only=True, metadata={"serializable": True})
     name: str = field(
-        default=Factory(lambda self: self.id, takes_self=True), kw_only=True, metadata={"serializable": True}
+        default=Factory(lambda self: self.id, takes_self=True),
+        kw_only=True,
+        metadata={"serializable": True},
     )
     value: Any = field()
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -228,7 +228,7 @@ ignore = [
     "D105", # undocumented-magic-method
     "D106", # undocumented-public-nested-class
     "D107", # undocumented-public-init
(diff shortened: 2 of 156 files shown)


griptape/artifacts/base_artifact.py
```diff
@@ -19,7 +19,9 @@ class BaseArtifact(SerializableMixin, ABC):
     reference: Optional[Reference] = field(default=None, kw_only=True, metadata={"serializable": True})
     meta: dict[str, Any] = field(factory=dict, kw_only=True, metadata={"serializable": True})
     name: str = field(
-        default=Factory(lambda self: self.id, takes_self=True), kw_only=True, metadata={"serializable": True}
+        default=Factory(lambda self: self.id, takes_self=True),
+        kw_only=True,
+        metadata={"serializable": True},
```
dylanholmes, line 24: I personally like this rule. When the last item ends with a comma, then the diff of appending a new item looks so clean.

pyproject.toml
```diff
@@ -229,6 +227,7 @@ ignore = [
     "D105", # undocumented-magic-method
     "D106", # undocumented-public-nested-class
     "D107", # undocumented-public-init
+    "COM812" # missing-trailing-comma -- disabled for compatability with formatter.
```
dylanholmes, line 230: Wait, don't you have trailing commas now? Should this ignore be here?

dylanholmes (changes requested):

griptape/artifacts/base_artifact.py
collindutter: Agreed, and that is what ruff mentions in the [justification](https://docs.astral.sh/ruff/rules/missing-trailing-comma/) of why it's good

pyproject.toml
collindutter: Yes, otherwise we get this warning:

```
The following rules may cause conflicts when used with the formatter: `COM812`. To avoid unexpected behavior, we recommend disabling these rules, either by removing them from the `select` or `extend-select` configuration, or adding them to the `ignore` configuration.
```

See the new comment's attached issue.

collindutter: The important part is that the formatter will add the commas so we effectively get the same behavior.

dylanholmes (approved):

PR #983, opened 2024-07-15
Add generic artifact

collindutter:


collindutter:
Not tied to the name. Would `AnyArtifact` be more appropriate? Or maybe even just `Artifact`?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Native function calling support to `OpenAiChatPromptDriver`, `AzureOpenAiChatPromptDriver`, `AnthropicPromptDriver`, `AmazonBedrockPromptDriver`, `GooglePromptDriver`, and `CoherePromptDriver`.
 - `OllamaEmbeddingDriver` for generating embeddings with Ollama.
 - `GriptapeCloudKnowledgeBaseVectorStoreDriver` to query Griptape Cloud Knowledge Bases.
+- `GenericArtifact` for storing any data.
```
dylanholmes, line 12: I don't feel strongly about it, but what about `AnyArtifact` since it is similar to the `Any` python type?

dylanholmes (dismissed):
Pretty elegant escape hatch 🐣 

Just mentioned one alternative naming choice which I will leave to your discretion.

CHANGELOG.md
dylanholmes: Now I vote for `AnyArtifact` since we both had the same thought independently.

collindutter: An argument against `AnyArtifact` is that it's very python-ey whereas the other artifact types are intentionally not. I.e. 
- `TextArtifact` -> `value: str`
- `BlobArtifact` -> `value: bytes`
- etc

collindutter: Let's see if @vasinov or @emjay07 have any thoughts. @emjay07 is going to play with it tomorrow so that may lead to some naming insight.

docs/examples/talk-to-a-video.md
```diff
@@ -0,0 +1,83 @@
...
+
+```python
+import time
+from griptape.structures import Agent
+from griptape.tasks import PromptTask
+from griptape.artifacts import ListArtifact, GenericArtifact, TextArtifact
+from griptape.config import GoogleStructureConfig
+import google.generativeai as genai
+
+video_file = genai.upload_file(path="tests/resources/griptape-comfyui.mp4")
```
emjay07, line 13: do we want to put a wrapper around this at all so that a user doesn't need to configure both Griptape and the genai API with a key?

CHANGELOG.md
emjay07, line 12: I'm wondering if `GriptapeArtifact` or event just `ArtifactWrapper` make sense here? Barring those options, I'm kind of more for just `Artifact`. 

Essentially I see this as "make the random information of mine into an object type that Griptape understands", so while `GenericArtifact` gets the point across, it's not very descriptive for what the intent is.

docs/examples/talk-to-a-video.md
```diff
@@ -0,0 +1,83 @@
...
+    raise ValueError(video_file.state.name)
+
+agent = Agent(
+    config=GoogleStructureConfig(),
+    tasks=[
+        PromptTask(
+            lambda task: ListArtifact(
+                [
+                    GenericArtifact(video_file),
+                    TextArtifact(task.full_context["args"][0]),
```
emjay07, line 28: I'm not sure if this is generalizable, so happy to discuss. Hoping we can simplify how this is called so it's a bit more intuitive and consistent with how other PromptTasks work. It's a bit confusing that I'm not passing in a "prompt" anymore, but instead this blob of info. Also remembering to access `task.full_context` doesn't really follow our "easy to use" interface.

Something like the following:
```
PromptTask(prompt, other_artifacts=list[Artifact(video_file)])
```

emjay07 (commented):
My comments are more around usability than anything. I know Google is a bit wonky on how they call stuff, so it may not be feasible to make it perfect in the first rev.

CHANGELOG.md
collindutter: @emjay07 take a look at the updated example, I simplified the syntax quite a bit.

docs/examples/talk-to-a-video.md
collindutter: @emjay07 take a look at the updated example, I simplified the syntax quite a bit.

docs/examples/talk-to-a-video.md
collindutter: I'm not really sure what to wrap it in since they're the only provider that offers this.

We could put it in the Prompt Driver, but then the API diverges. This is a but clunky but it's kind of an interesting way to show how you can mix griptape with native SDK functionality.

emjay07: That's a good point on the mix of native SDK + Griptape. We can leave it at that for the V1.

One thing I am thinking about for the future is if we can use our `StructureConfig` as a generic config to make API calls? It has the entire config to make it work. So essentially like `GoogleStructureConfig().api_call('upload_file', args={'path':'tests/resources/griptape-comfyui.mp4'})` type of thing.

Anyhow, future things if it comes up again.

docs/examples/talk-to-a-video.md
emjay07: Cool, I like it. 1:1 essentially with the API call right now.

emjay07 (approved):

vasinov (dismissed):
`Artifact` would be ideal but it's inconsistent with other abstraction names. I think `GenericArtifact` is fine.

griptape/artifacts/generic_artifact.py
```diff
@@ -0,0 +1,15 @@
...
+from typing import Any
+
+from attrs import define, field
+
+from griptape.artifacts import BaseArtifact
+
+
+@define
+class GenericArtifact(BaseArtifact):
+    value: Any = field(metadata={"serializable": True})
```
dylanholmes, line 12: What if its not serializable?

dylanholmes (approved):
Great job!

Just one question

dylanholmes (approved):

griptape/artifacts/generic_artifact.py
collindutter: I don't know off the top of my head but it'd be however Marshmallow handles it via `field.Raw`. We _do_ need it serializable though, so I think I'd rather just deal with it when we come across a value that is non-intuitively serializable.

Issue #826, opened 2024-06-04
Merge Config Does Not Reinitialize Non-Serializable Fields


collindutter:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Describe the bug**
`merge_config` does the following:
1. Call `to_dict` on the config.
2. Recursive merge in the new config
3. Return `from_dict` containing the merged config.

The problem is that `to_dict` throws away fields without `serializable: True` metadata such as `OpenAiChatPromptDriver.api_key`. Meaning if a non-serializable field is set before `merge_config`, it is lost on merge.

**To Reproduce**
**IMPORTANT**: Ensure `OPENAI_API_KEY` is not set in environment.
Run:
```python
config = StructureConfig(
    prompt_driver=OpenAiChatPromptDriver(
        model="gpt-4o",
        api_key="VALID KEY",
    )
).merge_config(
    {
        "prompt_driver": {"stream": True},
    }
)

agent = Agent(config=config)

agent.run("hello")
```
Output:
```
openai.OpenAIError: The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable
```

**Expected behavior**
`merge_config` should not throw away already set values.

collindutter:
It seems like replacing `self.to_dict` with `attrs.asdict(self)` could be a solution, but this fails with:
```
Traceback (most recent call last):
  File "/Users/collindutter/Documents/griptape/griptape-playground/src/misc/config-issue.py", line 21, in <module>
    config = StructureConfig(
  File "/Users/collindutter/Documents/griptape/griptape/griptape/config/base_structure_config.py", line 36, in merge_config
    base_config = asdict(self)
  File "/Users/collindutter/Documents/griptape/griptape-playground/.venv/lib/python3.9/site-packages/attr/_next_gen.py", line 211, in asdict
    return _asdict(
  File "/Users/collindutter/Documents/griptape/griptape-playground/.venv/lib/python3.9/site-packages/attr/_funcs.py", line 65, in asdict
    rv[a.name] = asdict(
  File "/Users/collindutter/Documents/griptape/griptape-playground/.venv/lib/python3.9/site-packages/attr/_funcs.py", line 56, in asdict
    v = getattr(inst, a.name)
AttributeError: model
```

Probably an `init=False` problem.

PR #984, opened 2024-07-15
Add EventsMixin

collindutter:

Closes https://github.com/griptape-ai/griptape/pull/914
Closes https://github.com/griptape-ai/griptape/issues/643

### DIFF
--- a/griptape/config/base_structure_config.py
+++ b/griptape/config/base_structure_config.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Optional
 from attrs import define, field
 
 from griptape.config import BaseConfig
+from griptape.events import EventListener
 from griptape.utils import dict_merge
 
 if TYPE_CHECKING:
@@ -19,6 +20,7 @@ if TYPE_CHECKING:
         BaseTextToSpeechDriver,
         BaseVectorStoreDriver,
     )
+    from griptape.structures import Structure
 
 
 @define
@@ -35,6 +37,26 @@ class BaseStructureConfig(BaseConfig, ABC):
     )
     text_to_speech_driver: BaseTextToSpeechDriver = field(kw_only=True, metadata={"serializable": True})
     audio_transcription_driver: BaseAudioTranscriptionDriver = field(kw_only=True, metadata={"serializable": True})
+
+    _structure: Optional[Structure] = field(default=None, kw_only=True, alias="structure")
+
+    @property
+    def structure(self) -> Optional[Structure]:
+        return self._structure
+
+    @structure.setter
+    def structure(self, structure: Structure) -> None:
+        self._structure = structure
+
+        event_listener = EventListener(self.structure.publish_event)
+
+        self.prompt_driver.add_event_listener(event_listener)
+        self.image_generation_driver.add_event_listener(event_listener)
+        self.image_query_driver.add_event_listener(event_listener)
+        self.embedding_driver.add_event_listener(event_listener)
+        self.vector_store_driver.add_event_listener(event_listener)
+        if self.conversation_memory_driver is not None:
+            self.conversation_memory_driver.add_event_listener(event_listener)
 
     def merge_config(self, config: dict) -> BaseStructureConfig:
         base_config = self.to_dict()
--- a/griptape/config/base_structure_config.py
+++ b/griptape/config/base_structure_config.py
@@ -7,6 +7,7 @@ from attrs import define, field
 
 from griptape.config import BaseConfig
 from griptape.events import EventListener
+from griptape.mixins.event_publisher_mixin import EventPublisherMixin
 from griptape.utils import dict_merge
 
 if TYPE_CHECKING:
@@ -38,7 +39,21 @@ class BaseStructureConfig(BaseConfig, ABC):
     text_to_speech_driver: BaseTextToSpeechDriver = field(kw_only=True, metadata={"serializable": True})
     audio_transcription_driver: BaseAudioTranscriptionDriver = field(kw_only=True, metadata={"serializable": True})
 
-    _structure: Optional[Structure] = field(default=None, kw_only=True, alias="structure")
+    _structure: Structure = field(default=None, kw_only=True, alias="structure")
+    _event_listener: Optional[EventListener] = field(default=None, kw_only=True, alias="event_listener")
+
+    @property
+    def drivers(self) -> list:
+        return [
+            self.prompt_driver,
+            self.image_generation_driver,
+            self.image_query_driver,
+            self.embedding_driver,
+            self.vector_store_driver,
+            self.conversation_memory_driver,
+            self.text_to_speech_driver,
+            self.audio_transcription_driver,
+        ]
 
     @property
     def structure(self) -> Optional[Structure]:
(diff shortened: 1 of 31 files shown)


griptape/tasks/prompt_task.py
```diff
@@ -68,8 +68,9 @@ def prompt_driver(self) -> BasePromptDriver:
 
     def preprocess(self, structure: Structure) -> PromptTask:
         super().preprocess(structure)
+
         if self.prompt_driver is not None:
-            self.prompt_driver.structure = structure
+            self.prompt_driver.add_event_listeners(self.structure.event_listeners)
```
collindutter, line 73: Calling this out. Is this weird? Should all tasks do this?

collindutter:
Still need to write up a changelog

griptape/mixins/events_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+from typing import TYPE_CHECKING
+
+from attrs import define, field
+
+if TYPE_CHECKING:
+    from griptape.events import BaseEvent, EventListener
+
+
+@define
+class EventsMixin:
```
dylanholmes, line 12: What about `EventPublisher` instead?

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -10,14 +10,14 @@
...
 from griptape.artifacts import BaseArtifact, ListArtifact, TextArtifact
-from griptape.mixins import SerializableMixin
+from griptape.mixins import EventsMixin, SerializableMixin
 
 if TYPE_CHECKING:
     from griptape.drivers import BaseEmbeddingDriver
 
 
 @define
-class BaseVectorStoreDriver(SerializableMixin, ABC):
```
dylanholmes, line 20: Should there be a `BaseDriver` class that has the common stuff for all drivers, like `EventsMixin` and `SerializableMixin`? Or are there exceptions to this?

griptape/tasks/prompt_task.py
dylanholmes: To me, it doesn't actually feel that much weirder than what was previously there, but the new way will result in different behavior if say you modify the event_listeners on the structure after the task was added to the structure. That wouldn't be that intuitive:

```py
agent = Agent(tasks=[PromptTask(...)])
agent.add_event_listener(...)
# now prompt_task.prompt_driver doesn't have the same event_listeners as the agent
```

griptape/utils/stream.py
```diff
@@ -64,8 +64,11 @@ def event_handler(event: BaseEvent) -> None:
             handler=event_handler,
             event_types=[CompletionChunkEvent, FinishPromptEvent, FinishStructureRunEvent],
         )
+
         self.structure.add_event_listener(stream_event_listener)
+        self.structure.config.prompt_driver.add_event_listener(stream_event_listener)
```
dylanholmes, line 69: Will this work for all cases?

dylanholmes (commented):
Overall, I like a lot of the change, just a couple of questions

griptape/mixins/events_mixin.py
collindutter: I like it!

griptape/drivers/vector/base_vector_store_driver.py
collindutter: Yeah probably. I've got an item to standardize all the Driver interfaces pre 1.0. So let's do that then.

griptape/tasks/prompt_task.py
collindutter: Yeah that's actually why I had to do [this](https://github.com/griptape-ai/griptape/blob/3622d31f20121e60b4e7878ef3f2aceedf401b71/tests/unit/events/test_event_listener.py#L42-L43) which I don't love. 

Any suggestions?

griptape/utils/stream.py
collindutter: Can you expand on which other cases you're envisioning?

dylanholmes: I suppose I was thinking about if prompt_driver is passed directly to structure, but I see now that that will implicitly create a config with the prompt driver, so I guess its not an issue.

Sorry for the vague comment, but I get the feeling like there are some edge cases that I'm missing. I suppose there probably aren't any for this particular usage since we know exactly which entities publish the event types listed above (just structure and prompt driver)

griptape/tasks/prompt_task.py
dylanholmes: Let me think about this for a couple minutes. I feel like this is going to make it way easier to shoot ourselves in the feet.

dylanholmes (commented):
Give me a minute to think about this one. Won't take too long

griptape/tasks/prompt_task.py
dylanholmes: Here's an idea: What if the structure adds an event listener to the drivers that just forwards the events to its own event listeners? The added event listener would have a reference to the structure, so nothing will ever get out of sync. If you add an event listener to the structure, it would still just work

collindutter: Check out the latest implementation, it cleaned up a few areas.

griptape/config/base_structure_config.py
```diff
@@ -36,6 +38,26 @@ class BaseStructureConfig(BaseConfig, ABC):
...
+    def structure(self, structure: Structure) -> None:
+        self._structure = structure
+
+        event_listener = EventListener(self.structure.publish_event)
+
+        self.prompt_driver.add_event_listener(event_listener)
+        self.image_generation_driver.add_event_listener(event_listener)
+        self.image_query_driver.add_event_listener(event_listener)
+        self.embedding_driver.add_event_listener(event_listener)
+        self.vector_store_driver.add_event_listener(event_listener)
```
dylanholmes, line 57: Potentially nice to add a method that returns all drivers in a config, then you could just iterate through them here

griptape/config/base_structure_config.py
```diff
@@ -36,6 +38,26 @@ class BaseStructureConfig(BaseConfig, ABC):
...
+        if self.conversation_memory_driver is not None:
+            self.conversation_memory_driver.add_event_listener(event_listener)
+
     def merge_config(self, config: dict) -> BaseStructureConfig:
```
dylanholmes, line 61: What will happen if you call `merge_config` when `_structure` is set? Would it basically be the same behavior as before when the drivers had references to the structure?

dylanholmes (approved):
Feels like the right direction, but I just want to make sure we define the expected behavior when mutating structure/structure config/drivers

(Accidentally approved this one)

dylanholmes (changes requested):
(that approval was accidental)

griptape/tasks/base_task.py
```diff
@@ -106,7 +106,7 @@ def is_executing(self) -> bool:
         return self.state == BaseTask.State.EXECUTING
 
     def before_run(self) -> None:
-        if self.structure:
+        if self.structure is not None:
```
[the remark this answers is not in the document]
collindutter: Yeah Tasks are intentionally tightly coupled with Structures -- they rely on them for more things than just Event Listeners.

griptape/config/base_structure_config.py
```diff
@@ -36,6 +38,26 @@ class BaseStructureConfig(BaseConfig, ABC):
...
+    def structure(self) -> Optional[Structure]:
+        return self._structure
+
+    @structure.setter
+    def structure(self, structure: Structure) -> None:
+        self._structure = structure
+
+        event_listener = EventListener(self.structure.publish_event)
+
+        self.prompt_driver.add_event_listener(event_listener)
```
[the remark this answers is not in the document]
collindutter: Added some logic/tests for this.

griptape/config/base_structure_config.py
collindutter: I'm not really sure. `merge_config` has all sorts of [issues](https://github.com/griptape-ai/griptape/issues/826) and should probably be removed.

dylanholmes (approved):
So much better!

PR #987, opened 2024-07-16
Add RouteLLM Prompt Driver

collindutter:

## Describe your changes
Adds `RouteLLMPromptDriver` for using [RouteLLM](https://github.com/lm-sys/RouteLLM/tree/main/routellm) to route between a strong Prompt Driver and weak Prompt Driver. 
NA

----
📚 Documentation preview 📚: https://griptape--987.org.readthedocs.build//987/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Native function calling support to `OpenAiChatPromptDriver`, `AzureOpenAiChatPromptDriver`, `AnthropicPromptDriver`, `AmazonBedrockPromptDriver`, `GooglePromptDriver`, and `CoherePromptDriver`.
 - `OllamaEmbeddingDriver` for generating embeddings with Ollama.
 - `GriptapeCloudKnowledgeBaseVectorStoreDriver` to query Griptape Cloud Knowledge Bases.
+- `RouteLLMPromptDriver` for using RouteLLM to route between a strong Prompt Driver and weak Prompt Driver. 
 
 ### Changed
 
--- a/griptape/drivers/prompt/routellm_prompt_driver.py
+++ b/griptape/drivers/prompt/routellm_prompt_driver.py
@@ -1,0 +1,103 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from attrs import Factory, define, field
+
+from griptape.drivers import BasePromptDriver
+from griptape.tokenizers.dummy_tokenizer import DummyTokenizer
+from griptape.utils.import_utils import import_optional_dependency
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from routellm.controller import Controller
+
+    from griptape.common import (
+        DeltaMessage,
+        Message,
+        PromptStack,
+    )
+    from griptape.tokenizers import BaseTokenizer
+
+
+@define
+class RouteLlmPromptDriver(BasePromptDriver):
+    """[RouteLlm](https://github.com/lm-sys/RouteLLM) Prompt Driver.
+
+    Attributes:
+        strong_prompt_driver: Prompt Driver to use when routing to the strong model.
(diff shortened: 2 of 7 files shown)


collindutter:
Should we consider other routing solutions before putting this in place?

griptape/drivers/prompt/routellm_prompt_driver.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if routed_model == self.strong_prompt_driver.model:
+            prompt_driver = self.strong_prompt_driver
+        elif routed_model == self.weak_prompt_driver.model:
+            prompt_driver = self.weak_prompt_driver
+        else:
+            raise ValueError(f"Model '{routed_model}' not found.")
+
+        self.model = prompt_driver.model
+        self.tokenizer = prompt_driver.tokenizer
```
collindutter, line 101: This feels a little weird, but I'm not sure what the alternative is

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Native function calling support to `OpenAiChatPromptDriver`, `AzureOpenAiChatPromptDriver`, `AnthropicPromptDriver`, `AmazonBedrockPromptDriver`, `GooglePromptDriver`, and `CoherePromptDriver`.
 - `OllamaEmbeddingDriver` for generating embeddings with Ollama.
 - `GriptapeCloudKnowledgeBaseVectorStoreDriver` to query Griptape Cloud Knowledge Bases.
+- `RouteLLMPromptDriver` for using RouteLLM to route between a strong Prompt Driver and weak Prompt Driver.
```
dylanholmes, line 12: In response to 

> Should we consider other routing solutions before putting this in place?

If you were referencing the choice of making a "composite/decorator" Prompt Driver: I really like the decorator pattern for this scenario. It seems ideal to allow users to use a familiar and existing api (Prompt Driver), especially since LLM routing would primarily be used as a mechanism for reducing costs that would be introduced after the initial business logic. Like first I'd implement my application and get it working, then I'd say wait I want to reduce my LLM costs. At that point all they need to do is swap out the Prompt Driver. Sounds like a good deal!

If you were referencing the choice of `RouteLLM` vs something else, you did put `RouteLLM` in the name, so it seems like nothing is preventing us from adding another Prompt Driver that does LLM routing differently.

griptape/drivers/prompt/routellm_prompt_driver.py
```diff
@@ -0,0 +1,103 @@
...
+        kw_only=True,
+        metadata={"serializable": False},
+    )
+    weak_prompt_driver: BasePromptDriver = field(
+        kw_only=True,
+        metadata={"serializable": False},
+    )
+    router: str = field(kw_only=True, default="mf", metadata={"serializable": True})
+    threshold: float = field(kw_only=True, metadata={"serializable": True})
+    client: Controller = field(
```
dylanholmes, line 48: I think you should name this field `controller` instead of `client`. When I read `client` I thought this thing was actually calling a service, but its not, its all local.

griptape/drivers/prompt/routellm_prompt_driver.py
dylanholmes: I agree that this kind of mutation inside of a `_get_prompt_driver` method is weird.

I think that `model` should be fixed to `route-llm`. Otherwise you'll get duplicate `StartPromptEvent` and `FinishPromptEvent` for the underlying `model`. (Like if you route to `gpt-4o`, then `RouteLlmPromptDriver` will emit `StartPromptEvent(model="gpt-4o")` and the underlying `OpenAiChatPromptDriver` will also do this.)

I think `tokenizer` should be left as `None`. Is there any reason why it can't be? I couldn't find a reason after searching for places where the field is referenced.

dylanholmes (commented):
Really cool! 🍦

collindutter:
Closing for now based on offline discussion.

PR #998, opened 2024-07-18
Datadog observability driver

cjkindel:

## Describe your changes
Add a Datadog observability driver that pushes observability traces to a Datadog agent.

### DIFF
--- a/griptape/drivers/observability/datadog_observability_driver.py
+++ b/griptape/drivers/observability/datadog_observability_driver.py
@@ -1,0 +1,17 @@
+from attrs import Factory, define, field
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.trace import SpanProcessor
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+from griptape.drivers.observability.open_telemetry_observability_driver import OpenTelemetryObservabilityDriver
+
+
+@define
+class DatadogObservabilityDriver(OpenTelemetryObservabilityDriver):
+    span_processor: SpanProcessor = field(
+        default=Factory(
+            lambda self: BatchSpanProcessor(OTLPSpanExporter(endpoint="http://localhost:4318/v1/traces")),
+            takes_self=True,
+        ),
+        kw_only=True,
+    )
--- a/griptape/drivers/observability/datadog_observability_driver.py
+++ b/griptape/drivers/observability/datadog_observability_driver.py
@@ -15,7 +15,7 @@ class DatadogObservabilityDriver(OpenTelemetryObservabilityDriver):
     )
     span_processor: SpanProcessor = field(
         default=Factory(
-            lambda self: BatchSpanProcessor(OTLPSpanExporter(endpoint=self.base_url)),
+            lambda self: BatchSpanProcessor(OTLPSpanExporter(endpoint=self.datadog_agent_endpoint)),
             takes_self=True,
         ),
         kw_only=True,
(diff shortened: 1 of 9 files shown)


griptape/drivers/observability/datadog_observability_driver.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@define
+class DatadogObservabilityDriver(OpenTelemetryObservabilityDriver):
+    datadog_agent_endpoint: str = field(
+        default=Factory(lambda: os.getenv("DD_AGENT_ENDPOINT", "http://localhost:4318")), kw_only=True
+    )
+    span_processor: SpanProcessor = field(
+        default=Factory(
+            lambda self: BatchSpanProcessor(OTLPSpanExporter(endpoint=self.datadog_agent_endpoint + "/v1/traces")),
```
collindutter, line 18: Use f-string instead of string concatenation.

griptape/drivers/observability/open_telemetry_observability_driver.py
```diff
@@ -18,18 +18,28 @@
...
             takes_self=True,
         ),
         kw_only=True,
     )
     _tracer: Optional[Tracer] = None
     _root_span_context_manager: Any = None
 
+    def _trace_provider_factory(self) -> TracerProvider:
+        attributes = {"service.name": self.service_name}
+        if self.service_version:
```
collindutter, line 37: Use explicit `is not None` check

griptape/drivers/observability/open_telemetry_observability_driver.py
```diff
@@ -18,18 +18,28 @@
...
+            attributes["service.version"] = self.service_version
+        if self.deployment_env:
```
collindutter, line 39: `is not None`

collindutter (changes requested):
Overall looks pretty good! I love how little work was needed to extend @dylanholmes's existing observability work.

collindutter (dismissed):
Approved but would still like @dylanholmes's eyes before merging.

griptape/drivers/observability/open_telemetry_observability_driver.py
```diff
@@ -18,18 +18,28 @@
 
 @define
 class OpenTelemetryObservabilityDriver(BaseObservabilityDriver):
-    service_name: str = field(kw_only=True)
+    service_name: str = field(default="griptape", kw_only=True)
```
dylanholmes, line 21: Small refactoring opportunity: If we have a default here, then we don't really need it in `GriptapeObservabilityDriver` anymore (which also mean that we need to override `trace_provider` there too). 

Though its fine by me if you don't want to do it in this PR

dylanholmes (approved):
Awesome!

Just one optional refactor opportunity

griptape/drivers/observability/open_telemetry_observability_driver.py
cjkindel: I assumed GriptapeCloud used a non-attribute trace_provider on purpose. If we are OK with GriptapeCloud using the OpenTelemetry tracer_provider, then happy to. Just not clear why it used its own trace_provider in the first place

dylanholmes: It was just added to work around some attrs awkwardness. The args were left out because gtc doesn't really need them, but they don't hurt either

dylanholmes (dismissed):
Nice work! Thanks for the small refactor too

dylanholmes (approved):

Issue #823, opened 2024-06-04
Add a `date_range` parameter to the WebSearch tool


shhlife:
- [x ] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
One of our customers would like the ability to specify a date_range within web search. I provided an example how to do it with rules, and how to modify the WebSearch tool to do it - but it would be nice to have it as part of the framework.

**Describe the solution you'd like**
An optional parameter to the WebSearch tool to include a date range. 

The Google search api requires it to be in the format:
`&sort=date:r:YYYYMMDD:YYYYMMDD`

For example:
`&sort=date:20240101:20240401`

To add it to the tool, I added an optional parameter for date_range:
```
                    schema.Optional(Literal(
                        "date_range",
                        description="Optional date range to search within. Format: YYYYMMDD:YYYYMMDD",
                    )): str

```

And then added some logic to the `search` method to use it:

```
    def search(self, props: dict) -> ListArtifact | ErrorArtifact:
        query = props["values"]["query"]
        date_range = props["values"].get("date_range", None)
        if date_range:
            query +=f"&sort=date:r:{date_range}"
        try:
            return ListArtifact([TextArtifact(str(result)) for result in self._search_google(query)])
        except Exception as e:
            return ErrorArtifact(f"error searching Google: {e}")
```

**Describe alternatives you've considered**
Tried adding it as a ruleset as well - that worked but didn't feel as solid.

```
agent = Agent(
    rulesets=[search_ruleset],
    rules=[
        Rule(
            "When asked to search within a date range, use the `date_range` parameter of the Search Client. Example: 20210101:20220101"
        )
    ],
    tools=[search_client],
)
```

shhlife:
With the latest update in 0.28.0, is it possible to take a look at this one again? Would it need to be on the driver?

collindutter:
Maybe -- do you want the LLM to come up with the date? Or do you want to set it yourself?

collindutter:
Closed in #1015

PR #1015, opened 2024-07-24
Add additional param support to WebSearch

collindutter:

## Describe your changes
Added
- Ability to set custom schema properties on Tool Activities via `extra_schema_properties`.

Changed
- **BREAKING**: `api_key` and `search_id` are now required fields in `GoogleWebSearchDriver`.
- `GoogleWebSearchDriver` and `DuckDuckGoWebSearchDriver` now use `kwargs` passed to the `run` method.
#823 

## Example
```python
import os
import schema
from griptape.drivers import GoogleWebSearchDriver
from griptape.tools import WebSearch


agent = Agent(
    tools=[
        WebSearch(
            web_search_driver=GoogleWebSearchDriver(
                api_key=os.environ["GOOGLE_API_KEY"],
                search_id=os.environ["GOOGLE_API_SEARCH_ID"],
            ),
            extra_schema_properties={
                "search": {
                    schema.Literal(
                        "sort",
                        description="Date range to search within. Format: date:r:YYYYMMDD:YYYYMMDD",
                    ): str
                }
            },
        )
    ],
)

agent.run("Search for articles about the history of the internet from 1990 to 2000")

```

```
[07/25/24 16:12:36] INFO     ToolkitTask c1f3c42a45e64167a597f2ec8b84097f
                             Input: Search for articles about the history of the internet from 1990 to 2000
[07/25/24 16:12:37] INFO     Subtask d93cdd7289f447d2aed929628a33046d
                             Actions: [
                               {
                                 "tag": "call_9B2ngzqoHA4jHAlKJNNc3q7O",
                                 "name": "WebSearch",
                                 "path": "search",
                                 "input": {
                                   "values": {
                                     "query": "history of the internet",
                                     "sort": "date:r:19900101:20001231"
                                   }
                                 }
                               }
                             ]
```

### DIFF
--- a/griptape/drivers/web_search/duck_duck_go_web_search_driver.py
+++ b/griptape/drivers/web_search/duck_duck_go_web_search_driver.py
@@ -19,7 +19,9 @@ class DuckDuckGoWebSearchDriver(BaseWebSearchDriver):
 
     def search(self, query: str, **kwargs) -> ListArtifact:
         try:
-            results = self.client.text(query, region=f"{self.language}-{self.country}", max_results=self.results_count)
+            results = self.client.text(
+                query, region=f"{self.language}-{self.country}", max_results=self.results_count, **kwargs
+            )
             return ListArtifact(
                 [
                     TextArtifact(
--- a/griptape/drivers/web_search/google_web_search_driver.py
+++ b/griptape/drivers/web_search/google_web_search_driver.py
@@ -11,24 +11,24 @@ from griptape.drivers import BaseWebSearchDriver
 
 @define
 class GoogleWebSearchDriver(BaseWebSearchDriver):
-    api_key: str = field(default=None, kw_only=True)
-    search_id: str = field(default=None, kw_only=True)
+    api_key: str = field(kw_only=True)
+    search_id: str = field(kw_only=True)
 
     def search(self, query: str, **kwargs) -> ListArtifact:
         return ListArtifact([TextArtifact(json.dumps(result)) for result in self._search_google(query, **kwargs)])
 
     def _search_google(self, query: str, **kwargs) -> list[dict]:
-        url = (
-            f"https://www.googleapis.com/customsearch/v1?"
-            f"key={self.api_key}&"
-            f"cx={self.search_id}&"
-            f"q={query}&"
-            f"start=0&"
-            f"lr=lang_{self.language}&"
-            f"num={self.results_count}&"
-            f"gl={self.country}"
-        )
-        response = requests.get(url)
+        query_params = {
+            "key": self.api_key,
+            "cx": self.search_id,
+            "q": query,
+            "start": 0,
+            "lr": f"lang_{self.language}",
+            "num": self.results_count,
+            "gl": self.country,
+            **kwargs,
+        }
+        response = requests.get("https://www.googleapis.com/customsearch/v1", params=query_params)
 
         if response.status_code == 200:
             data = response.json()
--- a/griptape/mixins/activity_mixin.py
+++ b/griptape/mixins/activity_mixin.py
@@ -3,16 +3,24 @@ from __future__ import annotations
 import inspect
 from typing import Callable, Optional
 
-import schema
 from attrs import Attribute, define, field
 from jinja2 import Template
-from schema import Literal, Schema
+from schema import Schema
 
 
 @define(slots=False)
 class ActivityMixin:
+    """Provides Tool Activity management functionality to Tools.
+
+    Attributes:
+        allowlist: List of Tool Activities to include in the Tool schema.
+        denylist: List of Tool Activities to remove from the Tool schema.
+        extra_schema_properties: Mapping of Activity name and extra properties to include in the activity's schema.
+    """
+
     allowlist: Optional[list[str]] = field(default=None, kw_only=True)
     denylist: Optional[list[str]] = field(default=None, kw_only=True)
+    extra_schema_properties: dict = field(default=None, kw_only=True)
 
     @allowlist.validator  # pyright: ignore[reportAttributeAccessIssue]
     def validate_allowlist(self, _: Attribute, allowlist: Optional[list[str]]) -> None:
@@ -80,20 +88,15 @@ class ActivityMixin:
     def activity_schema(self, activity: Callable) -> Optional[Schema]:
         if activity is None or not getattr(activity, "is_activity", False):
             raise Exception("This method is not an activity.")
-        elif getattr(activity, "config")["schema"]:
-            full_schema = {
-                "values": getattr(activity, "config")["schema"].schema if getattr(activity, "config")["schema"] else {},
-            }
+        elif getattr(activity, "config")["schema"] is not None:
+            full_schema = {"values": getattr(activity, "config")["schema"]}
+            activity_name = self.activity_name(activity)
+            if self.extra_schema_properties is not None and activity_name in self.extra_schema_properties:
+                full_schema["values"].schema.update(self.extra_schema_properties[activity_name])
 
             return Schema(full_schema)
         else:
             return None
-
-    def activity_to_input(self, activity: Callable) -> dict:
-        if self.activity_schema(activity):
-            return {Literal("input"): {"values": getattr(activity, "config")["schema"]}}
-        else:
-            return {schema.Optional("input"): {}}
 
     def _validate_tool_activity(self, activity_name: str) -> None:
         tool = self.__class__
--- a/griptape/mixins/activity_mixin.py
+++ b/griptape/mixins/activity_mixin.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import inspect
+from copy import deepcopy
 from typing import Callable, Optional
 
 from attrs import Attribute, define, field
@@ -20,7 +21,7 @@ class ActivityMixin:
 
     allowlist: Optional[list[str]] = field(default=None, kw_only=True)
     denylist: Optional[list[str]] = field(default=None, kw_only=True)
-    extra_schema_properties: dict[str, dict] = field(default=None, kw_only=True)
+    extra_schema_properties: Optional[dict[str, dict]] = field(default=None, kw_only=True)
 
     @allowlist.validator  # pyright: ignore[reportAttributeAccessIssue]
     def validate_allowlist(self, _: Attribute, allowlist: Optional[list[str]]) -> None:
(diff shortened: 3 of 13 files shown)


collindutter:
Note that in the example I had to _disable_ `use_native_tools` here. With the default of `use_native_tools=True`, OpenAi refused to use the `params` portion of the schema. Anthropic also had no issue which makes me think there is a lack of chain-of-thoughting happening behind the scenes in OpenAi's function calling.

I tried a few solutions to no avail:
- Simplify the Tool schema by removing `values`
- Add an `explanation` field to the schema as proposed [here](https://pierce-lamb.medium.com/improving-gpt-4-function-calling-with-an-explanation-parameter-4fba06a4c6bb). This _kind of_ worked, but it still took a few tries each time.  

This has me wondering whether `use_native_tools=True` is a good default for `OpenAiChatPromptDriver`.

griptape/drivers/web_search/google_web_search_driver.py
```diff
@@ -11,24 +11,24 @@
 
 @define
 class GoogleWebSearchDriver(BaseWebSearchDriver):
-    api_key: str = field(default=None, kw_only=True)
-    search_id: str = field(default=None, kw_only=True)
+    api_key: str = field(kw_only=True)
+    search_id: str = field(kw_only=True)
```
dylanholmes, line 15: Why were they not optional before? Did it just result in a runtime error?

griptape/tools/web_search/tool.py
```diff
@@ -15,25 +15,30 @@
 
 @define
 class WebSearch(BaseTool):
-    web_search_driver: BaseWebSearchDriver = field(default=None, kw_only=True)
+    web_search_driver: BaseWebSearchDriver = field(kw_only=True)
```
dylanholmes, line 18: This seems like a good change. Doesn't seem like it'd work without a driver...

griptape/tools/web_search/tool.py
```diff
@@ -15,25 +15,30 @@
...
 
     @activity(
         config={
-            "description": "Can be used for searching the web",
+            "description": "Can be used for searching the web via the {{ _self.web_search_driver.__class__.__name__}}.",
```
dylanholmes, line 22: Might be nice if the web search drivers had a method that returns the name of search engine. I assume that LLMs might like it a little better than this class name.

griptape/tools/web_search/tool.py
```diff
@@ -15,25 +15,30 @@
...
             "schema": Schema(
                 {
                     Literal(
                         "query",
                         description="Search engine request that returns a list of pages with titles, descriptions, and URLs",
                     ): str,
+                    Optional(
+                        "params",
+                        description="Additional params that can be passed with the request.",
```
dylanholmes, line 31: Not sure if it would help with the issues you were having, but maybe you could elaborate a little more on examples of what "params" is used for?

dylanholmes (dismissed):
Nice job, this seems like an improvement despite the trouble.

Just a couple of clarifying questions plus a couple optional ideas

griptape/drivers/web_search/google_web_search_driver.py
collindutter: Not sure why they were optional, as far as I can tell it just did not work without them.

griptape/tools/web_search/tool.py
collindutter: Yeah I was thinking how far to take this. A non-Driver name is probably better.

griptape/tools/web_search/tool.py
collindutter: Basically anything that you might want the LLM to pass at runtime. The date range example is what prompted this change, but in theory you could prompt the LLM to use any [of these](https://developers.google.com/custom-search/v1/reference/rest/v1/cse/list).

shhlife:
@collindutter in your example, you show `use_native_tools` to True, but it needs to be false for this to work correctly, right?

Also, it's unfortunate that you still need to use a rule to get the driver to do the right thing - is this something that can be owned on the driver side instead of the agent side?

It makes it so the user needs to do some prompt engineering if they're switching between drivers to get it to work correctly.

cheers,
Jason

collindutter:
@shhlife take another look at the new approach.

griptape/mixins/activity_mixin.py
```diff
@@ -3,16 +3,24 @@
...
+
+    Attributes:
+        allowlist: List of Tool Activities to include in the Tool schema.
+        denylist: List of Tool Activities to remove from the Tool schema.
+        extra_schema_properties: Mapping of Activity name and extra properties to include in the activity's schema.
+    """
+
     allowlist: Optional[list[str]] = field(default=None, kw_only=True)
     denylist: Optional[list[str]] = field(default=None, kw_only=True)
+    extra_schema_properties: dict = field(default=None, kw_only=True)
```
collindutter, line 23: Not in love with this name

griptape/tools/base_tool.py
```diff
@@ -95,18 +96,23 @@ def schema(self) -> dict:
...
+
+            activity_schema = self.activity_schema(activity)
+            if activity_schema is not None:
+                schema_dict[Literal("input")] = activity_schema.schema
+            else:
+                schema_dict[schema.Optional("input")] = {}
+
+            schemas.append(Schema(schema_dict))
+
+        return schemas
```
collindutter, line 115: Faced some really weird type errors when doing the dictionary unpacking. Refactoring to imperative fixed.

griptape/tools/web_search/tool.py
```diff
@@ -31,9 +31,11 @@ class WebSearch(BaseTool):
         },
     )
     def search(self, props: dict) -> ListArtifact | ErrorArtifact:
-        query = props["values"]["query"]
+        values = props["values"]
+        query = values["query"]
+        extra_keys = {k: values[k] for k in values.keys() - {"query"}}
```
collindutter, line 36: Should the extra keys be put in a separate field from `values` so that we don't need to do this?

griptape/tools/web_search/tool.py
vachillo: strip out the `WebSearchDriver` part of the name?

griptape/mixins/activity_mixin.py
vachillo: is it worth updating to a wrapper `Activity` class that holds this information? schema, extra props, etc.

griptape/tools/web_search/tool.py
vachillo: i think it still makes sense to have everything in `values`

griptape/mixins/activity_mixin.py
vachillo: also, does this contain descriptions about what the properties are? especially with the query params for search engines, they are basically just random characters that an LLM would not be able to figure out how to use.

griptape/drivers/web_search/duck_duck_go_web_search_driver.py
```diff
@@ -19,7 +19,9 @@ class DuckDuckGoWebSearchDriver(BaseWebSearchDriver):
 
     def search(self, query: str, **kwargs) -> ListArtifact:
         try:
-            results = self.client.text(query, region=f"{self.language}-{self.country}", max_results=self.results_count)
+            results = self.client.text(
+                query, region=f"{self.language}-{self.country}", max_results=self.results_count, **kwargs
+            )
```
vachillo, line 24: the plan here is to just leave it up to the user to figure out how to use the driver? not provide a few top-level fields that the drivers need to implement?

griptape/tools/web_search/tool.py
dylanholmes: Oh, I didn't mean elaborate for me, I meant for the LLM, by making the description less generic

griptape/mixins/activity_mixin.py
collindutter: Yeah we definitely should have an `Activity` type vs cramming extra attributes into `Callable`. I created [an issue](https://github.com/griptape-ai/griptape/issues/1023).

> Also, does this contain descriptions about what the properties are?

Take a look at the updated PR description. I just realized I was missing an example of how this change gets used.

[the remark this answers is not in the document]
collindutter: Take a look at the (previously missing) example, the API takes the _value_ of a `Schema` object which is [hard to type](https://github.com/keleshev/schema/blob/master/schema/__init__.py#L358). Should we take in a `Schema` instead? The reason I avoided this is because we're only using the value of the Schema -- throwing away things like `Schema.description`.

Either way,  I narrowed the scope a bit to `dict[str, dict]`

griptape/tools/base_tool.py
collindutter: See https://github.com/griptape-ai/griptape/pull/1015/files#r1695354991, out of scope for this PR but definitely something we should do.

griptape/tools/base_tool.py
```diff
@@ -95,18 +96,23 @@ def schema(self) -> dict:
...
+            schema_dict: dict[Literal | schema.Optional, Any] = {
+                Literal("name"): self.name,
+                Literal("path", description=self.activity_description(activity)): self.activity_name(activity),
+            }
+
+            activity_schema = self.activity_schema(activity)
+            if activity_schema is not None:
+                schema_dict[Literal("input")] = activity_schema.schema
+            else:
+                schema_dict[schema.Optional("input")] = {}
```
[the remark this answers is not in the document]
collindutter: Explanation is [here](https://github.com/griptape-ai/griptape/pull/883)

griptape/tools/web_search/tool.py
vachillo: could update the decorator to automatically unwrap `values` from `params`. since it looks like every tool has this pattern.
`def search(self, props: dict, values: dict) -> ListArtifact | ErrorArtifact:`

collindutter: I don't want to change the method signature for Tool Activities, that will break all existing Tools. I'm sure we could inspect the method parameters but that could get messy.

I think we either change the behavior so extra values are added in `props["extra_values"]` or we do this.

griptape/tools/base_tool.py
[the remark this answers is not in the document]
dylanholmes: Might be nice to add a comment in the code to make it more obvious (plus it wasn't clear to me from that PR which problem this was solving, the explanation in the PR seemed mostly related to the prompting change...)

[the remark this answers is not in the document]
dylanholmes: Might be nice to add a comment in the code to make it more obvious (plus it wasn't clear to me from that PR which problem this was solving, the explanation in the PR seemed mostly related to the prompting change...)

dylanholmes (approved):
Nice work!

Just one last comment about a comment

dylanholmes (approved):
Great work!

griptape/tools/base_tool.py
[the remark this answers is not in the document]
dylanholmes: Thanks!

PR #1018, opened 2024-07-25
Stable Diffusion 3 local support

andrewfrench:
- [x] Docs
- [x] Tests

## Describe your changes

This PR introduces drivers that can be used to generate images using Stable Diffusion 3 locally.

A model-agnostic HuggingFaceDiffusionPipelineImageGenerationDriver manages creating and running inferences on a HuggingFace diffusers pipeline. New model drivers: StableDiffusion3PipelineImageGenerationModelDriver, StableDiffusion3Img2ImgPipelineImageGenerationModelDriver, and StableDiffusion3ControlNetPipelineImageGenerationDriver extend the BaseDiffusionPipelineImageGenerationModelDriver to specify how to prepare the inference pipeline and format pipeline inputs.


griptape/drivers/image_generation_model/stable_diffusion_3_img_2_img_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,67 @@
...
+        sd3_img2img_pipeline = import_optional_dependency(
+            "diffusers.pipelines.stable_diffusion_3.pipeline_stable_diffusion_3_img2img"
+        ).StableDiffusion3Img2ImgPipeline
+        if os.path.isfile(model):
+            # If the model provided is a local file (not a directory),
+            # we load it using the from_single_file method.
+
+            raise NotImplementedError(
+                "StableDiffusion3Img2ImgPipeline does not yet support loading from a single file."
+            )
```
andrewfrench, line 35: For eventual ComfyUI convenience, we accept three model input types: 
- path to a single file containing a model
- path to a directory containing model files
- HuggingFace model repo name

The one exception to this is the StableDiffusion3Img2ImgPipeline, which doesn't support `.from_single_file()`. Models can still be loaded by path to a local directory or by model repo name (and not downloaded again if they're cached locally).

griptape/drivers/image_generation/huggingface_diffusion_pipeline_image_generation_driver.py
```diff
@@ -0,0 +1,88 @@
...
+
+from attrs import define, field
+
+from griptape.artifacts import ImageArtifact
+from griptape.drivers import BaseDiffusionPipelineImageGenerationModelDriver, BaseImageGenerationDriver
+from griptape.utils import import_optional_dependency
+
+
+@define
+class HuggingFaceDiffusionPipelineImageGenerationDriver(BaseImageGenerationDriver, ABC):
```
collindutter, line 15: Should this subclass `BaseImageGenerationDriver` instead?

collindutter, line 15: Wondering if we should just rename to `HuggingFacePipelineImageGenerationDriver` for similarity with the `HuggingFacePipelinePromptDriver` that uses `transformers`.

griptape/drivers/image_generation_model/stable_diffusion_3_controlnet_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Attributes:
+        controlnet_model: The ControlNet model to use for image generation.
+        controlnet_conditioning_scale: The conditioning scale for the ControlNet model. Defaults to None.
+    """
+
+    controlnet_model: str = field(kw_only=True)
+    controlnet_conditioning_scale: Optional[float] = field(default=None, kw_only=True, metadata={"serializable": True})
+
+    def prepare_pipeline(self, model: str, device: Optional[str]) -> Any:
```
collindutter, line 34: Are there any types we can use here?

griptape/drivers/image_generation_model/stable_diffusion_3_controlnet_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,85 @@
...
+
+from griptape.drivers import StableDiffusion3PipelineImageGenerationModelDriver
+from griptape.utils import import_optional_dependency
+
+if TYPE_CHECKING:
+    from PIL.Image import Image
+else:
+    StableDiffusion3ControlNetPipeline = import_optional_dependency(
+        "diffusers.pipelines.controlnet_sd3.pipeline_stable_diffusion_3_controlnet"
+    ).StableDiffusion3ControlNetPipeline
```
collindutter, line 16: Can't have this import here as it'll break bare installs.

griptape/drivers/image_generation/huggingface_diffusion_pipeline_image_generation_driver.py
```diff
@@ -0,0 +1,88 @@
...
+    def try_text_to_image(self, prompts: list[str], negative_prompts: Optional[list[str]] = None) -> ImageArtifact:
+        pipeline = self.model_driver.prepare_pipeline(self.model, self.device)
+
+        prompt = ", ".join(prompts)
+        output_image = pipeline(
+            prompt, **self.model_driver.make_additional_params(negative_prompts, self.device)
+        ).images[0]
+
+        buffer = io.BytesIO()
+        output_image.save(buffer, format="PNG")
```
collindutter, line 38: Other formats besides "PNG"? Maybe move to field?

griptape/drivers/image_generation_model/base_diffusion_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,24 @@
...
+from typing import TYPE_CHECKING, Any, Optional
+
+from attrs import define
+
+if TYPE_CHECKING:
+    from PIL.Image import Image
+
+
+@define
+class BaseDiffusionPipelineImageGenerationModelDriver(ABC):
```
collindutter, line 13: Do we need a new base class? Can we instead use `BaseImageGenerationModelDriver`?

griptape/drivers/image_generation/huggingface_diffusion_pipeline_image_generation_driver.py
```diff
@@ -0,0 +1,88 @@
...
+            value=buffer.getvalue(), format="png", height=output_image.height, width=output_image.width, prompt=prompt
+        )
+
+    def try_image_variation(
+        self, prompts: list[str], image: ImageArtifact, negative_prompts: Optional[list[str]] = None
+    ) -> ImageArtifact:
+        pipeline = self.model_driver.prepare_pipeline(self.model, self.device)
+
+        prompt = ", ".join(prompts)
+        input_image = import_optional_dependency("PIL.Image").open(io.BytesIO(image.value))
```
collindutter, line 50: Imports should happen at the top of the function as if it were a regular `import`.

griptape/drivers/image_generation_model/stable_diffusion_3_controlnet_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,85 @@
...
+        controlnet_pipeline_params = {}
+        if self.torch_dtype is not None:
+            pipeline_params["torch_dtype"] = self.torch_dtype
+            controlnet_pipeline_params["torch_dtype"] = self.torch_dtype
+
+        # For both Stable Diffusion and ControlNet, models can be provided either
+        # as a path to a local file or as a HuggingFace model repo name.
+        # We use the from_single_file method if the model is a local file and the
+        # from_pretrained method if the model is a local directory or hosted on HuggingFace.
+        sd3_controlnet_model = import_optional_dependency("diffusers.models.controlnet_sd3").SD3ControlNetModel
```
collindutter, line 45: Imports should happen at the top of the function as if it were a regular `import`.

griptape/drivers/image_generation_model/stable_diffusion_3_controlnet_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,85 @@
...
+            )
+
+        else:
+            pipeline_params["controlnet"] = sd3_controlnet_model.from_pretrained(
+                self.controlnet_model, **controlnet_pipeline_params
+            )
+
+        sd3_controlnet_pipeline = import_optional_dependency(
+            "diffusers.pipelines.controlnet_sd3.pipeline_stable_diffusion_3_controlnet"
+        ).StableDiffusion3ControlNetPipeline
```
collindutter, line 58: Imports should happen at the top of the function as if it were a regular `import`.

griptape/drivers/image_generation_model/stable_diffusion_3_controlnet_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,85 @@
...
+        if os.path.isfile(model):
+            pipeline = sd3_controlnet_pipeline.from_single_file(model, **pipeline_params)
+
```
collindutter, line 61: Weird that the autoformatter didn't format away this whitespace

griptape/drivers/image_generation_model/base_diffusion_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,24 @@
...
+    @abstractmethod
+    def prepare_pipeline(self, model: str, device: Optional[str]) -> Any: ...
+
+    @abstractmethod
+    def make_image_param(self, image: Optional[Image]) -> Optional[dict[str, Image]]: ...
```
collindutter, line 18: Why is image optional if every implementation seems to require it?

griptape/drivers/image_generation_model/stable_diffusion_3_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,94 @@
...
+            pipeline = sd3_pipeline.from_pretrained(model, **pipeline_params)
+
+        # Move inference to particular device if requested.
+        if device is not None:
+            pipeline.to(device)
+
+        return pipeline
+
+    def make_image_param(self, image: Optional[Image]) -> Optional[dict[str, Image]]:
+        return None
```
collindutter, line 67: I guess this is why image is optional, but this feels odd.

griptape/drivers/image_generation/huggingface_diffusion_pipeline_image_generation_driver.py
andrewfrench: I assume you mean `BaseMultiModelImageGenerationDriver`. That was the original intention, but the Pipeline drivers require a substantially different interface than those that inherit from `BaseImageGenerationModelDriver` as required by the `BaseMultiModelImageGenerationDriver`.

griptape/drivers/image_generation_model/stable_diffusion_3_controlnet_pipeline_image_generation_model_driver.py
andrewfrench: Whoops, leftover test change. Removed.

griptape/drivers/image_generation_model/base_diffusion_pipeline_image_generation_model_driver.py
andrewfrench: The interface difference between these two model driver types is substantial. Drivers for remote models package inputs and unpackage outputs, drivers for remote models package inputs and prepare diffusion pipelines. We could certainly merge these together, but I think that would increase the overall awkwardness of this implementation.

griptape/drivers/image_generation_model/stable_diffusion_3_controlnet_pipeline_image_generation_model_driver.py
```diff
@@ -0,0 +1,85 @@
...
+from attrs import define, field
+
+from griptape.drivers import StableDiffusion3PipelineImageGenerationModelDriver
+from griptape.utils import import_optional_dependency
+
+if TYPE_CHECKING:
+    from PIL.Image import Image
+else:
+    StableDiffusion3ControlNetPipeline = import_optional_dependency(
```
collindutter, line 14: Having `3` in the file/class name feels odd, maybe let's remove?

andrewfrench: Stable Diffusion's pipelines include versions in their class names and are distinct enough that we would need a driver for each, much in the same way that ControlNet and Img2Img are distinct here. Amaru pointed out that Stable Diffusion 3 is still quite new and many people still prefer their Stable Diffusion XL workflows &mdash; to support this, we'd need another driver typed for StableDiffusionXL (with similar stories for SD1.5 and SD2).

griptape/drivers/image_generation_model/stable_diffusion_3_pipeline_image_generation_model_driver.py
andrewfrench: This is odd, and a half-measure. Zooming out, I think the time has come to admit the design where each driver implements a suite of common image generation types (prompt, variation, inpainting, outpainting) isn't working so well: forcing NotImplementedErrors and backwards workarounds like this. We can move this conversation to another venue, but should maybe consider making image generation drivers more granular.

griptape/drivers/image_generation_model/stable_diffusion_3_controlnet_pipeline_image_generation_model_driver.py
andrewfrench: I'm mulling this one over, locking this down to one Pipeline type per instance was the original goal. The `.from_single_file()` and `.from_pretrained()` methods return a long list of inferred types, and I encountered some nasty conflicts when patching `isinstance()` to assert that we're returning the expected pipeline type.

griptape/drivers/image_generation_model/base_diffusion_pipeline_image_generation_model_driver.py
collindutter: Understood. What do you think about renaming to `BaseDiffusersImageGenerationModelDriver` since it's very specific to the library? Or `BaseHuggingfacePipelineImageGenerationModelDriver`? We should aim for consistency with `HuggingfacePipelinePromptDriver` since that uses the sister library `tranformers`.

collindutter: Understood. Should any of these model driver classes be renamed with the new name for `HuggingFacePipelineImageGenerationDriver`?

griptape/drivers/image_generation_model/stable_diffusion_3_pipeline_image_generation_model_driver.py
collindutter: Can you please create a ticket to track this work? This feels like an important refactor pre-1.0.

andrewfrench:
It seemed like I was trying to fit a square peg (model drivers that were actually specific to pipeline types, not models) into a round hole (model drivers meant to define interactions with a model), so I refactored these out to a new type: `BaseImageGenerationPipelineDriver`, that encapsulates specific preparation and input for diffusion pipeline image generation flows.

collindutter (dismissed):

collindutter (approved):

PR #1019, opened 2024-07-25
Improve performance of image generation tools

andrewfrench:

## Describe your changes

This change primarily updates the schema descriptions of positive and negative prompts for image generation tools. Past experience has shown that models can be confused by the more complex prompt schema being replaced here. In cases where the wrong prompt schema was provided to a tool, the likely outcome was an empty prompt sent to the image generation model which would then return an irrelevant and undesirable image.

By replacing the more complex prompt schema (`{values: {prompts: [prompt1, prompt2, ...]}}`) with a single prompt string (`{values: {prompt: "prompt1, prompt2"}}`), models are more likely to correctly format activity inputs.

Other changes here include correctly loading input image artifacts from files after a previous update to how the ImageLoader.load() method works, factoring out some duplicated code, and improvements to schema descriptions of image generation tools, as well as pyright, linting, and formatting changes.

### DIFF
--- a/griptape/tools/base_image_generation_client.py
+++ b/griptape/tools/base_image_generation_client.py
@@ -1,0 +1,16 @@
+from attrs import define
+
+from griptape.mixins import BlobArtifactFileOutputMixin
+from griptape.tools import BaseTool
+
+
+@define
+class BaseImageGenerationClient(BlobArtifactFileOutputMixin, BaseTool):
+    """A base class for tools that generate images from text prompts."""
+
+    PROMPT_DESCRIPTION = "Features and qualities to include in the generated image, descriptive and succinct."
+    NEGATIVE_PROMPT_DESCRIPTION = (
+        "Features and qualities to avoid in the generated image. Affirmatively describe "
+        "what to avoid, for example: to avoid the color red, include 'red' "
+        "rather than 'no red'."
+    )
--- a/griptape/tools/inpainting_image_generation_client/tool.py
+++ b/griptape/tools/inpainting_image_generation_client/tool.py
@@ -1,14 +1,14 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, cast
+from pathlib import Path
+from typing import TYPE_CHECKING, cast
 
 from attrs import define, field
 from schema import Literal, Schema
 
 from griptape.artifacts import ErrorArtifact, ImageArtifact
 from griptape.loaders import ImageLoader
-from griptape.mixins import BlobArtifactFileOutputMixin
-from griptape.tools import BaseTool
+from griptape.tools.base_image_generation_client import BaseImageGenerationClient
 from griptape.utils.decorators import activity
 from griptape.utils.load_artifact_from_memory import load_artifact_from_memory
 
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
 
 
 @define
-class InpaintingImageGenerationClient(BlobArtifactFileOutputMixin, BaseTool):
+class InpaintingImageGenerationClient(BaseImageGenerationClient):
     """A tool that can be used to generate prompted inpaintings of an image.
 
     Attributes:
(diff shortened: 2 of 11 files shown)


griptape/tools/inpainting_image_generation_client/tool.py
```diff
@@ -31,17 +31,11 @@ class InpaintingImageGenerationClient(BlobArtifactFileOutputMixin, BaseTool):
...
-                    Literal(
-                        "prompts",
-                        description="A detailed list of features and descriptions to include in the generated image.",
-                    ): list[str],
-                    Literal(
-                        "negative_prompts",
-                        description="A detailed list of features and descriptions to avoid in the generated image.",
-                    ): list[str],
+                    Literal("prompt", description=BaseImageGenerationClient.PROMPT_DESCRIPTION): str,
+                    Literal("negative_prompt", description=BaseImageGenerationClient.NEGATIVE_PROMPT_DESCRIPTION): str,
```
dylanholmes, line 38: Should this be `self.PROMPT_DESCRIPTION`? Otherwise, doesn't seem like inheritance is needed.

dylanholmes (commented):
Overall looks good!

Just one question

griptape/tools/inpainting_image_generation_client/tool.py
andrewfrench: Inheritance is probably overkill here, but it's a place to standardize which classes image generation tools inherit from and, more importantly, provided a common space to define this repeated schema descriptions.

This class may not be necessary after addressing Collin's comment below.

griptape/tools/base_image_generation_client.py
```diff
@@ -0,0 +1,16 @@
...
+@define
+class BaseImageGenerationClient(BlobArtifactFileOutputMixin, BaseTool):
+    """A base class for tools that generate images from text prompts."""
+
+    PROMPT_DESCRIPTION = "Features and qualities to include in the generated image, descriptive and succinct."
+    NEGATIVE_PROMPT_DESCRIPTION = (
+        "Features and qualities to avoid in the generated image. Affirmatively describe "
+        "what to avoid, for example: to avoid the color red, include 'red' "
+        "rather than 'no red'."
+    )
```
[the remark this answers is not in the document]
collindutter: Nvm, this is fine since these are Tool descriptions.

griptape/tools/inpainting_image_generation_client/tool.py
collindutter: I think if they're defined as class fields, `BaseImageGenerationClient.PROMPT_DESCRIPTION` is the recommended way of accessing.

But that raises the question of whether they _should_ be class fields. I'm leaning towards adding instance fields (`prompt_description`, `negative_prompt_description`) with defaults pointing to class fields (`DEFAULT_PROMPT_DESCRIPTION`, `DEFAULT_NEGATIVE_PROMPT_DESCRIPTION`).

collindutter, line 38: Are you making it singular so that the LLM will rely on parallel actions if needs to generate multiple images?

collindutter (commented):
Overall looks good, just a couple comments!

griptape/tools/inpainting_image_generation_client/tool.py
andrewfrench: I'm probably mistaken, but we don't have access to the instance in the decorator, right? I'm not sure we can pull `self.prompt_description` if it's an instance field.

collindutter: Argh you're right. You _should_ have access in the schema properties, but today you can only access it [in the schema description](https://github.com/griptape-ai/griptape/blob/39d2b95dc0bb82f43506a872cea4f0f1caf65c18/tests/mocks/mock_tool/tool.py#L17).

Ok good with how it is today.

collindutter (approved):

griptape/tools/inpainting_image_generation_client/tool.py
andrewfrench: This doesn't change how many images are generated per invocation, the behavior remains to generate one image per activity call. The core of this change is that asking an Agent to provide one continuous [negative] prompt string in the schema produces better adherence to the schema than asking for a JSON list of "prompts" concatenated into a string by the tool.

PR #1022, opened 2024-07-26
Add DataStax Astra DB vector store driver

### LINKED ISSUE #1021: Add a vector store driver for DataStax Astra DB
- [X] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

DataStax' "Astra DB" database has first-class vector/ANN support and it would be desirable to have a vector store driver to enable using that backend from Griptape.

Implementation of an `AstraDBVectorStoreDriver` subclass of `BaseVectorStoreDriver`.

n/a

**Additional context**

I can provide the implementation.

hemidactylus:

## Describe your changes

Adds an `AstraDBVectorStoreDriver` class, plus:

- mkdocs.yaml
- changelog
- a new end-to-end example on the docs ("query a web page" with Astra DB for vector store)
- a specific new extra in the pyproject.yaml
- unit test
- integration test (requires secrets/connection params to an Astra database)

### Notes

I can run the integration testing on my machine (all green). Depending on Griptape's policy on integration testing, I can share secrets to a dedicated CI database or similar, just let me know if that is a valid option.

Also, I noticed that the number of returned vectors in the `query` method is generally called `count` (as I did and as other drivers do); but at least [one example](https://docs.griptape.ai/stable/griptape-tools/official-tools/rag-client/) in the docs seems to refer to it as `top_n`. Is it preferred that the method accepts both parameter names, perhaps?

Closes #1021

### DIFF
--- a/docs/examples/query-webpage-astra-db.md
+++ b/docs/examples/query-webpage-astra-db.md
@@ -9,9 +9,16 @@ Python script:
 ```python
 import os
 
-from griptape.drivers import AstraDBVectorStoreDriver, OpenAiChatPromptDriver, OpenAiEmbeddingDriver
+from griptape.drivers import (
+    AstraDBVectorStoreDriver,
+    OpenAiChatPromptDriver,
+    OpenAiEmbeddingDriver,
+)
 from griptape.engines.rag import RagEngine
-from griptape.engines.rag.modules import PromptResponseRagModule, VectorStoreRetrievalRagModule
+from griptape.engines.rag.modules import (
+    PromptResponseRagModule,
+    VectorStoreRetrievalRagModule,
+)
 from griptape.engines.rag.stages import ResponseRagStage, RetrievalRagStage
 from griptape.loaders import WebLoader
 from griptape.structures import Agent
@@ -19,7 +26,9 @@ from griptape.tools import RagClient, TaskMemoryClient
 
 if __name__ == "__main__":
     namespace = "datastax_blog"
-    input_blogpost = "www.datastax.com/blog/indexing-all-of-wikipedia-on-a-laptop"
+    input_blogpost = (
+        "www.datastax.com/blog/indexing-all-of-wikipedia-on-a-laptop"
+    )
 
     vector_store_driver = AstraDBVectorStoreDriver(
         embedding_driver=OpenAiEmbeddingDriver(),
@@ -53,7 +62,13 @@ if __name__ == "__main__":
         {namespace: WebLoader(max_tokens=256).load(input_blogpost)}
     )
 
-    vector_store_tool = RagClient(description="A DataStax blog post", rag_engine=engine)
+    vector_store_tool = RagClient(
+        description="A DataStax blog post",
+        rag_engine=engine,
+    )
     agent = Agent(tools=[vector_store_tool, TaskMemoryClient(off_prompt=False)])
-    agent.run(f"What engine made possible to index such an amount of data, and what kind of tuning was required?")
+    agent.run(
+        "What engine made possible to index such an amount of data, "
+        "and what kind of tuning was required?"
+    )
 ```
--- a/docs/examples/query-webpage-astra-db.md
+++ b/docs/examples/query-webpage-astra-db.md
@@ -1,0 +1,59 @@
+Required Griptape extras:
+
+```
+pip install griptape[drivers-vector-astra-db,drivers-web-scraper-trafilatura]
+```
+
+Python script:
+
+```python
+import os
+
+from griptape.drivers import AstraDBVectorStoreDriver, OpenAiChatPromptDriver, OpenAiEmbeddingDriver
+from griptape.engines.rag import RagEngine
+from griptape.engines.rag.modules import PromptResponseRagModule, VectorStoreRetrievalRagModule
+from griptape.engines.rag.stages import ResponseRagStage, RetrievalRagStage
+from griptape.loaders import WebLoader
+from griptape.structures import Agent
+from griptape.tools import RagClient, TaskMemoryClient
+
+if __name__ == "__main__":
+    namespace = "datastax_blog"
+    input_blogpost = "www.datastax.com/blog/indexing-all-of-wikipedia-on-a-laptop"
+
+    vector_store_driver = AstraDBVectorStoreDriver(
+        embedding_driver=OpenAiEmbeddingDriver(),
+        api_endpoint=os.environ["ASTRA_DB_API_ENDPOINT"],
+        token=os.environ["ASTRA_DB_APPLICATION_TOKEN"],
+        collection_name="griptape_test_collection",
+        astra_db_namespace=os.environ.get("ASTRA_DB_KEYSPACE"),
+        dimension=1536,
+    )
+
+    engine = RagEngine(
+        retrieval_stage=RetrievalRagStage(
+            retrieval_modules=[
+                VectorStoreRetrievalRagModule(
+                    vector_store_driver=vector_store_driver,
+                    query_params={
+                        "count": 2,
+                        "namespace": namespace,
+                    },
+                )
+            ]
+        ),
+        response_stage=ResponseRagStage(
+            response_module=PromptResponseRagModule(
+                prompt_driver=OpenAiChatPromptDriver(model="gpt-4o")
+            )
+        )
+    )
+
+    vector_store_driver.upsert_text_artifacts(
+        {namespace: WebLoader(max_tokens=256).load(input_blogpost)}
+    )
+
+    vector_store_tool = RagClient(description="A DataStax blog post", rag_engine=engine)
+    agent = Agent(tools=[vector_store_tool, TaskMemoryClient(off_prompt=False)])
+    agent.run(f"What engine made possible to index such an amount of data, and what kind of tuning was required?")
+```
--- a/docs/examples/query-webpage-astra-db.md
+++ b/docs/examples/query-webpage-astra-db.md
@@ -36,7 +36,6 @@ if __name__ == "__main__":
         token=os.environ["ASTRA_DB_APPLICATION_TOKEN"],
         collection_name="griptape_test_collection",
         astra_db_namespace=os.environ.get("ASTRA_DB_KEYSPACE"),
-        dimension=1536,
     )
 
     engine = RagEngine(
--- a/docs/griptape-framework/drivers/vector-store-drivers.md
+++ b/docs/griptape-framework/drivers/vector-store-drivers.md
@@ -511,7 +511,6 @@ vector_store_driver = AstraDBVectorStoreDriver(
     token=token,
     collection_name="astra_db_demo",
     astra_db_namespace=astra_db_namespace,
-    dimension=1536,
 )
 
 # Load Artifacts from the web
--- a/docs/griptape-framework/drivers/vector-store-drivers.md
+++ b/docs/griptape-framework/drivers/vector-store-drivers.md
@@ -482,3 +482,47 @@ values = [r.to_artifact().value for r in results]
 
 print("\n\n".join(values))
 ```
+
+### Astra DB
+
+!!! info
+    This Driver requires the `drivers-vector-astra-db` [extra](../index.md#extras).
+
+The AstraDBVectorStoreDriver supports [DataStax Astra DB](https://www.datastax.com/products/datastax-astra).
+
+The following example shows how to store vector entries and query the information using the driver:
+
+```python
+import os
+from griptape.drivers import AstraDBVectorStoreDriver, OpenAiEmbeddingDriver
+from griptape.loaders import WebLoader
+
+# Astra DB secrets and connection parameters
+api_endpoint = os.environ["ASTRA_DB_API_ENDPOINT"]
+token = os.environ["ASTRA_DB_APPLICATION_TOKEN"]
+astra_db_namespace = os.environ.get("ASTRA_DB_KEYSPACE")  # optional
+
+# Initialize an Embedding Driver.
+embedding_driver = OpenAiEmbeddingDriver(api_key=os.environ["OPENAI_API_KEY"])
+
+vector_store_driver = AstraDBVectorStoreDriver(
+    embedding_driver=embedding_driver,
+    api_endpoint=api_endpoint,
+    token=token,
+    collection_name="astra_db_demo",
+    astra_db_namespace=astra_db_namespace,
+    dimension=1536,
+)
+
+# Load Artifacts from the web
+artifacts = WebLoader().load("https://www.griptape.ai")
+
+# Upsert Artifacts into the Vector Store Driver
+[vector_store_driver.upsert_text_artifact(a, namespace="griptape") for a in artifacts]
+
+results = vector_store_driver.query(query="What is griptape?")
+
+values = [r.to_artifact().value for r in results]
+
+print("\n\n".join(values))
+```
--- a/docs/griptape-framework/drivers/vector-store-drivers.md
+++ b/docs/griptape-framework/drivers/vector-store-drivers.md
@@ -518,7 +518,10 @@ vector_store_driver = AstraDBVectorStoreDriver(
 artifacts = WebLoader().load("https://www.griptape.ai")
 
 # Upsert Artifacts into the Vector Store Driver
-[vector_store_driver.upsert_text_artifact(a, namespace="griptape") for a in artifacts]
+[
+    vector_store_driver.upsert_text_artifact(a, namespace="griptape")
+    for a in artifacts
+]
 
 results = vector_store_driver.query(query="What is griptape?")
 
--- a/griptape/drivers/vector/astra_db_vector_store_driver.py
+++ b/griptape/drivers/vector/astra_db_vector_store_driver.py
@@ -36,6 +36,9 @@ class AstraDBVectorStoreDriver(BaseVectorStoreDriver):
         token: a Database Token ("AstraCS:...") secret to access Astra DB.
         collection_name: the name of the collection on Astra DB.
         dimension: the number of components for embedding vectors. If not provided, it will be guessed from the embedding driver.
+        metric: the similarity metric to use, one of "dot_product", "euclidean" or "cosine".
+            If omitted, the server default ("cosine") will be used. See also values of `astrapy.constants.VectorMetric`.
+            If the vectors are normalized to unit norm, choosing "dot_product" over cosine yields up to 2x speedup in searches.
         astra_db_namespace: optional specification of the namespace (in the Astra database) for the data.
             *Note*: not to be confused with the "namespace" mentioned elsewhere, which is a grouping within this vector store.
     """
@@ -44,6 +47,7 @@ class AstraDBVectorStoreDriver(BaseVectorStoreDriver):
     token: str = field(kw_only=True, metadata={"serializable": False})
     collection_name: str = field(kw_only=True, metadata={"serializable": False})
     dimension: Optional[int] = field(kw_only=True, default=None, metadata={"serializable": True})
+    metric: Optional[str] = field(kw_only=True, default=None, metadata={"serializable": True})
     astra_db_namespace: Optional[str] = field(default=None, kw_only=True, metadata={"serializable": True})
 
     collection: Collection = field(init=False)
@@ -66,6 +70,7 @@ class AstraDBVectorStoreDriver(BaseVectorStoreDriver):
             .create_collection(
                 name=self.collection_name,
                 dimension=self.dimension,
+                metric=self.metric,
                 indexing=COLLECTION_INDEXING,
                 check_exists=False,
             )
--- a/griptape/drivers/vector/astra_db_vector_store_driver.py
+++ b/griptape/drivers/vector/astra_db_vector_store_driver.py
@@ -85,7 +85,7 @@ class AstraDBVectorStoreDriver(BaseVectorStoreDriver):
         vector_id: Optional[str] = None,
         namespace: Optional[str] = None,
         meta: Optional[dict] = None,
-        **kwargs,
+        **kwargs: Any,
     ) -> str:
         """Write a vector to the Astra DB store.
 
--- a/griptape/drivers/vector/astra_db_vector_store_driver.py
+++ b/griptape/drivers/vector/astra_db_vector_store_driver.py
@@ -1,0 +1,142 @@
+from __future__ import annotations
+
+import logging
+import warnings
+from typing import TYPE_CHECKING, Any, Optional
+
+from attrs import define, field
+
+from griptape.drivers import BaseVectorStoreDriver
+from griptape.utils import import_optional_dependency
+
+if TYPE_CHECKING:
+    from astrapy import Collection
+
+GRIPTAPE_VERSION: Optional[str]
+try:
+    from importlib import metadata
+
+    GRIPTAPE_VERSION = metadata.version("griptape")
+except Exception:
+    GRIPTAPE_VERSION = None
+
+logging.basicConfig(level=logging.WARNING)
+
+
+COLLECTION_INDEXING = {"deny": ["meta.artifact"]}
+
+
+@define
+class AstraDBVectorStoreDriver(BaseVectorStoreDriver):
+    """A Vector Store Driver for Astra DB.
+
+    DOCSTRING TO DO
+    """
+
+    api_endpoint: str = field(kw_only=True, metadata={"serializable": True})
+    token: str = field(kw_only=True, metadata={"serializable": False})
+    collection_name: str = field(kw_only=True, metadata={"serializable": False})
+    dimension: int = field(kw_only=True, metadata={"serializable": True})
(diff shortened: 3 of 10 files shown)


collindutter:
Thanks for the PR @hemidactylus! Will review this in detail next week.

hemidactylus:
Thank you so much, @collindutter ! Meanwhile I should have addressed the test coverage issues reported above. Cheers!

docs/examples/query-webpage-astra-db.md
```diff
@@ -0,0 +1,73 @@
+Required Griptape extras:
+
+```
+pip install griptape[drivers-vector-astra-db,drivers-web-scraper-trafilatura]
+```
```
collindutter, line 5: Let's replace this bit with a sentence of two explaining what the example does. Then link to the AstraDb entry on `vector-store-drivers.md` which explains the extras required.

docs/examples/query-webpage-astra-db.md
```diff
@@ -0,0 +1,73 @@
...
+from griptape.engines.rag.modules import (
+    PromptResponseRagModule,
+    VectorStoreRetrievalRagModule,
+)
+from griptape.engines.rag.stages import ResponseRagStage, RetrievalRagStage
+from griptape.loaders import WebLoader
+from griptape.structures import Agent
+from griptape.tools import RagClient, TaskMemoryClient
+
+if __name__ == "__main__":
```
collindutter, line 27: Can remove, other examples do not use `__main__` pattern.

docs/griptape-framework/drivers/vector-store-drivers.md
```diff
@@ -482,3 +482,49 @@ values = [r.to_artifact().value for r in results]
...
+
+```python
+import os
+from griptape.drivers import AstraDBVectorStoreDriver, OpenAiEmbeddingDriver
+from griptape.loaders import WebLoader
+
+# Astra DB secrets and connection parameters
+api_endpoint = os.environ["ASTRA_DB_API_ENDPOINT"]
+token = os.environ["ASTRA_DB_APPLICATION_TOKEN"]
+astra_db_namespace = os.environ.get("ASTRA_DB_KEYSPACE")  # optional
```
collindutter, line 503: Can you please add these environment variables to [here](https://github.com/griptape-ai/griptape/blob/e1977ce2f42665221482a1d365cc92931c5c6179/.github/workflows/docs-integration-tests.yml#L23)?

We'll set up values for them in the background.

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+    from astrapy import Collection
+    from astrapy.authentication import TokenProvider
+
+GRIPTAPE_VERSION: Optional[str]
+try:
+    from importlib import metadata
+
+    GRIPTAPE_VERSION = metadata.version("griptape")
+except Exception:
+    GRIPTAPE_VERSION = None
```
collindutter, line 22: This is interesting, though I feel like this should be done in a more generalized place rather than just this Vector Store Driver. Is this required for this PR?

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+
+logging.basicConfig(level=logging.WARNING)
```
collindutter, line 24: Let's remove this, a specific Vector Store Driver shouldn't change the logging config.

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+COLLECTION_INDEXING = {"deny": ["meta.artifact"]}
```
collindutter, line 27: Please move into class attribute.

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+        token: a Database Token ("AstraCS:...") secret to access Astra DB. An instance of `astrapy.authentication.TokenProvider` is also accepted.
+        collection_name: the name of the collection on Astra DB.
+        environment: the environment ("prod", "hcd", ...) hosting the target Data API.
+            It can be omitted for production Astra DB targets. See `astrapy.constants.Environment` for allowed values.
+        dimension: the number of components for embedding vectors. If not provided, it will be guessed from the embedding driver.
+        metric: the similarity metric to use, one of "dot_product", "euclidean" or "cosine".
+            If omitted, the server default ("cosine") will be used. See also values of `astrapy.constants.VectorMetric`.
+            If the vectors are normalized to unit norm, choosing "dot_product" over cosine yields up to 2x speedup in searches.
+        astra_db_namespace: optional specification of the namespace (in the Astra database) for the data.
+            *Note*: not to be confused with the "namespace" mentioned elsewhere, which is a grouping within this vector store.
```
collindutter, line 46: _Should_ this be the same as `namespace`?

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+                token=self.token,
+                namespace=self.astra_db_namespace,
+            )
+            .create_collection(
+                name=self.collection_name,
+                dimension=self.dimension,
+                metric=self.metric,
+                indexing=COLLECTION_INDEXING,
+                check_exists=False,
+            )
```
collindutter, line 81: Vector Store Drivers shouldn't create any resources, they are intended only for data-plane operations.

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+    metric: Optional[str] = field(kw_only=True, default=None, metadata={"serializable": True})
+    astra_db_namespace: Optional[str] = field(default=None, kw_only=True, metadata={"serializable": True})
+
+    collection: Collection = field(init=False)
+
+    def __attrs_post_init__(self) -> None:
+        astrapy = import_optional_dependency("astrapy")
+        if not self.dimension:
+            # auto-compute dimension from the embedding
+            self.dimension = len(self.embedding_driver.embed_string("This is a sample text."))
```
collindutter, line 63: This is a clever trick, though I wonder if we should just make dimensions required.

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+    environment: Optional[str] = field(kw_only=True, default=None, metadata={"serializable": True})
+    dimension: Optional[int] = field(kw_only=True, default=None, metadata={"serializable": True})
+    metric: Optional[str] = field(kw_only=True, default=None, metadata={"serializable": True})
+    astra_db_namespace: Optional[str] = field(default=None, kw_only=True, metadata={"serializable": True})
+
+    collection: Collection = field(init=False)
+
+    def __attrs_post_init__(self) -> None:
+        astrapy = import_optional_dependency("astrapy")
+        if not self.dimension:
```
collindutter, line 61: If we end up keeping this field optional, replace with explicit `is not None` check. See reasoning [here](https://google.github.io/styleguide/pyguide.html#2144-decision).

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+
+        Returns:
+            the ID of the written vector (str).
+        """
+        if kwargs:
+            warnings.warn(
+                "Unhandled keyword argument(s) provided to AstraDBVectorStore.upsert_vector: "
+                f"'{','.join(sorted(kwargs.keys()))}'. These will be ignored.",
+                stacklevel=2,
+            )
```
collindutter, line 123: Let's remove this. Currently none of the Vector Store Drivers really use `kwargs`. This should be improved globally.

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+        Args:
+            vector_id: the ID of the required vector.
+            namespace: a namespace, within the vector store, to constrain the search.
+
+        Returns:
+            The vector entry (a `BaseVectorStoreDriver.Entry`) if found, otherwise None.
+        """
+        find_filter = {k: v for k, v in {"_id": vector_id, "namespace": namespace}.items() if v is not None}
+        match = self.collection.find_one(filter=find_filter, projection={"*": 1})
+        if match:
```
collindutter, line 148: Use explicit `is not None` check

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+            A list of vector (`BaseVectorStoreDriver.Entry`) entries,
+            with their `score` attribute set to the vector similarity to the query.
+        """
+        query_filter: Optional[dict[str, Any]] = kwargs.pop("filter", None)
+        if kwargs:
+            warnings.warn(
+                "Unhandled keyword argument(s) provided to AstraDBVectorStore.query: "
+                f"'{','.join(sorted(kwargs.keys()))}'. These will be ignored.",
+                stacklevel=2,
+            )
```
collindutter, line 202: Same as above, let's remove.

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,224 @@
...
+            include_vectors: whether to include vector data in the results.
+            kwargs: additional keyword arguments. Currently only the free-form dict `filter`
+                is recognized (and goes straight to the Data API query);
+                others will generate a warning and be ignored.
+
+        Returns:
+            A list of vector (`BaseVectorStoreDriver.Entry`) entries,
+            with their `score` attribute set to the vector similarity to the query.
+        """
+        query_filter: Optional[dict[str, Any]] = kwargs.pop("filter", None)
```
collindutter, line 196: Let's just use `kwargs.get` since the `kwargs` warning will be removed.

tests/unit/drivers/vector/test_astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,160 @@
...
+    meta=one_document["meta"],
+    namespace=one_document["namespace"],
+)
+one_query_entry = BaseVectorStoreDriver.Entry(
+    id=one_document["_id"],
+    vector=one_document["$vector"],
+    meta=one_document["meta"],
+    namespace=one_document["namespace"],
+    score=one_document["$similarity"],
+)
```
collindutter, line 27: Move to pytest fixtures.

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,228 @@
...
+        return False
+
+
+def astra_db_available() -> bool:
+    return all(
+        [
+            "ASTRA_DB_APPLICATION_TOKEN" in os.environ,
+            "ASTRA_DB_API_ENDPOINT" in os.environ,
+        ]
+    )
```
collindutter, line 30: Can remove. Our integration tests assume the `all` extra is installed.

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def _descore_entry(entry: BaseVectorStoreDriver.Entry) -> BaseVectorStoreDriver.Entry:
+    return BaseVectorStoreDriver.Entry.from_dict({k: v for k, v in entry.__dict__.items() if k != "score"})
+
```
collindutter, line 35: Move into class

collindutter, line 35: Move into class

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,228 @@
...
+    def try_embed_chunk(self, chunk: str) -> list[float]:
+        vector2d: list[float]
+        try:
+            fraction = float(json.loads(chunk))
+            angle = fraction * math.pi * 2
+            vector2d = [math.cos(angle), math.sin(angle)]
+        except Exception:
+            vector2d = [0.0, 0.0]
+        return (vector2d + [0] * (self.dimensions))[: self.dimensions]
```
collindutter, line 52: Can you use `MockEmbeddingDriver` instead? Maybe adding a `mock_output` to better control the output?

collindutter (changes requested):
Great work! Thank you for putting the time into creating consistent documentation and passing tests -- it is greatly appreciated :)

docs/examples/query-webpage-astra-db.md
hemidactylus: Done (but I chose to leave a note about the `drivers-web-scraper-trafilatura` extra being also needed. Is that ok?)

griptape/drivers/vector/astra_db_vector_store_driver.py
hemidactylus: Agree, I had inadvertently left it there from some early script. Apologies.

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
hemidactylus: Should I remove the check on the env. vars as well (i.e. assume valid secrets are present)?
Or would I refine the `astra_db_available` to e.g. checking that the env.vars are not the empty string? (in the interest of not having test failing just because of no secrets passed).

griptape/drivers/vector/astra_db_vector_store_driver.py
hemidactylus: I'm all too happy to comply (I also prefer the explicit None check btw)

griptape/drivers/vector/astra_db_vector_store_driver.py
hemidactylus: I can remove it, no problem.

griptape/drivers/vector/astra_db_vector_store_driver.py
hemidactylus: Made `dimension` required in the end.

griptape/drivers/vector/astra_db_vector_store_driver.py
hemidactylus: I would say no: in Astra DB, a "namespace" is a group of tables/collections. (it is exactly what in traditional Cassandra was termed "keyspace"). In RDBMS language, a "schema".
(users who create a collection in Astra get to choose in which namespace it shall be created, and so on. Think of it as part of a "connection string" to the database).

I went for that parameter name to emphasize that it is a rather different concept from the Griptape notion of "namespace", which is more a per-entry tagging achieving a(n optional) partitioning _within_ the store IIRC.

Would you suggest a different naming, or a more detailed description in the docstring?

griptape/drivers/vector/astra_db_vector_store_driver.py
hemidactylus: Umm, I think this point warrants careful examination. Thanks for pointing it out!

If I get it right, the pattern you suggest would unfold invariably as follows:
1. user creates the collection before running the Griptape application (e.g. through the Astra DB Web UI);
2. they run the Griptape application, which finds the collection provisioned already, and starts using it.

If this is something vector store drivers should strictly adhere to, it can be done of course. (my personal opinion is that it would introduce a little more friction in adoption, but I see the cleanliness of this separation of concerns all right).

I guess I am trying to figure out how "negotiable" this requirement is. Admittedly, if the driver has the possibility to create the collection itself, a couple of further problems would be avoided: dimension mismatch (on Astra, users need to specify the vector dimension), optimal indexing setting (tailored to the usage Griptape does of the various document field; such optimization is not available if creating collections through the UI).

Let me know your opinion on this point, and I will adapt to whatever suggestion you have. (in case we end up removing the "create collection" step, I will have to add something to the .md files to instruct the users better around the UI).

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
hemidactylus: Sure, I have subsumed the above into MockEmbeddingDriver.
Note that the tests here have to be able to dictate the resulting (2d, unit) vector computed from a string, so I have re-packed the string-to-vector logic into a function that is passed as a new field to `MockEmbeddingDriver`, called `mock_output_function`.

Is that what you suggested? (no functional changes to "ordinary" use of the MockEmbeddingDriver).

tests/mocks/mock_embedding_driver.py
```diff
@@ -12,6 +14,11 @@ class MockEmbeddingDriver(BaseEmbeddingDriver):
     dimensions: int = field(default=42, kw_only=True)
     max_attempts: int = field(default=1, kw_only=True)
     tokenizer: MockTokenizer = field(factory=lambda: MockTokenizer(model="foo bar"), kw_only=True)
+    mock_output_function: Callable[[str], Optional[list[float]]] = field(default=lambda chunk: None, kw_only=None)
```
collindutter, line 17: Let's rename to `mock_output` [for consistency](https://github.com/griptape-ai/griptape/blob/6b7e8e9989f022a0743ad8b3b02d8afb348f014e/tests/mocks/mock_prompt_driver.py#L33) with `MockPromptDriver` and have the default be `Factory(lambda chunk: [0, 1])`.

Then `try_embed_chunk` can just become:
```python
    def try_embed_chunk(self, chunk: str) -> list[float]:
        return self.mock_output_function(chunk)
```

griptape/drivers/vector/astra_db_vector_store_driver.py
collindutter: Got it, thank you for explaining. I think the current naming is perfect then.

docs/examples/query-webpage-astra-db.md
collindutter: Perfect!

griptape/drivers/vector/astra_db_vector_store_driver.py
collindutter: What you described is correct. Today, we require users to create the collection/index/etc ahead of using the Vector Store Driver. While we could introduce a `create_collection` method, I fear that this would lead to the Drivers feeling a bit inconsistent -- as a user I would assume similar methods exist in other places.

I do think this could be valuable functionality, but I think we should implement for all Drivers at once. I created [an issue](https://github.com/griptape-ai/griptape/issues/1030) to track this.

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
collindutter: Yeah let's remove the env check as well. If running the full test suite we can assume that all env vars are set. Otherwise users can run individual test files.

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
collindutter: Let's rename to `mock_output` [for consistency](https://github.com/griptape-ai/griptape/blob/6b7e8e9989f022a0743ad8b3b02d8afb348f014e/tests/mocks/mock_prompt_driver.py#L33) with `MockPromptDriver` and have the default be `Factory(lambda chunk: [0, 1])`.

Then `try_embed_chunk` can just become:
```python
    def try_embed_chunk(self, chunk: str) -> list[float]:
        return self.mock_output_function(chunk)
```

hemidactylus:
(replying about the renaming of the `mock_output` parameter)

Sounds good to me -- renamed.

Only, I had to take out the `Factory` layer: the `mock_output` has to be a function, so I went for

```
default=lambda chunk: [0, 1]
```

(alternatively, I guess `default=Factory(lambda: lambda chunk: [0, 1])` would work as well - but I suppose it's not necessary since the "innermost" lambda does not look like a 'dangerous' mutable thing. I may be mistaken though)

griptape/drivers/vector/astra_db_vector_store_driver.py
hemidactylus: Right, this sounds a very valid and tidy point of view. I will adapt the code, test and examples accordingly, thank you for the prompt reply!

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
hemidactylus: All right, thanks! (done).

griptape/drivers/vector/astra_db_vector_store_driver.py
hemidactylus: Done. The more I look at this, the more it makes sense to me.

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
hemidactylus: Done (slightly different `default=` than you suggest, see comment below)

collindutter: `alternatively, I guess default=Factory(lambda: lambda chunk: [0, 1]) would work as well `

Let's actually go with this. While I think you're correct that the innermost lambda is not a dangerous mutable thing, we use this pattern in [other](https://github.com/griptape-ai/griptape/blob/d2e14b9b247010e04a72f13e8dd0b6a941015d74/griptape/tasks/base_task.py#L36-L39) [places](https://github.com/griptape-ai/griptape/blob/d2e14b9b247010e04a72f13e8dd0b6a941015d74/griptape/engines/rag/modules/retrieval/vector_store_retrieval_rag_module.py#L22-L24) throughout the framework.

griptape/drivers/vector/astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,184 @@
...
+from griptape.drivers import BaseVectorStoreDriver
+from griptape.utils import import_optional_dependency
+
+if TYPE_CHECKING:
+    from astrapy import Collection
+    from astrapy.authentication import TokenProvider
+
+
+@define
+class AstraDBVectorStoreDriver(BaseVectorStoreDriver):
```
collindutter, line 16: Ah let's rename to `AstraDbVectorStoreDriver` (`astradb_vector_store_driver.py`) for consistency with other Drivers like `AmazonDynamoDbConversationMemoryDriver` and `MongoDbAtlasVectorStoreDriver`.

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
```diff
@@ -0,0 +1,149 @@
...
+from griptape.drivers import AstraDBVectorStoreDriver, BaseVectorStoreDriver
+from tests.mocks.mock_embedding_driver import MockEmbeddingDriver
+
+TEST_COLLECTION_NAME = "gt_int_test"
+TEST_COLLECTION_NAME_METRIC = "gt_int_test_dot"
+
+
+class TestAstraDBVectorStoreDriver:
+    def _descore_entry(self, entry: BaseVectorStoreDriver.Entry) -> BaseVectorStoreDriver.Entry:
+        return BaseVectorStoreDriver.Entry.from_dict({k: v for k, v in entry.__dict__.items() if k != "score"})
```
collindutter, line 18: Please move protected methods after public methods (sorry we don't have an [linter rule](https://github.com/astral-sh/ruff/issues/2425) for this yet)

griptape/drivers/vector/astra_db_vector_store_driver.py
hemidactylus: Agree; I think our all-uppercase "DB" as per branding guidelines is indeed a weaker principle than internal naming consistency within Griptape.

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
hemidactylus: Sure, no problem.

griptape/drivers/vector/astra_db_vector_store_driver.py
collindutter: @hemidactylus sorry, last request: please rename `astra_db_vector_store_driver.py` to `astradb_vector_store_driver.py`.

hemidactylus: No problem :)

collindutter:
Great work again @hemidactylus! And thanks for working with me through the requested changes.

Just need to do some testing on a live AstraDb instance and then we'll be good to merge 🥳

hemidactylus:
That's great news @collindutter ! Do you want me to assist you with the (actually very simple) setup for integration testing? Please do let me know!

collindutter:
I appreciate the offer! I should be good, but will reach out if I have any questions.

collindutter:
@hemidactylus I tested it out, everything worked great! I'm recreating [the PR](https://github.com/griptape-ai/griptape/pull/1022) in our repo so that integration tests can run with our secrets, but all the commits from your fork are still present.

Thanks again for the excellent integration.

PR #1037, opened 2024-08-01
Add PromptEngine that adds events to PromptDrivers

dylanholmes:
WORK IN PROGRESS

Changes:
- Add a `PromptEngine` class:
  - Has a prompt driver. 
  - Publishes prompt events.
  - Retries when prompt driver fails.
  - Branches `run` logic based on value of `stream` for use in a structure.
- Edit `BasePromptDriver` responsibilities:
  - Remove event publishing.
  - Remove retry logic.
  - Remove `stream` branching logic.
- Edit consumers within the framework to use `PromptEngine` rather than `BasePromptDriver`
  - Required for behavior parity (event publishing + retry logic)
  - TODO: list the consumers, (there weren't that many...)
- Edit `StructureConfig`:
  - Remove references to event listeners and structure.
  - After making the other (non-prompt) drivers "stateless", you'll be able to use the same structure config instance across multiple structures without conflicts. (Though you should still avoid mutating the structure config after doing so, this PR doesn't solve that issue. Making structure config immutable would though)
 
 To summarize the main idea:
 1. Keep drivers "stateless".
 1. Allow engines to be "stateful".
 1. Client framework code should use engines rather than drivers, so that event publishing continues to work by default.

 In our case, this means moving event publishing logic from drivers into engines.
 
> Note: When I say "stateless", I don't actually mean no state; I actually mean "stateless" w.r.t. a specific piece of client code. I think its fine if say a driver has for example an s3 client, just as long as that wouldn't prevent a single instance of the driver from being used simultaneously in multiple contexts, like multiple structures.

> Note: There are some alternative options to this, that I'd like to quickly explore before making any decisions. Publishing this for discussion purposes only.

### DIFF
--- a/griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
+++ b/griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
@@ -56,8 +56,8 @@ class AmazonBedrockPromptDriver(BasePromptDriver):
     use_native_tools: bool = field(default=True, kw_only=True, metadata={"serializable": True})
     tool_choice: dict = field(default=Factory(lambda: {"auto": {}}), kw_only=True, metadata={"serializable": True})
 
-    @observable
-    def try_run(self, prompt_stack: PromptStack) -> Message:
+    @observable(tags=["PromptDriver.run()"])
+    def run(self, prompt_stack: PromptStack) -> Message:
         response = self.bedrock_client.converse(**self._base_params(prompt_stack))
 
         usage = response["usage"]
@@ -69,8 +69,8 @@ class AmazonBedrockPromptDriver(BasePromptDriver):
             usage=Message.Usage(input_tokens=usage["inputTokens"], output_tokens=usage["outputTokens"]),
         )
 
-    @observable
-    def try_stream(self, prompt_stack: PromptStack) -> Iterator[DeltaMessage]:
+    @observable(tags=["PromptDriver.stream()"])
+    def stream(self, prompt_stack: PromptStack) -> Iterator[DeltaMessage]:
         response = self.bedrock_client.converse_stream(**self._base_params(prompt_stack))
 
         stream = response.get("stream")
--- a/griptape/engines/extraction/base_extraction_engine.py
+++ b/griptape/engines/extraction/base_extraction_engine.py
@@ -9,7 +9,7 @@ from griptape.chunkers import BaseChunker, TextChunker
 
 if TYPE_CHECKING:
     from griptape.artifacts import ErrorArtifact, ListArtifact
-    from griptape.drivers import BasePromptDriver
+    from griptape.engines import PromptEngine
     from griptape.rules import Ruleset
 
 
@@ -17,10 +17,10 @@ if TYPE_CHECKING:
 class BaseExtractionEngine(ABC):
     max_token_multiplier: float = field(default=0.5, kw_only=True)
     chunk_joiner: str = field(default="\n\n", kw_only=True)
-    prompt_driver: BasePromptDriver = field(kw_only=True)
+    prompt_engine: PromptEngine = field(kw_only=True)
     chunker: BaseChunker = field(
         default=Factory(
-            lambda self: TextChunker(tokenizer=self.prompt_driver.tokenizer, max_tokens=self.max_chunker_tokens),
+            lambda self: TextChunker(tokenizer=self.prompt_engine.tokenizer, max_tokens=self.max_chunker_tokens),
             takes_self=True,
         ),
         kw_only=True,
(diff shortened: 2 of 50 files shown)


griptape/engines/extraction/base_extraction_engine.py
```diff
@@ -35,13 +37,13 @@ def validate_max_token_multiplier(self, _: Attribute, max_token_multiplier: int)
 
     @property
     def max_chunker_tokens(self) -> int:
-        return round(self.prompt_driver.tokenizer.max_input_tokens * self.max_token_multiplier)
+        return round(self.prompt_engine.prompt_driver.tokenizer.max_input_tokens * self.max_token_multiplier)
```
dylanholmes: Add a convenient `tokenizer` property method to `PromptEngine` for retrieving the tokenizer, since this access pattern is somewhat common.

In general, it'd be best to encapsulate the implementation details of `PromptEngine` as much as possible. Ideally, we should never be calling methods on `prompt_engine.prompt_driver`.

griptape/engines/extraction/csv_extraction_engine.py
```diff
@@ -85,7 +87,9 @@ def _extract_rec(
 
             rows.extend(
                 self.text_to_csv_rows(
-                    self.prompt_driver.run(PromptStack(messages=[Message(partial_text, role=Message.USER_ROLE)])).value,
+                    self.prompt_engine.prompt_driver.run(
```
dylanholmes: This should call `prompt_engine.run()` instead.

griptape/engines/prompt/prompt_engine.py
```diff
@@ -0,0 +1,106 @@
...
+from griptape.mixins import EventPublisherMixin, ExponentialBackoffMixin, SerializableMixin
+
+if TYPE_CHECKING:
+    from griptape.drivers.prompt.base_prompt_driver import BasePromptDriver
+
+
+@define(kw_only=True)
+class PromptEngine(SerializableMixin, ExponentialBackoffMixin, EventPublisherMixin, ABC):
+    stream: bool = field(default=False, kw_only=True, metadata={"serializable": True})
+    prompt_driver: BasePromptDriver = field(metadata={"serializable": True})
```
dylanholmes: Swap order of these fields

griptape/engines/rag/modules/response/prompt_response_rag_module.py
```diff
@@ -9,30 +9,32 @@
 from griptape.utils import J2
 
 if TYPE_CHECKING:
-    from griptape.drivers import BasePromptDriver
+    from griptape.engines.prompt.prompt_engine import PromptEngine
```
dylanholmes, line 12: Export `PromptEngine` from `griptape.engines`

griptape/structures/structure.py
```diff
@@ -164,9 +167,6 @@ def default_config(self) -> BaseStructureConfig:
 
             embedding_driver = OpenAiEmbeddingDriver() if self.embedding_driver is None else self.embedding_driver
 
-            if self.stream is not None:
```
dylanholmes, line 167: Remove `self.stream is not None` check from `if self.prompt_driver is not None or self.embedding_driver is not None or self.stream is not None:`

tests/mocks/mock_failing_prompt_driver.py
```diff
@@ -4,41 +4,22 @@
...
 from griptape.tokenizers import BaseTokenizer, OpenAiTokenizer
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
+    from griptape.common import DeltaMessage, Message, PromptStack
+
 
 @define
 class MockFailingPromptDriver(BasePromptDriver):
```
dylanholmes, line 17: Will need to re-create this for `PromptEngine` instead, since the retry logic is there now.

griptape/engines/extraction/base_extraction_engine.py
dylanholmes: done

griptape/engines/extraction/csv_extraction_engine.py
dylanholmes: done

griptape/engines/prompt/prompt_engine.py
dylanholmes: done

griptape/engines/rag/modules/response/prompt_response_rag_module.py
dylanholmes: done

griptape/structures/structure.py
dylanholmes: done

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
```diff
@@ -70,7 +70,7 @@ def try_run(self, prompt_stack: PromptStack) -> Message:
         )
 
     @observable
-    def try_stream(self, prompt_stack: PromptStack) -> Iterator[DeltaMessage]:
+    def stream(self, prompt_stack: PromptStack) -> Iterator[DeltaMessage]:
```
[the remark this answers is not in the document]
dylanholmes: done

dylanholmes:
Closing in favor of branch with better name: https://github.com/griptape-ai/griptape/pull/1038/files#diff-b11216b7a80ee74cad6522dac6b806dfd25bd3efda2a0251d80fab61446d6ba2L104-R104

PR #1038, opened 2024-08-01
Add PromptEngine

dylanholmes:
WORK IN PROGRESS

NOTE: I have not added docs for `PromptEngine`. Will do that if the discussion gets more serious

Changes:
- Add a `PromptEngine` class:
  - Has a prompt driver. 
  - Publishes prompt events.
  - Retries when prompt driver fails.
  - Branches `run` logic based on value of `stream` for use in a structure.
- Edit `BasePromptDriver` responsibilities:
  - Remove event publishing.
  - Remove retry logic.
  - Remove `stream` branching logic.
- Edit consumers within the framework to use `PromptEngine` rather than `BasePromptDriver`
  - Required for behavior parity (event publishing + retry logic)
  - TODO: list the consumers, (there weren't that many...)
  
Next Steps:
- Add an engine for each driver, which will have the event publishing, retry logic, and any other fancy thing.
- Update driver consumers to use engines instead.
- Edit `StructureConfig`:
  - Remove references to event listeners and structure.
  - After making the other (non-prompt) drivers "stateless", you'll be able to use the same structure config instance across multiple structures without conflicts. (Though you should still avoid mutating the structure config after doing so, this PR doesn't solve that issue. Making structure config immutable would though)
 
 To summarize the main idea:
 1. Keep drivers "stateless".
 1. Allow engines to be "stateful".
 1. Client framework code should use engines rather than drivers, so that event publishing continues to work by default.

 In our case, this means moving event publishing logic from drivers into engines.
 
> Note: When I say "stateless", I don't actually mean no state; I actually mean "stateless" w.r.t. a specific piece of client code. I think its fine if say a driver has for example an s3 client, just as long as that wouldn't prevent a single instance of the driver from being used simultaneously in multiple contexts, like multiple structures.

> Note: There are some alternative options to this, that I'd like to quickly explore before making any decisions. Publishing this for discussion purposes only.


dylanholmes:
This has been superseded by @collindutter 's work on adding an EventBus

PR #1045, opened 2024-08-06
Global Config

collindutter:

## Describe your changes
This change completely refactors how configuration is handled in the framework. Since the beginning we have relied on `Structure` to serve as the source of truth for configuration values. However, as the framework grew in complexity, it became increasingly difficult (and sometimes wrong) to propagate a `Structure` reference around in order to access these configuration values.

This PR introduces a global singleton, `griptape.config.config`, that stores these configuration values. Today, it stores configuration for drivers (previously called `StructureConfig`), and `logging`. The framework has been updated to pull from this object for default values.

Since this configuration now lives in the `config` object, I have removed these configuration values from `Structure`. Note that `Agent.prompt_driver` and `Agent.stream` were kept since the purpose of an Agent is to provide a syntactically concise way to create a Structure with either a `PromptTask`/`ToolkitTask`. These fields felt consistent with the other helper field: `Agent.tools`.

---

Overall I really like this change, but you might notice that not all drivers are included in `DriversConfig`. Does it makes sense to provide a default value for something like a `WebScraperDriver`? Or should this be left to the specific areas of the framework that require it? We could just provide a `None` default, but this may mask the fact that something like the `WebSearch` Tool does actually require one.

---

### Added:
-  Global config, `griptape.config.config`, for setting global configuration defaults.
### Changed:
- **BREAKING**: Removed `Pipeline.prompt_driver` and `Workflow.prompt_driver`. `Agent.prompt_driver` has not been removed.
- **BREAKING**: Removed `Pipeline.stream` and `Workflow.stream`. `Agent.stream` has not been removed.
- **BREAKING**: Removed `Structure.embedding_driver`, set this via `griptape.config.config.drivers.embedding` instead.
- **BREAKING**: Removed `Structure.custom_logger` and `Structure.logger_level`, set these via `griptape.config.config.logger` instead. 
- Engines that previously required Drivers now pull from `griptape.config.config.drivers` by default.
- **BREAKING**: Removed `BaseStructureConfig.merge_config`.
- **BREAKING**: Renamed `StructureConfig` to `DriverConfig`, and renamed fields accordingly.

Closes https://github.com/griptape-ai/griptape/issues/826 (`merge_config` is removed).

----
📚 Documentation preview 📚: https://griptape--1045.org.readthedocs.build//1045/


collindutter:
Still looking over docs, but ready for eyes

vachillo:
can you write up the motivation for these changes and an example? hard to look at 123 files that seems like mostly small config changes and formatting

griptape/utils/chat.py
```diff
@@ -25,12 +25,16 @@ class Chat:
     )
 
     def default_output_fn(self, text: str) -> None:
-        if self.structure.config.prompt_driver.stream:
+        from griptape.config import config
+
+        if config.drivers.prompt.stream:
```
vachillo, line 30: this could be potentially overriden right? the global config is the source of truth for streaming?

griptape/utils/chat.py
```diff
@@ -40,7 +44,7 @@ def start(self) -> None:
                 self.output_fn(self.exiting_text)
                 break
 
-            if self.structure.config.prompt_driver.stream:
+            if config.drivers.prompt.stream:
```
vachillo, line 47: same here

griptape/utils/stream.py
```diff
@@ -34,7 +34,9 @@ class Stream:
 
     @structure.validator  # pyright: ignore[reportAttributeAccessIssue]
     def validate_structure(self, _: Attribute, structure: Structure) -> None:
-        if not structure.config.prompt_driver.stream:
+        from griptape.config import config
+
+        if not config.drivers.prompt.stream:
```
vachillo, line 39: same here

griptape/utils/chat.py
collindutter: Great point, updated the logic to look at the actual Tasks.

griptape/utils/chat.py
collindutter: Updated logic.

griptape/utils/stream.py
collindutter: Updated logic

griptape/utils/stream.py
```diff
@@ -34,10 +34,13 @@ class Stream:
...
-        from griptape.config import config
+        from griptape.tasks import PromptTask
 
-        if not config.drivers.prompt.stream:
-            raise ValueError("prompt driver does not have streaming enabled, enable with stream=True")
+        streaming_tasks = [
+            task for task in structure.tasks if isinstance(task, PromptTask) and task.prompt_driver.stream
+        ]
+        if not streaming_tasks:
+            raise ValueError("Structure does not have any streaming tasks, enable with stream=True")
```
vachillo, line 43: looking at this again, what would happen if you had one streaming task and the others were not? would the utility still work?

collindutter: Yes it would still work, it would just "hang" while working on non-streaming tasks.

dylanholmes (dismissed):
Couple of comments, but they can be addressed in another PR if at all. No more revisions on this plz

vachillo (approved):

CHANGELOG.md
```diff
@@ -13,10 +13,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Method `try_find_task` to `Structure`.
 - `TranslateQueryRagModule` `RagEngine` module for translating input queries.
 - Global event bus, `griptape.events.event_bus`, for publishing and subscribing to events.
+- Global config, `griptape.config.config`, for setting global configuration defaults.
 
 ### Changed
 - **BREAKING**: Removed all uses of `EventPublisherMixin` in favor of `event_bus`.
 - **BREAKING**: Removed `EventPublisherMixin`.
+- **BREAKING**: Removed `Pipeline.prompt_driver` and `Workflow.prompt_driver`. `Agent.prompt_driver` has not been removed.
```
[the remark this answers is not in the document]
collindutter: From the PR description:
> Note that Agent.prompt_driver and Agent.stream were kept since the purpose of an Agent is to provide a syntactically concise way to create a Structure with either a PromptTask/ToolkitTask. These fields felt consistent with the other helper field: Agent.tools.

Let me know if this is not a sufficient explanation, happy to go into more detail.

CHANGELOG.md
```diff
@@ -13,10 +13,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **BREAKING**: Removed `Pipeline.stream` and `Workflow.stream`. `Agent.stream` has not been removed.
```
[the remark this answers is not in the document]
collindutter: Replied above.

docs/griptape-framework/drivers/embedding-drivers.md
```diff
@@ -128,7 +128,7 @@ The [CohereEmbeddingDriver](../../reference/griptape/drivers/embedding/cohere_em
 ```
 
 ### Override Default Structure Embedding Driver
-Here is how you can override the Embedding Driver that is used by default in Structures. 
+Here is how you can override the Embedding Driver that is used by default in Structures.
 
 ```python
 --8<-- "docs/griptape-framework/drivers/src/embedding_drivers_10.py"
```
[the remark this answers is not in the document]
collindutter: Agreed, but the type checking/linting was essential during development.

docs/griptape-framework/misc/src/events_3.py
```diff
@@ -16,9 +16,9 @@
     ]
 )
 
-pipeline = Pipeline(
-    config=OpenAiStructureConfig(prompt_driver=OpenAiChatPromptDriver(model="gpt-4o", stream=True)),
-)
+config.drivers = OpenAiDriverConfig(prompt=OpenAiChatPromptDriver(model="gpt-4o", stream=True))
+
+pipeline = Pipeline()
```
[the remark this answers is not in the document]
collindutter: Agreed, I just moved the location to a more natural location.

griptape/structures/agent.py
```diff
@@ -32,11 +36,20 @@ def validate_fail_fast(self, _: Attribute, fail_fast: bool) -> None:  # noqa: FB
...
     def __attrs_post_init__(self) -> None:
         super().__attrs_post_init__()
+
+        self.prompt_driver.stream = self.stream
         if len(self.tasks) == 0:
             if self.tools:
-                task = ToolkitTask(self.input, tools=self.tools, max_meta_memory_entries=self.max_meta_memory_entries)
+                task = ToolkitTask(
+                    self.input,
+                    prompt_driver=self.prompt_driver,
```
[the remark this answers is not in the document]
collindutter: It does. `self.prompt_driver` defaults to `config.drivers.prompt`

griptape/structures/agent.py
```diff
@@ -32,11 +36,20 @@ def validate_fail_fast(self, _: Attribute, fail_fast: bool) -> None:  # noqa: FB
...
+                    tools=self.tools,
+                    max_meta_memory_entries=self.max_meta_memory_entries,
+                )
             else:
-                task = PromptTask(self.input, max_meta_memory_entries=self.max_meta_memory_entries)
+                task = PromptTask(
+                    self.input, prompt_driver=self.prompt_driver, max_meta_memory_entries=self.max_meta_memory_entries
```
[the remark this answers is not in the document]
collindutter: Good catch, updated `stream` to default to `config.drivers.prompt.stream`.

docs/griptape-framework/misc/src/events_3.py
```diff
@@ -16,13 +15,11 @@
...
 
-pipeline = Pipeline(
-    config=OpenAiStructureConfig(prompt_driver=OpenAiChatPromptDriver(model="gpt-4o", stream=True)),
-)
-
+pipeline = Pipeline()
 pipeline.add_tasks(
     ToolkitTask(
         "Based on https://griptape.ai, tell me what griptape is.",
+        prompt_driver=OpenAiChatPromptDriver(model="gpt-4o", stream=True),
```
dylanholmes, line 22: I suppose passing `stream` in to the driver works, but I think its a little confusing still. I'd expect either  `pipeline.run(stream=True)` or maybe `Pipeline(stream=True)`...

griptape/structures/agent.py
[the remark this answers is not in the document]
dylanholmes: I guess I mean, why do we need `prompt_driver` on agent at all?

dylanholmes (approved):
Looks fine, but I replied to some comments

docs/griptape-framework/misc/src/events_3.py
collindutter: A Pipeline may not contain any Tasks that use Prompt Drivers therefore streaming would be irrelevant. For Agents, it's always either be a Prompt Task or Toolkit Task (unless a user overrides `tasks`).

griptape/structures/agent.py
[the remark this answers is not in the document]
collindutter: We technically don't, it's just a convenience field to avoid dealing with the configuration object.

dylanholmes (approved):

PR #1046, opened 2024-08-06
Added `TranslateQueryRagModule` and updated RAG docs

vasinov:

## Describe your changes

* Added `TranslateQueryRagModule`.
* Updated RAG docs.
* Minor `RagEngine` refactors.

----
📚 Documentation preview 📚: https://griptape--1046.org.readthedocs.build//1046/

### DIFF
--- a/docs/griptape-framework/engines/rag-engines.md
+++ b/docs/griptape-framework/engines/rag-engines.md
@@ -21,7 +21,7 @@ search:
 
 #### Query
 
-- `TranslateQueryRagModule` is for translating the query into another langugage.
+- `TranslateQueryRagModule` is for translating the query into another language.
 
 #### Retrieval
 - `TextRetrievalRagModule` is for retrieving text chunks.
--- a/docs/griptape-framework/engines/rag-engines.md
+++ b/docs/griptape-framework/engines/rag-engines.md
@@ -21,7 +21,7 @@ search:
 
 #### Query
 
-No modules implemented yet.
+- `TranslateQueryRagModule` is for translating the query into another langugage.
 
 #### Retrieval
 - `TextRetrievalRagModule` is for retrieving text chunks.
--- a/docs/griptape-framework/engines/rag-engines.md
+++ b/docs/griptape-framework/engines/rag-engines.md
@@ -12,7 +12,7 @@ search:
 
 ### RAG Stages
 
-`RagEngine`s consist of three stages: `QueryRagStage`, `RetrievalRagStage`, and `ResponseRagStage`. Stages are always executed sequentially. `RagEngine`s are not meant to replace [Workflows](../../reference/griptape/structures/workflows.md); that's why they don't implement arbitrary DAGs.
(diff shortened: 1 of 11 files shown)


docs/griptape-framework/engines/rag-engines.md
```diff
@@ -8,70 +8,101 @@ search:
...
-
-`RagContext` is a container object for passing around RAG context. 
+[Rag Engine](../../reference/griptape/engines/rag/index.md) is an abstraction for implementing modular retrieval augmented generation (RAG) pipelines.
 
 ### RAG Stages
-- `QueryRagStage` is for parsing and expanding queries.
-- `RetrievalRagStage` is for retrieving content.
-- `ResponseRagStage` is for augmenting and generating outputs.
+
+`RagEngine`s consist of three stages: `QueryRagStage`, `RetrievalRagStage`, and `ResponseRagStage`. Stages are always executed sequentially. `RagEngine`s are not meant to replace [Workflows](../structures/workflows.md); that's why they don't implement arbitrary DAGs.
```
collindutter, line 15: > Stages are always executed sequentially. `RagEngine`s are not meant to replace [Workflows]

Seems like the confusion would targeted more be towards Pipelines?

griptape/engines/rag/modules/query/translate_query_rag_module.py
```diff
@@ -0,0 +1,32 @@
...
+    from griptape.engines.rag import RagContext
+
+
+@define(kw_only=True)
+class TranslateQueryRagModule(BaseQueryRagModule):
+    prompt_driver: BasePromptDriver = field()
+    language: str = field()
+    generate_user_template: Callable[[str, str], str] = field(
+        default=Factory(lambda self: self.default_user_template_generator, takes_self=True),
+    )
```
collindutter, line 21: Should this be a `J2` object for consistency with [other places](https://github.com/griptape-ai/griptape/blob/a20309ea63d08c140dce2bd449ea56af4f086ed7/griptape/engines/summary/prompt_summary_engine.py#L23-L24)?

collindutter (commented):
Nice work!

griptape/engines/rag/modules/query/translate_query_rag_module.py
vasinov: We should actually move `SummaryConversationMemory` and other static templates to be `Callables` so that the user has more control over how the template is generated. Similar to what we do in `PromptTask` and `ToolkitTask`.

docs/griptape-framework/engines/rag-engines.md
vasinov: I added this to address the workflows concern that already came up twice in other contexts. I'll mention pipelines as well.

griptape/engines/rag/modules/query/translate_query_rag_module.py
collindutter: Fair point, though the type signature of `Callable[[str, str], str]` might be confusing for users to interact with. Maybe we should do `Callable[dict], str`?

vasinov: `dict` doesn't make sense because I only pass query and language to the method.

collindutter (approved):

PR #1048, opened 2024-08-07
Modify arguments/return type of add_parent and add_child

collindutter:

### Changed
- **BREAKING**: `BaseTask.add_parent/child` now take a `BaseTask` instead of `str | BaseTask`.
- `BaseTask.add_parent/child` will now call `self.structure.add_task` if possible.
- `BaseTask.add_parent/child` returns the added parent/child Task allowing for chaining.

Allowing for chaining such as:
```python
from griptape.structures import Workflow
from griptape.tasks import ToolkitTask, CodeExecutionTask
from griptape.rules import Rule
from griptape.artifacts import GenericArtifact, ListArtifact

regions = ["sfbay", "nyc", "la", "boston", "seattle", "austin", "dc", "denver"]
workflow = Workflow()

for region in regions:
    ToolkitTask(
        tools=[
            WebScraper(),
        ],
        rules=[
            Rule(
                "return only the number of pages as an integer. no other text is allowed."
            )
        ],
        input=f"go to gregslist.com/{region}/search and use the site content to return the number of pages that are available.",
        structure=workflow,
        id=f"get_pages_{region}",
    ).add_child(
        CodeExecutionTask(
            run_fn=lambda task: GenericArtifact(int(task.parents_output_text)),
            structure=workflow,
            id=f"convert_pages_{region}",
        )
    ).add_child(
        CodeExecutionTask(
            run_fn=lambda task: ListArtifact(list(task.parents.values())),
            id="aggregate_pages",
        )
    )

num_pages = workflow.run().output
print(num_pages)
```

### DIFF
--- a/docs/griptape-framework/structures/workflows.md
+++ b/docs/griptape-framework/structures/workflows.md
@@ -201,3 +201,31 @@ output:
 [06/18/24 09:52:23] INFO     PromptTask new-animal
                              Output: elephant
 ```
+
+### Bitshift Composition
+
+Task relationships can also be set up with the Python bitshift operators `>>` and `<<`. The following four statements are all functionally equivalent:
+
+```python
+task1 >> task2
+task1.add_child(task2)
+
+task2 << task1
+task2.add_parent(task1)
+```
+
+When using the bitshift to compose operators, the relationship is set in the direction that the bitshift operator points.
+For example, `task1 >> task2` means that `task1` runs first and `task2` runs second.
+Multiple operators can be composed – keep in mind the chain is executed left-to-right and the rightmost object is always returned. For example:
+
+```python
+task1 >> task2 >> task3 << task4
+```
+
+is equivalent to:
+
+```python
+task1.add_child(task2)
+task2.add_child(task3)
+task3.add_parent(task4)
+```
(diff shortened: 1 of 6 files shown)


vachillo:
a few other methods return `self`, is this going to be confusing?

collindutter:
Hmm good point. I do think we should return _something_, but I can see the argument for retuning `self`.

collindutter:
@dylanholmes 
> I understand why, I just don't think the method name is right for that.

What about something like what [pydags](https://github.com/DavidTorpey/pydags?tab=readme-ov-file#example-usage) does with `after`? Though is that confusing to have `add_child` (returns self), and `after` (returns child)?

collindutter:
Ok what about using bitshift operators like [Airflow does](https://airflow.apache.org/docs/apache-airflow/1.10.3/concepts.html#bitshift-composition)? Note that they do _not_ return anything in their equivalent of `add_parent`/`add_child`.

Would it be confusing to return `self` from `add_parent`/`add_child`, but return `other` when using the bitshift?

docs/griptape-framework/structures/workflows.md
```diff
@@ -201,3 +201,31 @@ output:
 [06/18/24 09:52:23] INFO     PromptTask new-animal
                              Output: elephant
 ```
+
+### Bitshift Composition
```
dylanholmes, line 205: I like it!

dylanholmes (dismissed):
Nice work! Love the creativity.

dylanholmes (dismissed):

dylanholmes (approved):

PR #1050, opened 2024-08-07
updating structure run result to return BaseArtifact

emjay07:

## Describe your changes
- Update output Artifact type from structure runs to `BaseArtifact`

N/A

### DIFF
--- a/griptape/drivers/structure_run/griptape_cloud_structure_run_driver.py
+++ b/griptape/drivers/structure_run/griptape_cloud_structure_run_driver.py
@@ -6,7 +6,7 @@ from urllib.parse import urljoin
 
 from attrs import Factory, define, field
 
-from griptape.artifacts import BaseArtifact, ErrorArtifact, InfoArtifact, TextArtifact
+from griptape.artifacts import BaseArtifact, ErrorArtifact, InfoArtifact
 from griptape.drivers.structure_run.base_structure_run_driver import BaseStructureRunDriver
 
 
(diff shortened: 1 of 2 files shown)


collindutter (commented):
Thanks! Can you please update the changelog?

collindutter (approved):

griptape/drivers/structure_run/griptape_cloud_structure_run_driver.py
```diff
@@ -44,7 +44,7 @@ def try_run(self, *args: BaseArtifact) -> BaseArtifact:
         except (exceptions.RequestException, HTTPError) as err:
             return ErrorArtifact(str(err))
 
-    def _get_structure_run_result(self, structure_run_id: str) -> InfoArtifact | TextArtifact | ErrorArtifact:
+    def _get_structure_run_result(self, structure_run_id: str) -> InfoArtifact | BaseArtifact | ErrorArtifact:
```
vachillo, line 47: should `InfoArtifact` and `ErrorArtifact` be removed from the type hint?

emjay07: Those are explicitly separate code paths, so I thought it was useful to leave them. I do know they inherit from BaseArtifact though, so *technically* they don't need to be there.

collindutter: That's how I reasoned about it too, I think we should keep them.

PR #1051, opened 2024-08-07
Add global event bus

collindutter:

### Added
- Global event bus, `griptape.events.EventBus`, for publishing and subscribing to events.

### Changed
- **BREAKING**: Removed all uses of `EventPublisherMixin` in favor of `EventBus`.
- **BREAKING**: Removed `EventPublisherMixin`.

A couple things I'm unsure about:
1. Should this be a singleton or should we have a namespaced factory method similar to python's `logging` module? Something like:
```python
def get_event_bus(namespace: str = DEFAULT_NAMESPACE) ...
```
I can't think of any scenario where you'd want multiple busses, especially since you can pass in many `EventListeners`.
2. If we stick with a singleton, does the PascalCase naming of the variable make sense? The idea was that it might convey the global nature of it, similar to how you might access static fields.
NA

----
📚 Documentation preview 📚: https://griptape--1051.org.readthedocs.build//1051/


docs/griptape-framework/misc/events.md
```diff
@@ -131,24 +131,26 @@ Handler 2 <class 'griptape.events.finish_structure_run_event.FinishStructureRunE
...
+from griptape.events import CompletionChunkEvent, EventListener, EventBus
 from griptape.tasks import ToolkitTask
 from griptape.structures import Pipeline
 from griptape.tools import WebScraper, TaskMemoryClient
 from griptape.config import OpenAiStructureConfig
 from griptape.drivers import OpenAiChatPromptDriver
 
 
+
+EventBus.event_listeners = [
```
dylanholmes, line 143: I think I'd prefer `EventBus.add_event_listener()` similar to `EventBus.publish_event()`

griptape/config/base_structure_config.py
```diff
@@ -39,43 +36,6 @@ class BaseStructureConfig(BaseConfig, ABC):
     text_to_speech_driver: BaseTextToSpeechDriver = field(kw_only=True, metadata={"serializable": True})
     audio_transcription_driver: BaseAudioTranscriptionDriver = field(kw_only=True, metadata={"serializable": True})
 
-    _structure: Structure = field(default=None, kw_only=True, alias="structure")
-    _event_listener: Optional[EventListener] = field(default=None, kw_only=True, alias="event_listener")
```
dylanholmes, line 43: So nice

griptape/events/event_bus.py
```diff
@@ -32,3 +32,6 @@ def remove_event_listener(self, event_listener: EventListener) -> None:
     def publish_event(self, event: BaseEvent, *, flush: bool = False) -> None:
         for event_listener in self.event_listeners:
             event_listener.publish_event(event, flush=flush)
+
+
+EventBus = _EventBus()
```
dylanholmes, line 37: I could maybe guess at reasons for not doing this, but why not associate fields/methods with the class?

I only ask, because it feels like we're trying to mimic a class with state, but not actually doing it, and I don't exactly understand why

griptape/events/event_bus.py
```diff
@@ -9,7 +9,7 @@
 
 
 @define
-class EventPublisherMixin:
+class _EventBus:
```
dylanholmes, line 12: I like how little change was needed in this file

docs/griptape-framework/misc/events.md
dylanholmes: Oh, now I realize that there is a method for this. Can we prefer that one in examples?

dylanholmes (commented):
I really like the direction!

The changes look functionally good, but I'd like to chat before approving.

About the multiple buses scenario -- I can't really imagine a use case, would events get published to all of them or just some of them (if some of them, how would the publisher decide which)? If you just want to listen to a subset of events, the event listener should be able to filter out events it doesn't care about.

docs/griptape-framework/misc/events.md
collindutter: Yep updated 👍

collindutter:
@dylanholmes what do you think about dropping the `event` portion on all methods/fields in `EventBus`? `EventBus.add_listener`, for example.

griptape/events/event_bus.py
```diff
@@ -9,8 +9,12 @@
 
 
 @define
-class EventPublisherMixin:
-    event_listeners: list[EventListener] = field(factory=list, kw_only=True)
+class _EventBus:
+    _event_listeners: list[EventListener] = field(factory=list, kw_only=True, alias="_event_listeners")
```
collindutter, line 13: https://www.attrs.org/en/stable/init.html#private-attributes-and-aliases

griptape/events/event_bus.py
dylanholmes: chatted about this, resolving

dylanholmes:
> @dylanholmes what do you think about dropping the `event` portion on all methods/fields in `EventBus`? `EventBus.add_listener`, for example.

Not sure, I kind of like `add_event_listener`, but maybe I'm brainwashed by javascript in the browser. I don't really care either way honestly though. Another alternative could be `EventBus.subscribe(...)`.

This seems like a good opportunity to mention another "syntax golfing" idea I had that I wasn't going to share publicly. Accept the `EventListener` init params directly in the method to avoid the boilerplate (`...EventListener(...)...`)
```py
event_bus.add_event_listener(my_callback, event_types=[...])
```

griptape/events/event_bus.py
dylanholmes: This is uglier than just not using attrs imo. Plus it still adds `_event_listeners` as an keyword arg in `init`.

I don't care that much though

collindutter:
> Not sure, I kind of like add_event_listener
Let's just leave it as is.

> Accept the EventListener init params directly in the method to avoid the boilerplate
I like the conciseness, but I'd rather introduce even more breaking changes at this point. Let's evaluate in the future.

docs/griptape-framework/drivers/event-listener-drivers.md
```diff
@@ -14,26 +14,27 @@ import os
...
-agent = Agent(
-    rules=[
-        Rule(
-            value="You will be provided with a block of text, and your task is to extract a list of keywords from it."
-        )
-    ],
-    event_listeners=[
+event_bus.add_event_listeners(
+    [
         EventListener(
```
dylanholmes, line 24: Does this need to be plural? Can there be a singular `add_event_listener` too?

dylanholmes (approved):
Nice work!

Just one optional suggestion

docs/griptape-framework/drivers/event-listener-drivers.md
collindutter: It does not _need_to be plural, no. Let's just leave it as is, docs will get an overhaul at some point soon.

PR #1059, opened 2024-08-12
Add logging filters, update config namespaces

### LINKED ISSUE #902: Suppress logs for inputs and outputs on stdout

I dont want always see the inputs and outputs of all tasks that are running. If I am doing webscraping, for example, the entire webscraped output being printed multiple times on stdout is not useful.

toggle the input and output logging for tasks. we currently allow overriding of a log level or the entire logger, but its not entirely clear how I would leverage that for this case. 

probably not implemented this way but this is what i am currently doing:
```
class BaseTask(ABC):
    .....
    log_input: bool = field(default=True, kw_only=True)
    log_output: bool = field(default=True, kw_only=True)
    
```
and in the various `Task`s that  the inputs/outputs are logged:
```
    def before_run(self) -> None:
        super().before_run()

        self.structure.logger.info(
            f"{self.__class__.__name__} {self.id}\nInput: {self.input.to_text() if self.log_input else self.input.name}"
        )

    def after_run(self) -> None:
        super().after_run()

        self.structure.logger.info(
            f"{self.__class__.__name__} {self.id}\nOutput: {self.output.to_text() if self.log_output else self.output.name}"
        )
```

**Additional context**

vachillo:

## Describe your changes
update log levels for tasks so that outputs can be toggled off

closes #902


collindutter:
After thinking about this some more I'm not sure I'm a fan of moving Task output to `logging.debug` statements since most of the time this is useful information. To keep the same functionality as today we'd have to set griptape's default logging level to `logging.DEBUG` which is ["typically only of interest to a developer trying to diagnose a problem."](https://docs.python.org/3/library/logging.html#logging-levels)

The original ticket is regarding suppressing logs, so why don't we use [filter objects](https://docs.python.org/3/library/logging.html#filter-objects) built into the logging module? If added to the [config module](https://github.com/griptape-ai/griptape/pull/1045), it could look something like this:

```python
import logging

from griptape.config import config


class TruncateFilter(logging.Filter):
    MAX_LOG_LENGTH = 1000

    def filter(self, record):
        message = record.getMessage()

        if len(message) > self.MAX_LOG_LENGTH:
            record.msg = f"{message[:self.MAX_LOG_LENGTH]}... [TRUNCATED]"
            record.args = ()
        # Could also return False to prevent the log from being emitted

        return True


config.logging.add_filter(TruncateFilter())


agent = Agent(
    input="Tell me about {{ args[0] }}",
    tools=[
        WebScraper(),
    ],
)
agent.run("https://en.wikipedia.org/wiki/Vim_(text_editor)")
```

griptape/config/logging/filters/__init__.py
collindutter, line 1: These can probably just live in `config.logging`, what do you think?

griptape/config/logging/logging_config.py
```diff
@@ -14,11 +14,16 @@ class LoggingConfig:
...
 
     def __attrs_post_init__(self) -> None:
         logger = logging.getLogger(self.logger_name)
 
         logger.propagate = False
         logger.setLevel(self.logger_level)
 
-        logger.handlers = [RichHandler(show_time=True, show_path=False)]
+        logger.handlers = self.logger_handlers
+        [logger.addFilter(logger_filter) for logger_filter in self.logger_filters]
```
collindutter, line 29: Avoid using list comprehension when the return value is not used.

griptape/config/logging/logging_config.py
```diff
@@ -14,11 +14,16 @@ class LoggingConfig:
...
+    logger_filters: list[logging.Filter] = field(factory=list, kw_only=True)
 
     def __attrs_post_init__(self) -> None:
         logger = logging.getLogger(self.logger_name)
 
         logger.propagate = False
         logger.setLevel(self.logger_level)
 
-        logger.handlers = [RichHandler(show_time=True, show_path=False)]
+        logger.handlers = self.logger_handlers
```
collindutter, line 28: Is there an `addHandler` method we can use?

collindutter (changes requested):
Nice 👍 

Looks like docs are not building, and can you please adding a logging section to the existing config doc?

griptape/config/logging/filters/__init__.py
vachillo: should _everything_ get exported to the top level? wasnt sure what the standard is

collindutter:
@vachillo all tests appear to be failing

vachillo:
> @vachillo all tests appear to be failing

missed imports

docs/griptape-framework/structures/config.md
```diff
@@ -5,45 +5,47 @@ search:
 
 ## Overview
 
-The [DriverConfig](../../reference/griptape/config/driver_config.md) class allows for the customization of Structures within Griptape, enabling specific settings such as Drivers to be defined for Tasks. 
+TODO
```
collindutter, line 8: TODO

vachillo: bad at words

collindutter (approved):

PR #1061, opened 2024-08-14
Update dependencies, fix emergent issues

collindutter:

## Describe your changes
Update dependencies, fix emergent issues
NA


griptape/drivers/event_listener/griptape_cloud_event_listener_driver.py
```diff
@@ -25,12 +26,14 @@ class GriptapeCloudEventListenerDriver(BaseEventListenerDriver):
         default=Factory(lambda: os.getenv("GT_CLOUD_BASE_URL", "https://cloud.griptape.ai")),
         kw_only=True,
     )
-    api_key: str = field(default=Factory(lambda: os.getenv("GT_CLOUD_API_KEY")), kw_only=True)
+    api_key: Optional[str] = field(default=Factory(lambda: os.getenv("GT_CLOUD_API_KEY")), kw_only=True)
```
dylanholmes, line 29: Why should `api_key` be allowed to be `None`?

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
```diff
@@ -14,7 +14,7 @@
 
 
 class TestAstraDbVectorStoreDriver:
-    @pytest.fixture()
+    @pytest.fixture
```
dylanholmes, line 17: Why reverting this change? Did we change formatting rules?

(looks like this is the biggest diff contributor in the pr)

griptape/drivers/event_listener/griptape_cloud_event_listener_driver.py
collindutter: Because `os.getenv("GT_CLOUD_API_KEY")` returns `str | None`. Should we instead change to `os.environ`?

tests/integration/drivers/vector/test_astra_db_vector_store_driver.py
collindutter: Ah I just realized that the original is the ["correct"](https://docs.astral.sh/ruff/rules/pytest-fixture-incorrect-parentheses-style/) way. I had it backwards, will investigate and revert.

vachillo (dismissed):

vachillo (approved):

PR #1062, opened 2024-08-14
Lazily initialize config drivers field

collindutter:

## Describe your changes
Lazily initializes the `config.driver` field so that the `OpenAiChatPromptDriver` doesn't initialize itself when the user doesn't intend to use it.
NA

### DIFF
--- a/griptape/config/base_config.py
+++ b/griptape/config/base_config.py
@@ -1,6 +1,6 @@
 from abc import ABC
 
-from attrs import define
+from attrs import define, field
 
 from griptape.mixins.serializable_mixin import SerializableMixin
 
@@ -10,5 +10,5 @@ from .logging_config import LoggingConfig
 
 @define(kw_only=True)
 class BaseConfig(SerializableMixin, ABC):
-    drivers: BaseDriverConfig
-    logging: LoggingConfig
+    logging: LoggingConfig = field()
+    _drivers: BaseDriverConfig = field()
--- a/griptape/config/base_config.py
+++ b/griptape/config/base_config.py
@@ -1,14 +1,22 @@
+from __future__ import annotations
+
 from abc import ABC
+from typing import TYPE_CHECKING, Optional
 
 from attrs import define, field
 
 from griptape.mixins.serializable_mixin import SerializableMixin
 
-from .base_driver_config import BaseDriverConfig
-from .logging_config import LoggingConfig
+if TYPE_CHECKING:
+    from .base_driver_config import BaseDriverConfig
+    from .logging_config import LoggingConfig
 
 
 @define(kw_only=True)
 class BaseConfig(SerializableMixin, ABC):
-    logging: LoggingConfig = field()
-    _drivers: BaseDriverConfig = field()
+    _logging: Optional[LoggingConfig] = field()
+    _drivers: Optional[BaseDriverConfig] = field()
+
+    def reset(self) -> None:
+        self._logging = None
+        self._drivers = None
--- a/griptape/config/base_config.py
+++ b/griptape/config/base_config.py
@@ -14,8 +14,8 @@ if TYPE_CHECKING:
 
 @define(kw_only=True)
 class BaseConfig(SerializableMixin, ABC):
-    _logging: Optional[LoggingConfig] = field()
-    _drivers: Optional[BaseDriverConfig] = field()
+    _logging: Optional[LoggingConfig] = field(alias="logging")
+    _drivers: Optional[BaseDriverConfig] = field(alias="drivers")
 
     def reset(self) -> None:
         self._logging = None
--- a/griptape/config/config.py
+++ b/griptape/config/config.py
@@ -6,10 +6,21 @@ from .logging_config import LoggingConfig
 from .openai_driver_config import OpenAiDriverConfig
 
 
-@define
+@define(kw_only=True)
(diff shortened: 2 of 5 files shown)


collindutter:
Not sure how to unit test this because of schrodinger's initialization. Inspecting the `config.driver` property immediately gives it a value.

vachillo:
[Agent calls `config.drivers` on init](https://github.com/griptape-ai/griptape/blob/dev/griptape/structures/agent.py#L27)

collindutter:
Technically there is no `_drivers` because of the `alias`. I'll remove the attrs magic since users will never create a `_Config` directly. It'll always be used via the setters.

vachillo:
> Technically there is no `_drivers` because of the `alias`. I'll remove the attrs magic since users will never create a `_Config` directly. It'll always be used via the setters.

the alias is just for init isnt it? the actual field still exists on the class.

griptape/config/config.py
```diff
@@ -6,10 +6,21 @@
...
-    logging: LoggingConfig = field(default=Factory(lambda: LoggingConfig()), kw_only=True)
+    logging: LoggingConfig = field(default=Factory(lambda: LoggingConfig()))
+    _drivers: BaseDriverConfig = field(default=None)
+
+    @property
+    def drivers(self) -> BaseDriverConfig:
+        """Lazily instantiates the drivers configuration to avoid client errors like missing API key."""
+        if self._drivers is None:
+            self._drivers = OpenAiDriverConfig()
+        return self._drivers
```
dylanholmes, line 19: encapsulation to the rescue

griptape/config/config.py
```diff
@@ -6,10 +6,21 @@
 from .openai_driver_config import OpenAiDriverConfig
 
 
-@define
+@define(kw_only=True)
 class _Config(BaseConfig):
-    drivers: BaseDriverConfig = field(default=Factory(lambda: OpenAiDriverConfig()), kw_only=True)
-    logging: LoggingConfig = field(default=Factory(lambda: LoggingConfig()), kw_only=True)
+    logging: LoggingConfig = field(default=Factory(lambda: LoggingConfig()))
```
dylanholmes, line 11: Should we be proactive and do this for logging too?

dylanholmes (approved):
I can't think of any reason this would be bad off the top of my head, but it makes me a little nervous

collindutter:
Ok I really don't know how to test this because of this [root conftest](https://github.com/griptape-ai/griptape/blob/8b9e10c7eeb37e1ddca85c2fd4cc45f2d2595a0a/tests/unit/conftest.py#L17-L21) which immediately initializes the value for all tests. According to Codecov we're already covered here 🤷‍♂️

griptape/config/config.py
collindutter: I think I'd rather avoid introducing weird hacks like this if we can avoid it. Until there's a clear reason to introduce for logging, I say we avoid -- the API won't change either way.

vachillo:
> Ok I really don't know how to test this because of this [root conftest](https://github.com/griptape-ai/griptape/blob/8b9e10c7eeb37e1ddca85c2fd4cc45f2d2595a0a/tests/unit/conftest.py#L17-L21) which immediately initializes the value for all tests. According to Codecov we're already covered here 🤷‍♂️

what about setting it to `None` at the top of the test file and then importing everything? including setting `os.environ["OPENAI_API_KEY"] = None`

griptape/config/config.py
```diff
@@ -6,10 +6,21 @@
...
+    _drivers: BaseDriverConfig = field(default=None)
```
vachillo, line 12: [fyi on aliases (aliai)](https://www.attrs.org/en/stable/init.html#private-attributes-and-aliases)
> attrs follows the doctrine that [there is no such thing as a private argument](https://github.com/hynek/characteristic/issues/6) and strips the underscores from the name when writing the __init__ method signature


so removing `alias=drivers` makes no difference and its just less obvious

dylanholmes (approved):
Seems fine enough, but added a couple of prodding questions

griptape/config/base_config.py
```diff
@@ -1,14 +1,22 @@
...
 
-from .base_driver_config import BaseDriverConfig
-from .logging_config import LoggingConfig
+if TYPE_CHECKING:
+    from .base_driver_config import BaseDriverConfig
+    from .logging_config import LoggingConfig
 
 
 @define(kw_only=True)
 class BaseConfig(SerializableMixin, ABC):
```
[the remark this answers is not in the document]
collindutter: Agreed that it's probably redundant here (and maybe even adds some confusion), but I think I'd prefer to keep consistent with the pattern of having a base class for any inheritance hierarchy.

Who knows what the config classes may evolve into over time.

tests/unit/conftest.py
```diff
@@ -15,7 +15,19 @@ def mock_event_bus():
...
 
 @pytest.fixture(autouse=True)
-def mock_config():
+def mock_config(request):
+    from griptape.config import config
+
+    config.reset()
+
+    # Some tests we don't want to use the autouse fixture's MockDriverConfig
+    if "skip_autouse_fixture" in request.keywords:
```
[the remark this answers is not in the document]
collindutter: Updated the name to `skip_mock_config`

vachillo (approved):

PR #1075, opened 2024-08-19
Lazy load driver config fields

collindutter:

## Describe your changes
Lazy Load Driver Config fields to avoid instantiation of Drivers if they're not used.
NA

### DIFF
--- a/griptape/config/drivers/base_driver_config.py
+++ b/griptape/config/drivers/base_driver_config.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
-from abc import ABC
+from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Optional
 
 from attrs import define, field
 
 from griptape.mixins import SerializableMixin
+from griptape.utils.decorators import lazy_property
 
 if TYPE_CHECKING:
     from griptape.drivers import (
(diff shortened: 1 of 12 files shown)


griptape/utils/decorators.py
```diff
@@ -27,3 +28,20 @@ def wrapper(self: Any, *args, **kwargs) -> Any:
         return wrapper
 
     return decorator
+
+
+def lazy_property(func: Callable[[Any], T]) -> property:
+    attr_name = f"_{func.__name__}"
```
vachillo, line 34: make this an override-able argument?

griptape/config/drivers/base_driver_config.py
```diff
@@ -22,15 +23,57 @@
...
+        default=None, kw_only=True, metadata={"serializable": True}, alias="conversation_memory"
+    )
+    _text_to_speech: BaseTextToSpeechDriver = field(
+        default=None, kw_only=True, metadata={"serializable": True}, alias="text_to_speech"
+    )
+    _audio_transcription: BaseAudioTranscriptionDriver = field(
+        default=None, kw_only=True, metadata={"serializable": True}, alias="audio_transcription"
+    )
+
+    @lazy_property()
```
dylanholmes, line 49: Is this required on the abstract method? (Like is it functionally required vs just for documentation? What happens if a subclass forgets it?)

dylanholmes (dismissed):
This is a lot cleaner, just one question

griptape/config/drivers/base_driver_config.py
collindutter: It's not technically required, but I think it's good for documentation. If a subclass forgets it, everything should still work fine but that field will not be lazy loaded.

dylanholmes (approved):

PR #1093, opened 2024-08-21
Improve ListArtifact

collindutter:

### Added
- Generic type support to `ListArtifact`.
- Iteration support to `ListArtifact`.
NA

### DIFF
--- a/griptape/artifacts/list_artifact.py
+++ b/griptape/artifacts/list_artifact.py
@@ -9,7 +9,7 @@ from griptape.artifacts import BaseArtifact
 if TYPE_CHECKING:
     from collections.abc import Iterator, Sequence
 
-T = TypeVar("T", bound=BaseArtifact)
+T = TypeVar("T", bound=BaseArtifact, covariant=True)
 
 
 @define
--- a/griptape/artifacts/list_artifact.py
+++ b/griptape/artifacts/list_artifact.py
@@ -1,23 +1,25 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Generic, Optional, TypeVar
 
 from attrs import Attribute, define, field
 
 from griptape.artifacts import BaseArtifact
 
 if TYPE_CHECKING:
-    from collections.abc import Sequence
+    from collections.abc import Iterator, Sequence
+
+T = TypeVar("T", bound=BaseArtifact)
 
 
 @define
-class ListArtifact(BaseArtifact):
-    value: Sequence[BaseArtifact] = field(factory=list, metadata={"serializable": True})
+class ListArtifact(BaseArtifact, Generic[T]):
+    value: Sequence[T] = field(factory=list, metadata={"serializable": True})
     item_separator: str = field(default="\n\n", kw_only=True, metadata={"serializable": True})
     validate_uniform_types: bool = field(default=False, kw_only=True, metadata={"serializable": True})
 
     @value.validator  # pyright: ignore[reportAttributeAccessIssue]
-    def validate_value(self, _: Attribute, value: list[BaseArtifact]) -> None:
+    def validate_value(self, _: Attribute, value: list[T]) -> None:
         if self.validate_uniform_types and len(value) > 0:
             first_type = type(value[0])
 
@@ -31,17 +33,20 @@ class ListArtifact(BaseArtifact):
         else:
             return None
 
-    def __getitem__(self, key: int) -> BaseArtifact:
+    def __getitem__(self, key: int) -> T:
         return self.value[key]
 
     def __bool__(self) -> bool:
         return len(self) > 0
 
+    def __add__(self, other: BaseArtifact) -> ListArtifact[T]:
+        return ListArtifact(self.value + other.value)
+
+    def __iter__(self) -> Iterator[T]:
+        return iter(self.value)
+
     def to_text(self) -> str:
         return self.item_separator.join([v.to_text() for v in self.value])
-
-    def __add__(self, other: BaseArtifact) -> BaseArtifact:
-        return ListArtifact(self.value + other.value)
 
     def is_type(self, target_type: type) -> bool:
         if self.value:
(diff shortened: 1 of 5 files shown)


griptape/artifacts/list_artifact.py
```diff
@@ -31,18 +34,21 @@ def child_type(self) -> Optional[type]:
...
             return None
 
-    def __getitem__(self, key: int) -> BaseArtifact:
+    def __getitem__(self, key: int) -> T:
         return self.value[key]
 
     def __bool__(self) -> bool:
         return len(self) > 0
 
+    def __add__(self, other: BaseArtifact) -> ListArtifact[T]:
```
vachillo, line 43: should this be `T` as well?

collindutter: No, otherwise the override will be incompatible with the base class.

vachillo: feels inconsistent then

collindutter: A generic does not belong here. When adding something to the `ListArtifact`, it should be a `BaseArtifact`, not whatever we define in the generic.

From pyright:
```
Method "__add__" overrides class "BaseArtifact" in an incompatible manner
     Parameter 2 type mismatch: base parameter is type "BaseArtifact", override parameter is type "T@ListArtifact"
       Type "BaseArtifact" is incompatible with type "T@ListArtifact" [reportIncompatibleMethodOverride]
```

griptape/artifacts/list_artifact.py
```diff
@@ -9,15 +10,17 @@
 if TYPE_CHECKING:
     from collections.abc import Sequence
 
+T = TypeVar("T", bound=BaseArtifact)
```
vachillo, line 13: what is the benefit to the generic type if its bound to `BaseArtifact`?

vachillo: i understand,  but what is the benefit of that vs what is currently there, ie:
```
value: Sequence[BaseArtifact]
```
is that not the same thing?

collindutter: It is not the same thing. If we don't use the generic there it will not catch a scenario like this:

```python
assert ListArtifact[BlobArtifact]([TextArtifact("foo")])
```

vachillo: i see, its for type hinting. makes sense now

griptape/artifacts/list_artifact.py
```diff
@@ -1,23 +1,25 @@
...
-    value: Sequence[BaseArtifact] = field(factory=list, metadata={"serializable": True})
+class ListArtifact(BaseArtifact, Generic[T]):
+    value: Sequence[T] = field(factory=list, metadata={"serializable": True})
     item_separator: str = field(default="\n\n", kw_only=True, metadata={"serializable": True})
     validate_uniform_types: bool = field(default=False, kw_only=True, metadata={"serializable": True})
 
     @value.validator  # pyright: ignore[reportAttributeAccessIssue]
-    def validate_value(self, _: Attribute, value: list[BaseArtifact]) -> None:
+    def validate_value(self, _: Attribute, value: list[T]) -> None:
         if self.validate_uniform_types and len(value) > 0:
```
vachillo, line 23: is `validate_uniform_types` still applicable here?

collindutter: I think `validate_uniform_types` is still useful as a runtime check.

collindutter: I think `validate_uniform_types` is still useful as a runtime check (although we don't actually use it anywhere lol). In your example, it would fail type check but succeed at runtime.

dylanholmes (commented):
Awesome!

griptape/artifacts/list_artifact.py
```diff
@@ -1,23 +1,25 @@
...
 
 from attrs import Attribute, define, field
 
 from griptape.artifacts import BaseArtifact
 
 if TYPE_CHECKING:
-    from collections.abc import Sequence
+    from collections.abc import Iterator, Sequence
+
+T = TypeVar("T", bound=BaseArtifact)
```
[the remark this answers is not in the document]
collindutter: Great suggestion! Will update.

griptape/artifacts/list_artifact.py
```diff
@@ -1,23 +1,25 @@
...
 
 
 @define
-class ListArtifact(BaseArtifact):
-    value: Sequence[BaseArtifact] = field(factory=list, metadata={"serializable": True})
+class ListArtifact(BaseArtifact, Generic[T]):
+    value: Sequence[T] = field(factory=list, metadata={"serializable": True})
```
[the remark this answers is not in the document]
collindutter: `list` is not covariant, `Sequence` is. Here is an example pyright error:

```
    Return type mismatch: base method returns type "BaseArtifact | list[BaseArtifact]", override returns type "list[CsvRowArtifact]"
      Type "list[CsvRowArtifact]" is incompatible with type "BaseArtifact | list[BaseArtifact]"
        "list[CsvRowArtifact]" is incompatible with "BaseArtifact"
        "list[CsvRowArtifact]" is incompatible with "list[BaseArtifact]"
          Type parameter "_T@list" is invariant, but "CsvRowArtifact" is not the same as "BaseArtifact"
          Consider switching from "list" to "Sequence" which is covariant (reportIncompatibleMethodOverride)
```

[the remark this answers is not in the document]
dylanholmes: TIL!

After looking into why `list` is not covariant and why `Sequence` is, its because `list` is mutable!

example:
```py
class Animal:
    pass

class Dog(Animal):
    pass

dogs: List[Dog] = [Dog(), Dog()]
animals: List[Animal] = dogs  # This would be allowed if lists were covariant

animals.append(Animal())  # Adding an Animal (not necessarily a Dog!) to a list of Dogs
```

So this leads me to the question, `ListArtifact` is not intended to be mutable, is it?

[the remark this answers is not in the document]
collindutter: > ListArtifact is not intended to be mutable, is it?

Good question, probably not. I can investigate changing back to `list`, but it will have implications throughout the codebase so I think it should be done in a separate PR.

collindutter:
Many Artifact refactors incoming

PR #1101, opened 2024-08-23
Remove many instances of catching exceptions

collindutter:

### Changed
- **BREAKING**: Drivers, Loaders, and Engines will now raises exceptions rather than returning `ErrorArtifact`s.

Note that this change was not made to Tools, Tasks, or Structures for the following reasons:
- Tools: Tool output always goes to the LLM, therefore if a Tool errors out we want to convey that to the LLM so that it can course correct. A Tool does not need to return `ErrorArtifact`s on its own, it can raise an exception which will be [caught and converted](https://github.com/griptape-ai/griptape/blob/be18083838eb5dc3c5db780ec43eb966b4e0ac52/griptape/tools/base_tool.py#L120-L129) in BaseTool.
- Tasks & Structures: it is useful for Tasks and Structures to have an output value even if they errored. If [fail_fast](https://github.com/griptape-ai/griptape/blob/be18083838eb5dc3c5db780ec43eb966b4e0ac52/griptape/structures/structure.py#L37) is disabled, the Structure will continue running and you can inspect the output of all Tasks (successful or not) afterwards.

NA

----
📚 Documentation preview 📚: https://griptape--1101.org.readthedocs.build//1101/

### DIFF
--- a/griptape/tasks/code_execution_task.py
+++ b/griptape/tasks/code_execution_task.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
-from typing import Callable
+from typing import TYPE_CHECKING, Callable
 
 from attrs import define, field
 
-from griptape.artifacts import BaseArtifact, ErrorArtifact
 from griptape.tasks import BaseTextInputTask
+
+if TYPE_CHECKING:
+    from griptape.artifacts import BaseArtifact
 
 
 @define
@@ -13,7 +15,4 @@ class CodeExecutionTask(BaseTextInputTask):
     run_fn: Callable[[CodeExecutionTask], BaseArtifact] = field(kw_only=True)
 
     def run(self) -> BaseArtifact:
-        try:
-            return self.run_fn(self)
-        except Exception as e:
-            return ErrorArtifact(f"error during Code Execution Task: {e}")
+        return self.run_fn(self)
--- a/griptape/tools/file_manager/tool.py
+++ b/griptape/tools/file_manager/tool.py
@@ -93,9 +93,16 @@ class FileManagerTool(BaseTool):
 
         for artifact in list_artifact.value:
             formatted_file_name = f"{artifact.name}-{file_name}" if len(list_artifact) > 1 else file_name
-            result = self.file_manager_driver.save_file(os.path.join(dir_name, formatted_file_name), artifact.value)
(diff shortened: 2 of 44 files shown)


griptape/tasks/code_execution_task.py
```diff
@@ -1,19 +1,18 @@
...
 @define
 class CodeExecutionTask(BaseTextInputTask):
     run_fn: Callable[[CodeExecutionTask], BaseArtifact] = field(kw_only=True)
 
     def run(self) -> BaseArtifact:
-        try:
-            return self.run_fn(self)
-        except Exception as e:
-            return ErrorArtifact(f"error during Code Execution Task: {e}")
+        return self.run_fn(self)
```
vachillo, line 18: other tasks dont do this do they?

griptape/tools/file_manager/tool.py
```diff
@@ -93,9 +93,16 @@ def save_memory_artifacts_to_disk(self, params: dict) -> ErrorArtifact | InfoArt
...
+            try:
+                self.file_manager_driver.save_file(os.path.join(dir_name, formatted_file_name), artifact.value)
+            except FileNotFoundError:
+                return ErrorArtifact("Path not found")
+            except IsADirectoryError:
+                return ErrorArtifact("Path is a directory")
+            except NotADirectoryError:
+                return ErrorArtifact("Not a directory")
+            except Exception as e:
+                return ErrorArtifact(f"Failed to load file: {str(e)}")
```
vachillo, line 105: should this just be the one `except` statement?

vachillo (approved):

griptape/tasks/code_execution_task.py
collindutter: They do not which is why I removed it. `BaseTask` [will catch this already](https://github.com/griptape-ai/griptape/blob/f2b529ab361c33b2a26a4510f78f85181b96af4f/griptape/tasks/base_task.py#L163-L178).

griptape/tools/file_manager/tool.py
collindutter: I don't think so. We want to convey each of the exceptions to the LLM in a slightly different way. These are copied directly from the Driver that previously threw them.

PR #1114, opened 2024-08-28
Refactor Artifacts

collindutter:

## Describe your changes
<s>### Added
- `BaseArtifact.to_bytes()` method to convert an Artifact to bytes.

### Changed
- **BREAKING**: Removed `MediaArtifact`, use `ImageArtifact` or `AudioArtifact` instead.
- **BREAKING**: Removed `BooleanArtifact`, use `JsonArtifact` instead.
- **BREAKING**: Removed `CsvRowArtifact`.
- **BREAKING**: `CsvLoader`, `DataframeLoader`, and `SqlLoader` now return `list[TextArtifact]`.
- **BREAKING**: Removed `ImageArtifact.media_type`.
- **BREAKING**: Removed `AudioArtifact.media_type`.
- **BREAKING**: Moved `ImageArtifact.prompt` and `ImageArtifact.model` into `ImageArtifact.meta`.
- **BREAKING**: `ImageArtifact.to_text()` now returns the base64 encoded image.
- Updated `JsonArtifact` value converter to properly handle more types. 
- `AudioArtifact` now subclasses `BaseArtifact` instead of `MediaArtifact`.
- `ImageArtifact` now subclasses `BaseArtifact` instead of `MediaArtifact`.
- Passing a dictionary as the value to `TextArtifact` will convert to a key-value formatted string. 
- Removed `__add__` method from `BaseArtifact`, implemented it where necessary.

The main goal of this PR was to clearly define what Artifacts are and remove any Artifacts that don't fit this definition.

Artifacts are an abstraction for passing data around in the framework. They fall into two categories:

1. **LLM input/output**: Artifacts that contain values which can be directly passed to LLMs, or that LLMs can directly output.
   - **TextArtifact**: LLMs can input/output text (duh).
   - **ImageArtifact**: LLMs can sometimes input/output images.
   - **AudioArtifact**: LLMs can sometimes input/output audio.
   - **JsonArtifact**: LLMs can sometimes output structured JSON data.
   - **BlobArtifact**: LLMs can sometimes input binary data ([Bedrock's documents](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html), probably Anthropic Artifacts).
   - **GenericArtifact**: LLMs can sometimes input arbitrary data (e.g., Google's video analysis).
   - **ActionArtifact**: LLMs can input/output function calls.
   
While we could introduce an ABC to represent this category of Artifacts, this category feels like the "default" type of Artifact.

2. **System Artifacts**: Artifacts that don't represent a modality but are still useful to have. They must be transformed in some way before they can be used as LLM input/output.
   - **InfoArtifact**: Must be transformed into a TextArtifact before being used as LLM input.
   - **ErrorArtifact**: Must be transformed into a TextArtifact before being used as LLM input.
   - **ListArtifact**: Must be flattened into TextArtifact/ImageArtifacts before being used as LLM input.

With this in mind, here is why I removed the following:

- **MediaArtifact**: I felt that this class only added hierarchical complexity. We can re-evaluate a similar abstract base class in the future, but it did not seem necessary given we only have `ImageArtifact` and `AudioArtifact`.
- **BooleanArtifact**: This class primarily provided a utility method, `BooleanArtifact.parse_bool`, to convert LLM-outputted strings into boolean values (e.g., "true" -> True). While this is a valuable feature, I don't think it's worth breaking the mental model that Artifacts must be input/outputs for an LLM. Additionally, a more generalized version of this feature will likely be implemented in the upcoming structured output features.
- **CsvRowArtifact**: This Artifact being a TextArtifact always felt odd given its type signature (`value: dict`). If we had been more diligent about banning direct access to `value` in favor of `to_text()`, this might have been less of an issue. But as it stands, this Artifact doesn't provide much value over just cramming the data into a `TextArtifact`. A limitation of this new approach is that we don't have the underlying data available without parsing out the textual value, but I can't find a significant issue with this. If you need the raw data, don't use Griptape Loaders that produce Artifacts.

Some changes that may surprise people:

- **ListArtifact lives on**: There was much talk of removing this Artifact, but I ultimately decided to keep it because it it's deeply integrated into our codebase, and I feel that it still serves a valuable purpose as a system Artifact.
- **No TableArtifact**: This didn't fit the new Artifact model for the same reasons as CsvRowArtifact. There were probably ways to make this work, but it felt like unnecessary complexity. Additionally, chunking a TableArtifact would be challenging.
- **Pointer Artifact**: I think there could still be value in introducing a PointerArtifact that subclasses InfoArtifact, but it did not seem necessary for this PR.</s>

The main goal of this PR is to clearly define what Artifacts are, and remove any Artifacts that don't fit this definition.

The original draft of this PR put Artifacts into two categories: "LLM input/output" and "System." However, this approach presented several issues:
- "System" Artifacts lacked a clear definition, essentially becoming a catch-all for anything that wasn’t an LLM input/output Artifact.
- Artifacts became too tightly coupled to LLMs, even though they are used by many non-LLM components in Griptape.

The updated definition of Artifacts is: Artifacts are the core data structure in Griptape, used to encapsulate data and enrich it with metadata.

Regarding the relationship between Artifacts and LLMs, the [Prompt Stack](https://github.com/griptape-ai/griptape/blob/dev/griptape/common/prompt_stack/prompt_stack.py#L64) converts Artifacts into [Message Content](https://github.com/griptape-ai/griptape/tree/dev/griptape/common/prompt_stack/contents), bridging the gap.
It’s important to note that since Message Contents still _use_ Artifacts to encapsulate data, each Message Content type requires a corresponding Artifact type. This is _ok_ and an intended use of Artifacts.

Although `MediaArtifact` technically meets the new definition of Artifacts, it was removed for the reason mentioned in the original draft.

In a future PR, we can evaluate whether introducing new abstract base classes would simplify the process for users to create custom Artifacts without needing to subclass the existing concrete Artifacts.

----
📚 Documentation preview 📚: https://griptape--1114.org.readthedocs.build//1114/


griptape/artifacts/generic_artifact.py
```diff
@@ -9,7 +9,13 @@
 
 @define
 class GenericArtifact(BaseArtifact):
+    """Serves as an escape hatch for artifacts that don't fit into any other category.
+
+    Attributes:
+        value: The value of the Artifact.
+    """
+
     value: Any = field(metadata={"serializable": True})
```
vachillo, line 19: is this a good opportunity to update this artifact to `Artifact` and making it the top-level base? [this pattern](https://github.com/griptape-ai/griptape/blob/dev/griptape/configs/drivers/drivers_config.py) is [used](https://github.com/griptape-ai/griptape/blob/dev/griptape/memory/structure/conversation_memory.py) in a few other places.

griptape/artifacts/text_artifact.py
```diff
@@ -11,23 +11,36 @@
...
+    if isinstance(value, dict):
+        return "\n".join(f"{key}: {val}" for key, val in value.items())
+    else:
+        return str(value)
+
+
 @define
 class TextArtifact(BaseArtifact):
-    value: str = field(converter=str, metadata={"serializable": True})
+    value: str = field(converter=value_to_str, metadata={"serializable": True})
```
vachillo, line 23: any reason for the custom method? cant we just do 
```
value: str = field(converter=str, metadata={"serializable": True})
```
?

griptape/artifacts/text_artifact.py
```diff
@@ -11,23 +11,36 @@
...
-
     def __add__(self, other: BaseArtifact) -> TextArtifact:
         return TextArtifact(self.value + other.value)
 
     def __bool__(self) -> bool:
         return bool(self.value.strip())
 
+    @property
+    def embedding(self) -> Optional[list[float]]:
+        return None if len(self._embedding) == 0 else self._embedding
```
vachillo, line 36: should we just default `_embedding` to `None` and rename the field?

vachillo (commented):
> the mental model that Artifacts must be input/outputs for an LLM

I'm not sure if I agree with this model. `Artifacts` currently used as inputs and outputs to `Tasks`, and as such, need to be directly interacted with by developers, especially for more complex `Pipelines`/`Workflows`. 

The existence of artifacts like `BooleanArtifact` are conveniences for allowing `Tasks` to declare their own outputs and tell downstream `Tasks` how to interpret the output. Without it, subsequent tasks need to do parsing/value checks, which couples tasks tightly together. 

For example, if I have a task that I want to send some python `bool` value to the LLM, why should I care what data format the LLM takes? The framework should abstract that. I don't think I should be forced to convert that into text myself if I don't have to.

Same thing with output. I want to ask the LLM a yes/no question and I am expecting a boolean value. The framework should be able to send back a python representation of a `bool` without me having to parse throught text myself every time. Especially useful for downstream tasks in a `Pipeline`/`Workflow`.

griptape/artifacts/generic_artifact.py
collindutter: I think we should avoid subclassing concrete classes (even though we do it in this PR). So the other places are in violation of this and should be fixed IMO.

griptape/artifacts/text_artifact.py
collindutter: Note the special formatting for dictionaries:
```python
    if isinstance(value, dict):
        return "\n".join(f"{key}: {val}" for key, val in value.items())
```
This removes a lot of the dictionary noise when conveying the artifact as text.

griptape/artifacts/text_artifact.py
collindutter: Yeah I'm not really sure what this pattern gives us, besides having to check for `None`.

collindutter: What does that get us specifically? I think the intended behavior here is to distinguish between non-generated vs empty embeddings, right?

collindutter:
Re-refactored with @vachillo's suggestions in mind.

griptape/artifacts/text_artifact.py
vachillo: its just less verbose and would allow us to remove the `@property` method while accomplishing the same thing

vachillo:
```
embedding: Optional[list[float]] = field(default=None, kw_only=True)
```
should be functionally equivalent to this `@property` no?

collindutter: Updated

griptape/artifacts/text_artifact.py
vachillo: i see. i dont disagree that its probably better for the LLM, but do we want to commit to potentially doing this for a bunch of other types too? what about `set` or `list`?

collindutter: Good point. Maybe this should live on the loader that creates the Artifact then?

griptape/artifacts/generic_artifact.py
vachillo: any particular reason for that? Is subclassing concrete types a bad thing?

collindutter: It adds depth to the class hierarchy which can be difficult to maintain. There are times when it's unavoidable, but while we're still changing things around I think we should strive for a relatively flat hierarchy.

griptape/artifacts/text_artifact.py
vachillo: yeah that probably makes sense. the loader can choose to do special formatting tricks if needed.

vachillo: although im thinking that it would probably be _best_ to do it when it gets converted to a `MessageContent`, thats probably out of scope and too complex for this right now.

collindutter: That conversion happens by calling `Artifact.to_text()`. Honestly I think `CsvRowArtifact` actually works with this new model of Artifacts, but there are still too many issues with `value: dict`.

We could make `value` a property that automatically calls `to_text()`, but that feels like a hack.

vachillo: you mean changing value on the base class to:
```
_value: Any = field()

@property
def value(self):
    return self._value
```
that actually seems like a great idea.

vachillo: that would give child artifacts way more flexibility on init and when `.value` is called

collindutter: I meant updating `CsvRowArtifact` specifically to something like:
```python
_value: dict = field()

@property
def value(self):
    return self.to_text()
```

Not sure I fully follow your suggestion.

vachillo: i am suggesting the same thing, because the base class needs to be updated with:
```
_value: Any = field()

@property
def value(self):
    return self._value
```
in order for your suggestion to be possible.

I tried basically this same thing for `JsonArtifact` and i couldn't do it because of the base class

griptape/artifacts/base_artifact.py
```diff
@@ -24,22 +35,7 @@ class BaseArtifact(SerializableMixin, ABC):
...
-
-    @classmethod
-    def value_to_dict(cls, value: Any) -> dict:
-        dict_value = value if isinstance(value, dict) else json.loads(value)
-
-        return dict(dict_value.items())
-
-    def to_text(self) -> str:
-        return str(self.value)
+    encoding: str = field(default="utf-8", kw_only=True)
```
vachillo, line 38: dont see this in the docstring. do we need this? can it just be passed as a param directly to `to_text`? if not, maybe we can rename it to `text_encoding` or `to_text_encoding`.

griptape/artifacts/blob_artifact.py
```diff
@@ -1,26 +1,39 @@
...
 from attrs import define, field
 
 from griptape.artifacts import BaseArtifact
 
 
+def value_to_bytes(value: Any) -> bytes:
+    if isinstance(value, bytes):
+        return value
+    else:
+        return str(value).encode()
```
vachillo, line 14: should we move these converters to `utils`?

griptape/artifacts/csv_row_artifact.py
```diff
@@ -1,34 +1,28 @@
...
 
-@define
-class CsvRowArtifact(TextArtifact):
-    value: dict[str, str] = field(converter=BaseArtifact.value_to_dict, metadata={"serializable": True})
-    delimiter: str = field(default=",", kw_only=True, metadata={"serializable": True})
+def value_to_str(value: Any) -> str:
+    if isinstance(value, dict):
+        return "\n".join(f"{key}: {val}" for key, val in value.items())
+    else:
+        return str(value)
```
vachillo, line 15: do we still want to do special `dict` handling here?

griptape/artifacts/image_artifact.py
```diff
@@ -1,23 +1,38 @@
...
-        prompt: Optionally specify the prompt used to generate the media.
+        value: The image data.
+        format: The format of the image data. Defaults to "png".
+        width: The width of the image.
+        height: The height of the image
     """
 
-    media_type: str = "image"
+    value: bytes = field(metadata={"serializable": True})
+    format: str = field(default="png", kw_only=True, metadata={"serializable": True})
```
vachillo, line 22: probably should require the format

griptape/artifacts/image_artifact.py
```diff
@@ -1,23 +1,38 @@
...
-        model: Optionally specify the model used to generate the media.
-        prompt: Optionally specify the prompt used to generate the media.
+        value: The image data.
+        format: The format of the image data. Defaults to "png".
+        width: The width of the image.
+        height: The height of the image
     """
 
-    media_type: str = "image"
+    value: bytes = field(metadata={"serializable": True})
```
vachillo, line 21: does this need the bytes converter here?

griptape/artifacts/list_artifact.py
```diff
@@ -1,23 +1,37 @@
...
-    from collections.abc import Sequence
+    from collections.abc import Iterator, Sequence
+
+T = TypeVar("T", bound=BaseArtifact, covariant=True)
 
 
 @define
-class ListArtifact(BaseArtifact):
-    value: Sequence[BaseArtifact] = field(factory=list, metadata={"serializable": True})
+class ListArtifact(BaseArtifact, Generic[T]):
```
vachillo, line 16: should we implement other common `list` methods? `append`, `extend`, etc

griptape/artifacts/text_artifact.py
```diff
@@ -16,26 +16,28 @@ class TextArtifact(BaseArtifact):
...
-        self._embedding.extend(driver.embed_string(str(self.value)))
+    def to_text(self) -> str:
+        return self.value
+
+    def to_bytes(self) -> bytes:
+        return str(self.value).encode(encoding=self.encoding, errors=self.encoding_error_handler)
+
+    def generate_embedding(self, driver: BaseEmbeddingDriver) -> list[float]:
+        if self.embedding is None:
+            self.embedding = []
```
vachillo, line 35: should we keep it `None` until the embedding driver returns? could be an error

griptape/tasks/tool_task.py
```diff
@@ -84,7 +84,11 @@ def run(self) -> BaseArtifact:
             subtask.after_run()
 
             if isinstance(subtask.output, ListArtifact):
-                self.output = subtask.output[0]
+                first_artifact = subtask.output[0]
+                if isinstance(first_artifact, BaseArtifact):
+                    self.output = first_artifact
+                else:
+                    self.output = ErrorArtifact(f"Output is not an Artifact: {type(subtask.output[0])}")
```
vachillo, line 91: should this throw an error or is this a case we want to catch

griptape/tasks/base_audio_generation_task.py
```diff
@@ -6,14 +6,14 @@
...
-from griptape.mixins import BlobArtifactFileOutputMixin, RuleMixin
+from griptape.mixins import ArtifactFileOutputMixin, RuleMixin
 from griptape.tasks import BaseTask
 
 logger = logging.getLogger(Defaults.logging_config.logger_name)
 
 
 @define
-class BaseAudioGenerationTask(BlobArtifactFileOutputMixin, RuleMixin, BaseTask, ABC):
+class BaseAudioGenerationTask(ArtifactFileOutputMixin, RuleMixin, BaseTask, ABC):
```
vachillo, line 16: out of scope for thisPR, but should we add some kind of optional persistence driver to `BaseTask` rather than have this mixin?

griptape/artifacts/text_artifact.py
collindutter: Updated to have this use-case specific logic on `CsvRowArtifact`

griptape/artifacts/blob_artifact.py
collindutter: I'm not sure, some of them do conversions very specific to the Artifact type. `CsvRowArtifact`, for instance.

I just updated the logic to live in the class, let me know what you think.

griptape/artifacts/csv_row_artifact.py
collindutter: I think so. Because now the logic is isolated to this very specific Artifact rather than the much more broad `TextArtifact`.

griptape/artifacts/image_artifact.py
collindutter: Updated

griptape/tasks/tool_task.py
collindutter: Good point, maybe this is worth raising.

griptape/artifacts/text_artifact.py
collindutter: Great catch.

griptape/tasks/base_audio_generation_task.py
collindutter: Yes totally

griptape/artifacts/list_artifact.py
collindutter: Maybe in a future PR. This change was only made because the other loader PR required it.

griptape/artifacts/image_artifact.py
collindutter: Added.

griptape/artifacts/base_artifact.py
collindutter: Just updated the docstring and added a missing implementation. This field already existed in a number of Artifacts, I just bumped to the top since we now have a `to_bytes`.

griptape/artifacts/image_artifact.py
collindutter: Added

griptape/artifacts/blob_artifact.py
vachillo: the `CsvRowArtifact` one makes sense. im still not convinced that this one is class-specific enough. can we one-line it until it doesnt have to be?
```
converter=lambda x: x if isinstance(x, bytes) else str(x).encode()
```
?

collindutter: Sure, just updated.

griptape/artifacts/audio_artifact.py
```diff
@@ -1,12 +1,26 @@
...
+
+    @property
+    def mime_type(self) -> str:
+        return f"audio/{self.format}"
+
+    def to_bytes(self) -> bytes:
+        return self.value
+
+    def to_text(self) -> str:
+        return f"Audio, format: {self.format}, size: {len(self.value)} bytes"
```
vachillo, line 26: > Just updated the docstring and added a missing implementation. This field already existed in a number of Artifacts, I just bumped to the top since we now have a to_bytes.

how does the top level encoding work here?

collindutter: Yeah it's not used at all. I really don't like using `init=False` to remove fields in subclasses. What about updating the docstring to explain this behavior?

Or do you think encoding doesn't belong at the top?

griptape/artifacts/base_artifact.py
vachillo: i would say we could just pass `encoding`/`encoding_error_handler` directly to the `to_bytes`/`to_text` methods, and have them default to the same thing as whatever `encode`/`decode` defaults to.

griptape/artifacts/audio_artifact.py
vachillo: i would say we could just pass encoding/encoding_error_handler directly to the to_bytes/to_text methods, and have them default to the same thing as whatever encode/decode defaults to.

collindutter: I'm now realizing Loaders also take an encoding and set it on the Artifact.

```
TextLoader(encoding="ascii").load("my_file.txt").to_bytes() # should use the Loader-provided `ascii` encoding
```

Maybe these fields should just stay on `TextArtifact` and `BlobArtifact`, but then we'd have to make `to_bytes` abstract which feels like a barrier for creating new Artifact types.

vachillo (approved):

vachillo (approved):

vachillo (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- `BaseArtifact.to_bytes()` method to convert an Artifact to bytes.
```
vachillo, line 10: not: should this be `Artifact value to bytes`?

collindutter: Updated

CHANGELOG.md
```diff
@@ -6,6 +6,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **BREAKING**: `CsvLoader`, `DataframeLoader`, and `SqlLoader` now return `list[TextArtifact]`.
+- **BREAKING**: Removed `ImageArtifact.media_type`.
+- **BREAKING**: Removed `AudioArtifact.media_type`.
+- **BREAKING**: Removed `BlobArtifact.dir_name`.
+- **BREAKING**: Moved `ImageArtifact.prompt` and `ImageArtifact.model` into `ImageArtifact.meta`.
+- **BREAKING**: `ImageArtifact.format` is now required.
+- Updated `JsonArtifact` value converter to properly handle more types. 
+- `AudioArtifact` now subclasses `BlobArtifact` instead of `MediaArtifact`.
+- `ImageArtifact` now subclasses `BlobArtifact` instead of `MediaArtifact`.
+- Passing a dictionary as the value to `TextArtifact` will convert to a key-value formatted string.
```
vachillo, line 25: should be `CsvRowArtifact` right?

collindutter: Good catch updated

vachillo (dismissed):

CHANGELOG.md
```diff
@@ -6,6 +6,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- `BaseArtifact.to_bytes()` method to convert an Artifact's value to bytes.
+- `BlobArtifact.base64` property for converting a `BlobArtifact`'s value to a base64 strings.
```
vachillo, line 11: hyper nit: `string` not `strings`

griptape/engines/extraction/csv_extraction_engine.py
```diff
@@ -20,6 +20,9 @@ class CsvExtractionEngine(BaseExtractionEngine):
     column_names: list[str] = field(default=Factory(list), kw_only=True)
     system_template_generator: J2 = field(default=Factory(lambda: J2("engines/extraction/csv/system.j2")), kw_only=True)
     user_template_generator: J2 = field(default=Factory(lambda: J2("engines/extraction/csv/user.j2")), kw_only=True)
+    formatter_fn: Callable[[Any], str] = field(
+        default=lambda value: "\n".join(f"{key}: {val}" for key, val in value.items()), kw_only=True
+    )
```
vachillo, line 26: it is not immediately obvious that a dictionary will be passed to this function. 
should it be renamed to make it more clear? or move the `dict(zip(...` logic to the default formatter?

collindutter: Type hint should've been Callable[[dict], str], updated

CHANGELOG.md
collindutter: Fixed

vachillo (approved):

griptape/artifacts/base_artifact.py
```diff
@@ -15,6 +14,20 @@
 
 @define
 class BaseArtifact(SerializableMixin, ABC):
+    """Serves as the base class for all Artifacts.
+
+    Artifacts are used to store data that can be provided as input to, or received as output from, a language model (LLM).
```
dylanholmes, line 19: We change this definition, right? Aren't they more like general containers for data + metadata now, right?

griptape/artifacts/csv_row_artifact.py
```diff
@@ -10,25 +9,20 @@
 
 @define
 class CsvRowArtifact(TextArtifact):
```
dylanholmes, line 11: I thought this was being removed?

dylanholmes (changes requested):
Just two comments

griptape/tasks/tool_task.py
```diff
@@ -84,7 +84,11 @@ def run(self) -> BaseArtifact:
             subtask.after_run()
 
             if isinstance(subtask.output, ListArtifact):
-                self.output = subtask.output[0]
+                first_artifact = subtask.output[0]
+                if isinstance(first_artifact, BaseArtifact):
+                    self.output = first_artifact
+                else:
+                    raise ValueError(f"Output is not an Artifact: {type(first_artifact)}")
```
dylanholmes, line 91: Why would this happen? Did you add this after observing an issue?

griptape/schemas/base_schema.py
```diff
@@ -56,6 +56,10 @@ def _get_field_for_type(cls, field_type: type) -> fields.Field | fields.Nested:
 
         field_class, args, optional = cls._get_field_type_info(field_type)
 
+        # Resolve TypeVars to their bound type
+        if isinstance(field_class, TypeVar):
+            field_class = field_class.__bound__
```
dylanholmes, line 61: Why was this needed?

griptape/common/prompt_stack/prompt_stack.py
```diff
@@ -81,6 +78,7 @@ def __to_message_content(self, artifact: str | BaseArtifact) -> list[BaseMessage
                 return [ActionResultMessageContent(output, action=action)]
         elif isinstance(artifact, ListArtifact):
             processed_contents = [self.__to_message_content(artifact) for artifact in artifact.value]
+
             return [sub_content for processed_content in processed_contents for sub_content in processed_content]
         else:
-            raise ValueError(f"Unsupported artifact type: {type(artifact)}")
+            return [TextMessageContent(TextArtifact(artifact.to_text()))]
```
dylanholmes, line 84: Why this change in behavior?

Plus, you updated `TextMessageContent` to accept a `BaseArtifact` instead of `TextArtifact`, so do you need to convert it here at all?

griptape/artifacts/json_artifact.py
```diff
@@ -12,10 +12,20 @@
...
+    """
+
+    value: Json = field(converter=lambda value: JsonArtifact.value_to_json(value), metadata={"serializable": True})
+
+    @classmethod
+    def value_to_json(cls, value: Any) -> Json:
+        if isinstance(value, str):
+            return json.loads(value)
+        else:
+            return json.loads(json.dumps(value))
```
dylanholmes, line 28: Why, dumps then loads? Deep copy?

griptape/artifacts/info_artifact.py
```diff
@@ -7,7 +7,15 @@
 
 @define
 class InfoArtifact(BaseArtifact):
```
dylanholmes, line 9: Is this not a TextArtifact because we don't want it to match `isinstance` on `TextArtifact` in order to make code branching a little easier?

griptape/artifacts/image_artifact.py
```diff
@@ -2,22 +2,29 @@
...
+    """Stores image data.
 
     Attributes:
-        value: Raw bytes representing media data.
-        media_type: The type of media, defaults to "image".
-        format: The format of the media, like png, jpeg, or gif.
-        name: Artifact name, generated using creation time and a random string.
-        model: Optionally specify the model used to generate the media.
-        prompt: Optionally specify the prompt used to generate the media.
+        format: The format of the image data.
```
dylanholmes, line 13: Should clarify that this is part of the mime type.

griptape/artifacts/base_artifact.py
```diff
@@ -50,5 +49,8 @@ def __bool__(self) -> bool:
     def __len__(self) -> int:
         return len(self.value)
 
+    def to_bytes(self) -> bytes:
+        return self.to_text().encode(encoding=self.encoding, errors=self.encoding_error_handler)
```
dylanholmes, line 53: This is basically for code reuse within the most common "Text-like Artifacts"? 

How would/when would I set encoding_error_handler? Have you seen the need arise?

docs/griptape-framework/data/artifacts.md
```diff
@@ -5,60 +5,50 @@ search:
...
+## Info
 
-## Audio
+[InfoArtifact](../../reference/griptape/artifacts/info_artifact.md)s store small pieces of textual information. These are useful for conveying messages about the execution or results of an operation, such as "No results found" or "Operation completed successfully."
 
-An [AudioArtifact](../../reference/griptape/artifacts/audio_artifact.md) allows the Framework to interact with audio content. An Audio Artifact includes binary audio content as well as metadata like format, duration, and prompt and model information for audio returned generative models. It inherits from [BlobArtifact](#blob).
+## JSON
 
-## Boolean
+[JsonArtifact](../../reference/griptape/artifacts/json_artifact.md)s store JSON-serializable data. Any data assigned to the `value` property is processed using `json.dumps(json.loads(value))`.
```
dylanholmes, line 42: I saw a branch in there

dylanholmes (commented):
Ok, more comments, tried to be more thorough

griptape/tasks/tool_task.py
collindutter: This was introduced by adding generics to `ListArtifact`. Changed this to satisfy pyright.

griptape/schemas/base_schema.py
collindutter: Being able to serialize ListArtifact with generics.

griptape/common/prompt_stack/prompt_stack.py
collindutter: Originally it was to allow any Artifact (such as a custom user Artifact) to be passed without having to subclass ours. But maybe we should revert and just have users subclass.

griptape/artifacts/base_artifact.py
collindutter: > This is basically for code reuse within the most common "Text-like Artifacts"?

Most immediately, it's needed to keep this behavior [here](https://github.com/griptape-ai/griptape/blob/35e2461cfd468bb54d5c9d793b29cf3d9bd90d95/griptape/mixins/artifact_file_output_mixin.py#L45).

> How would/when would I set encoding_error_handler? Have you seen the need arise?

It's definitely an edge case field, you can see an example [here](https://stackoverflow.com/questions/3224268/python-unicode-encode-error). I have not run into a use myself, but it was there before so I kept it.

griptape/artifacts/info_artifact.py
collindutter: Correct. If we wanted to re-use code we could create a `BaseTextArtifact` but I'd like to keep things flat and it feels unnecessary at this time.

griptape/artifacts/json_artifact.py
collindutter: Need to convert to a JSON string before we can parse it as one.

griptape/common/prompt_stack/prompt_stack.py
collindutter: Actually, yeah it was to avoid handling Artifacts beyond the ones explicitly matched here. For instance, `ErrorArtifact` or `InfoArtifact`. I am going to revert the change to `text_message_content` though.

vachillo (approved):

dylanholmes (approved):

PR #1162, opened 2024-09-09
Drivers Config Context Manager

collindutter:

## Describe your changes
Added ability to use Drivers Configs as Context Managers for temporarily setting the default Drivers.
NA

----
📚 Documentation preview 📚: https://griptape--1162.org.readthedocs.build//1162/

### DIFF
--- a/docs/griptape-framework/structures/src/config_defaults.py
+++ b/docs/griptape-framework/structures/src/config_defaults.py
@@ -1,0 +1,12 @@
+from griptape.configs import Defaults
+from griptape.configs.drivers import AnthropicDriversConfig, OpenAiDriversConfig
+from griptape.drivers.prompt.anthropic_prompt_driver import AnthropicPromptDriver
+from griptape.structures import Agent
+
+Defaults.drivers_config = OpenAiDriversConfig()  # Default
+openai_agent = Agent()
+
+Defaults.drivers_config = AnthropicDriversConfig()
+anthropic_agent = Agent(
+    prompt_driver=AnthropicPromptDriver(model="claude-3-5-sonnet-20240620"),  # Override the default prompt driver
+)
--- a/docs/griptape-framework/structures/src/drivers_config_with.py
+++ b/docs/griptape-framework/structures/src/drivers_config_with.py
@@ -1,0 +1,31 @@
+import schema
+
+from griptape.configs.drivers import AnthropicDriversConfig, OpenAiDriversConfig
+from griptape.drivers import AnthropicPromptDriver, OpenAiChatPromptDriver
+from griptape.engines import JsonExtractionEngine
+from griptape.structures import Agent
+from griptape.tasks import ToolTask
+from griptape.tools import ExtractionTool
+
+with OpenAiDriversConfig():  # Agent will be created with OpenAi Drivers
+    openai_agent = Agent()
+
+with AnthropicDriversConfig():  # Agent will be created with Anthropic Drivers
+    anthropic_agent = Agent(
+        tasks=[
+            ToolTask(
+                "Extract sentiment from this text: {{ args[0] }}",
+                prompt_driver=OpenAiChatPromptDriver(model="gpt-4o"),  # Override this particular Task's prompt driver
+                tool=ExtractionTool(
+                    extraction_engine=JsonExtractionEngine(
+                        prompt_driver=AnthropicPromptDriver(  # Override this particular Engine's prompt driver
+                            model="claude-3-opus-20240229"
+                        ),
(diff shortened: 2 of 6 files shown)


griptape/configs/drivers/base_drivers_config.py
```diff
@@ -48,6 +48,22 @@ class BaseDriversConfig(ABC, SerializableMixin):
...
+
+        Defaults.drivers_config = self
+
+        return self
+
+    def __exit__(self, type, value, traceback) -> None:  # noqa: ANN001, A002
+        from griptape.configs import Defaults
+
+        Defaults.drivers_config = self._old_drivers_config
+
```
vachillo, line 66: probably should add `self._old_drivers_config = None`

vachillo (commented):
this is cool. maybe the docs example be updated to show creating a custom `DriversConfig` and using that as a context manager?

docs/griptape-framework/structures/src/config_defaults_override.py
collindutter, line 1: @zachgiordano I think this example may be doing too much for what we're trying to illustrate.

What do you think about instead beefing up the [existing custom example](https://docs.griptape.ai/stable/griptape-framework/structures/configs/#custom)? @vachillo what do you think?

griptape/configs/drivers/base_drivers_config.py
collindutter: Updated

docs/griptape-framework/structures/src/config_defaults_override.py
zachgiordano: That sounds good to me

vachillo (approved):

docs/griptape-framework/structures/src/config_defaults.py
```diff
@@ -0,0 +1,12 @@
+from griptape.configs import Defaults
+from griptape.configs.drivers import AnthropicDriversConfig, OpenAiDriversConfig
+from griptape.drivers.prompt.anthropic_prompt_driver import AnthropicPromptDriver
+from griptape.structures import Agent
+
+Defaults.drivers_config = OpenAiDriversConfig()  # Default
+openai_agent = Agent()
+
+Defaults.drivers_config = AnthropicDriversConfig()
```
dylanholmes, line 9: Not sure if it is currently possible, but I'd love to see an example of "clearing" the defaults, so that an error is thrown if you don't set a prompt driver for a specific use case. In some instances, I want to be sure that I'm not accidentally relying on a driver that attempts to makes a network call that I didn't want because of some default that I missed or didn't think about.

Maybe this exists via a `DummyDriversConfig`?

docs/griptape-framework/structures/src/drivers_config_with.py
```diff
@@ -0,0 +1,31 @@
...
+
+from griptape.configs.drivers import AnthropicDriversConfig, OpenAiDriversConfig
+from griptape.drivers import AnthropicPromptDriver, OpenAiChatPromptDriver
+from griptape.engines import JsonExtractionEngine
+from griptape.structures import Agent
+from griptape.tasks import ToolTask
+from griptape.tools import ExtractionTool
+
+with OpenAiDriversConfig():  # Agent will be created with OpenAi Drivers
+    openai_agent = Agent()
```
dylanholmes, line 11: Can you illustrate what is expected for this common scenario:

```py
def create_anthropic_agent():
    with AnthropicDriversConfig():
        return Agent()

anthropic_agent = create_anthropic_agent()

# Are there (and will there never be) any lazily instantiated objects
# inside Agent that could cause this to not use anthropic drivers?
anthropic_agent.run() 
```

dylanholmes (approved):
🎉  Just what I needed!

docs/griptape-framework/structures/src/config_defaults.py
collindutter: That's what [DriversConfig](https://github.com/griptape-ai/griptape/blob/9d9b643976f30fdf0af16b75c9b92503b076888e/griptape/configs/drivers/drivers_config.py#L34-L65) provides -- it's all Dummy Drivers. The ["Custom"](https://docs.griptape.ai/stable/griptape-framework/structures/configs/#custom) portion of this doc touches on this.

docs/griptape-framework/structures/src/drivers_config_with.py
collindutter: > Are there (and will there never be) any lazily instantiated objects inside Agent that could cause this to not use anthropic drivers?
Shouldn't be -- we don't lazily instantiate any objects in Structures.

dylanholmes (approved):

PR #1165, opened 2024-09-10
Add JsonSchemaRule

collindutter:

### Added
- `JsonSchemaRule` for instructing the LLM to output a JSON object that conforms to a schema.
NA

----
📚 Documentation preview 📚: https://griptape--1165.org.readthedocs.build//1165/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Parameter `meta: dict` on `BaseEvent`.
 - `AzureOpenAiTextToSpeechDriver`.
 - Ability to use Event Listeners as Context Managers for temporarily setting the Event Bus listeners.
+- `JsonSchemaRule` for instructing the LLM to output a JSON object that conforms to a schema.
 
 ### Changed
 - **BREAKING**: Drivers, Loaders, and Engines now raise exceptions rather than returning `ErrorArtifacts`.
--- a/docs/griptape-framework/structures/src/json_schema_rule.py
+++ b/docs/griptape-framework/structures/src/json_schema_rule.py
@@ -1,3 +1,5 @@
+import json
+
 import schema
 
 from griptape.rules.json_schema_rule import JsonSchemaRule
@@ -11,4 +13,6 @@ agent = Agent(
     ]
 )
 
-agent.run("What is the sentiment of this message?: 'I am so happy!'")
+output = agent.run("What is the sentiment of this message?: 'I am so happy!'").output
+
+print(json.dumps(json.loads(output.value), indent=2))
--- a/docs/griptape-framework/structures/src/json_schema_rule.py
+++ b/docs/griptape-framework/structures/src/json_schema_rule.py
@@ -1,0 +1,14 @@
+import schema
+
+from griptape.rules.json_schema_rule import JsonSchemaRule
+from griptape.structures import Agent
+
+agent = Agent(
+    rules=[
+        JsonSchemaRule(
+            schema.Schema({"answer": str, "relevant_emojis": schema.Schema(["str"])}).json_schema("Output Format")
+        )
+    ]
+)
+
+agent.run("What is the sentiment of this message?: 'I am so happy!'")
--- a/docs/griptape-framework/structures/src/json_schema_rule_pydantic.py
+++ b/docs/griptape-framework/structures/src/json_schema_rule_pydantic.py
@@ -1,0 +1,22 @@
+from __future__ import annotations
+
+import pydantic
+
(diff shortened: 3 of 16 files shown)


collindutter:
Question: should this be the mechanism by which we enable features like OpenAI Structured Outputs? If so, we'll need to make the prompt rendering happen at the Driver level rather than the Task level.

docs/griptape-framework/structures/src/json_schema_rule.py
```diff
@@ -0,0 +1,14 @@
+import schema
+
+from griptape.rules.json_schema_rule import JsonSchemaRule
+from griptape.structures import Agent
+
+agent = Agent(
+    rules=[
+        JsonSchemaRule(
+            schema.Schema({"answer": str, "relevant_emojis": schema.Schema(["str"])}).json_schema("Output Format")
```
dylanholmes, line 9: Would be cool to give an example using pydantic model to create the schema, then converting the output text back into a python object. I can think of some people in particular who might appreciate such an example. I suppose it doesn't need to be pydantic, but showing how to get a python object back seems like a good way to highlight the power of this new rule.

(Like tada, now you can use autocomplete in your editor again!)

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Parameter `meta: dict` on `BaseEvent`.
 - `AzureOpenAiTextToSpeechDriver`.
+- `JsonSchemaRule` for instructing the LLM to output a JSON object that conforms to a schema.
```
dylanholmes, line 16: Responding with an inline comment, so it's easier to reply.

> Question: should this be the mechanism by which we enable features like OpenAI Structured Outputs? If so, we'll need to make the prompt rendering happen at the Driver level rather than the Task level.

If we didn't use this mechanism for OpenAI Structured Outputs, then what would the mechanism for that look like? Would it require passing a json schema? (I would expect so) If it requires passing a json schema, then we'd have two choices, which doesn't seem like the right approach (imagine that we had taken that approach for tools 🤮 ).

Seems like making prompt rendering at the Driver level is the way forward because I don't really see duplicate ways of passing in a json schema as a viable option. This could technically wait until introducing native structured outputs, right?

(Also, if we do use this rule for native structure outputs, what mechanism will we use to opt-in/out of structure outputs when a driver supports it? Should we that the same approach as a `use_native_tools` flag?)

dylanholmes (approved):
Nice!

Approving because I think there is some immediate value to this. Also, I like the choice of using a rule to enforce json schema, since that's the same as what you'd do now. It's intuitive imo.

Even if we need to change some things to make it work with say OpenAI Structure Outputs, I like this as a "Rule" from a API perspective.

CHANGELOG.md
collindutter: > Would it require passing a json schema?
Yes, it'd likely require passing a schema to the Prompt Driver 🤢. 

Rendering at the Prompt Driver level allows us to lean on more provider-specific nuances so I think it's a worthwhile change. But as you point out it's not required for this PR -- this can still serve as a good intermediate solution.

docs/griptape-framework/structures/src/json_schema_rule_pydantic.py
```diff
@@ -0,0 +1,18 @@
...
+class OutputFormat(pydantic.BaseModel):
+    answer: str
+    relevant_emojis: list[str]
+
+
+agent = Agent(rules=[JsonSchemaRule(OutputFormat.model_json_schema())])
+
+output = agent.run("What is the sentiment of this message?: 'I am so happy!'").output
+
+print(OutputFormat.model_validate_json(output.value))
```
dylanholmes, line 18: How about highlighting some of the sugar!

```py
from pydantic import BaseModel, Field

class SentimentAnalysis(BaseModel):
    sentiment: str = Field(description="Sentiment of the message, one of 'positive', 'neutral', 'negative'")
    relevant_emojis: list[str]

...

sentiment_analysis_text = ...

sentiment_analysis = SentimentAnalysis.model_validate_json(...)

# Autocomplete via dot notation 🤩 
print("sentiment: ", sentiment_analysis.sentiment)
print("relevant_emojis: ", sentiment_analysis.relevant_emojis)
```

dylanholmes (dismissed):
Yay!

Optional suggestions on the sample

docs/griptape-framework/structures/src/json_schema_rule_pydantic.py
collindutter: Sugar added!

dylanholmes (dismissed):

dylanholmes (dismissed):
x2

dylanholmes (approved):

PR #1184, opened 2024-09-18
Chore/main

collindutter:


.github/actions/init-bare-environment/action.yml
```diff
@@ -6,7 +6,8 @@ runs:
       - name: Checkout actions
         uses: actions/checkout@v3
 
-      - name: Set up Python ${{ matrix.python-version }}
+      - id: setup-python
+        name: Set up Python ${{ matrix.python-version }}
```
collindutter, line 10: Tbh not sure why this is showing in the diff since it should already be on dev.

vachillo (approved):

PR #1191, opened 2024-09-23
Feature/workflow improvements

collindutter:

### Added
- `Workflow.input_tasks` and `Workflow.output_tasks` to access the input and output tasks of a Workflow.
- Ability to pass nested list of `Tasks` to `Structure.tasks` allowing for more complex declarative Structure definitions.
NA

### DIFF
--- a/griptape/structures/agent.py
+++ b/griptape/structures/agent.py
@@ -60,15 +60,15 @@ class Agent(Structure):
         return self.tasks[0]
 
     def add_task(self, task: BaseTask) -> BaseTask:
-        self.tasks.clear()
+        self._tasks.clear()
 
         task.preprocess(self)
 
-        self.tasks.append(task)
+        self._tasks.append(task)
 
         return task
 
-    def add_tasks(self, *tasks: BaseTask) -> list[BaseTask]:
+    def add_tasks(self, *tasks: BaseTask | list[BaseTask]) -> list[BaseTask]:
         if len(tasks) > 1:
             raise ValueError("Agents can only have one task.")
         return super().add_tasks(*tasks)
--- a/griptape/structures/structure.py
+++ b/griptape/structures/structure.py
@@ -72,10 +72,6 @@ class Structure(ABC):
                 tasks.append(task)
         return tasks
 
-    @tasks.setter
-    def tasks(self, value: list[BaseTask | list[BaseTask]]) -> None:
-        self._tasks = value
-
     @property
     def execution_args(self) -> tuple:
         return self._execution_args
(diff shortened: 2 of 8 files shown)


griptape/structures/agent.py
```diff
@@ -60,15 +60,15 @@ def task(self) -> BaseTask:
...
         task.preprocess(self)
 
-        self.tasks.append(task)
+        self._tasks.append(task)
 
         return task
 
-    def add_tasks(self, *tasks: BaseTask) -> list[BaseTask]:
+    def add_tasks(self, *tasks: BaseTask | list[BaseTask]) -> list[BaseTask]:
         if len(tasks) > 1:
```
vachillo, line 72: need to update this check?

griptape/structures/structure.py
```diff
@@ -24,7 +24,7 @@ class Structure(ABC):
     id: str = field(default=Factory(lambda: uuid.uuid4().hex), kw_only=True)
     rulesets: list[Ruleset] = field(factory=list, kw_only=True)
     rules: list[BaseRule] = field(factory=list, kw_only=True)
-    tasks: list[BaseTask] = field(factory=list, kw_only=True)
+    _tasks: list[BaseTask | list[BaseTask]] = field(factory=list, kw_only=True, alias="tasks")
```
vachillo, line 27: based on the property below, this should still just be `list[BaseTask]` no?

griptape/structures/workflow.py
```diff
@@ -26,13 +26,21 @@ def input_task(self) -> Optional[BaseTask]:
     def output_task(self) -> Optional[BaseTask]:
         return self.order_tasks()[-1] if self.tasks else None
 
+    @property
+    def input_tasks(self) -> list[BaseTask]:
+        return [task for task in self.tasks if not task.parents]
+
+    @property
+    def output_tasks(self) -> list[BaseTask]:
+        return [task for task in self.tasks if not task.children]
```
vachillo, line 35: do we need to call `self.order_tasks()` like above?

griptape/structures/structure.py
collindutter: This is the input to the Structure, it needs to be `llist[BaseTask | list[BaseTask]]` to allow for:
```python
workflow = Workflow(
    tasks=[
        [
            PromptTask("parent", id=f"parent_{i}"),
            PromptTask("child", id=f"child_{i}", parent_ids=[f"parent_{i}"]),
        ]
        for i in range(3)
    ],
)
```

griptape/structures/workflow.py
collindutter: I don't think so? Because we're not making any assumptions about `self.tasks`'s ordering. We're just grabbing the ones that don't have children.

dylanholmes (approved):

Issue #1178, opened 2024-09-16
Add LumaAI as a driver


shhlife:
- [x ] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

LumaLabs now has an api available for Dream Machine. Can we add it as a driver to griptape?

https://lumalabs.ai/dream-machine/api

PR #1199, opened 2024-09-23
Feature/Luma Ai Video Generation Driver

william-price01:
- [ ] I have read and agree to the contributing guidelines for [submitting new pull requests](https://github.com/griptape-ai/griptape?tab=readme-ov-file#submitting-pull-requests).

## Describe your changes
Added Luma AI as a driver, tool, task, to generate videos.
[Add Luma Ai as a driver](https://github.com/griptape-ai/griptape/issues/1178)

### DIFF
--- a/griptape/artifacts/video_artifact.py
+++ b/griptape/artifacts/video_artifact.py
@@ -1,0 +1,29 @@
+from __future__ import annotations
+
+from attrs import define, field
+
+from griptape.artifacts import BlobArtifact
+
+
+@define
+class VideoArtifact(BlobArtifact):
+    """Stores video binary data and relevant metadata.
+
+    Attributes:
+        value: The video binary data.
+        mime_type: The video MIME type.
+        resolution: The resolution of the video (e.g., 1920x1080).
+        duration: Duration of the video in seconds.
+    """
+
+    aspect_ratio: tuple[int, int] = field(default=(16, 9), kw_only=True)
+
+    @property
+    def mime_type(self) -> str:
+        return "video/mp4"  # Or make this flexible based on the video format
+
+    def get_aspect_ratio(self) -> tuple[int, int]:
+        return self.aspect_ratio
+
+    def to_text(self) -> str:
+        raise NotImplementedError("VideoArtifact cannot be converted to text.")
--- a/griptape/drivers/__init__.py
+++ b/griptape/drivers/__init__.py
@@ -134,6 +134,9 @@ from .observability.open_telemetry_observability_driver import OpenTelemetryObse
 from .observability.griptape_cloud_observability_driver import GriptapeCloudObservabilityDriver
 from .observability.datadog_observability_driver import DatadogObservabilityDriver
 
+from .video_generation.base_video_generation_driver import BaseVideoGenerationDriver
+from .video_generation.dream_machine_video_generation_driver import DreamMachineVideoGenerationDriver
+
 __all__ = [
     "BasePromptDriver",
     "OpenAiChatPromptDriver",
@@ -240,4 +243,6 @@ __all__ = [
     "OpenTelemetryObservabilityDriver",
     "GriptapeCloudObservabilityDriver",
     "DatadogObservabilityDriver",
+    "DreamMachineVideoGenerationDriver",
+    "BaseVideoGenerationDriver",
 ]
--- a/griptape/drivers/video_generation/base_video_generation_driver.py
+++ b/griptape/drivers/video_generation/base_video_generation_driver.py
@@ -1,0 +1,36 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+from attrs import define
+
+from griptape.events import EventBus, FinishVideoGenerationEvent, StartVideoGenerationEvent
+from griptape.mixins.exponential_backoff_mixin import ExponentialBackoffMixin
+from griptape.mixins.serializable_mixin import SerializableMixin
+
+if TYPE_CHECKING:
+    from griptape.artifacts import VideoArtifact
+
+
+@define
+class BaseVideoGenerationDriver(SerializableMixin, ExponentialBackoffMixin, ABC):
+    def before_run(self, prompt: str) -> None:
+        EventBus.publish_event(StartVideoGenerationEvent(prompt=prompt))
+
+    def after_run(self) -> None:
+        EventBus.publish_event(FinishVideoGenerationEvent())
+
+    def run_text_to_video(self, prompt: str) -> VideoArtifact:
+        for attempt in self.retrying():
+            with attempt:
+                self.before_run(prompt)
+                result = self.try_text_to_video(prompt)
+                self.after_run()
+
+                return result
+        else:
+            raise Exception("Failed to run text to video generation")
+
+    @abstractmethod
+    def try_text_to_video(self, prompt: str) -> VideoArtifact: ...
--- a/griptape/drivers/video_generation/dream_machine_video_generation_driver.py
+++ b/griptape/drivers/video_generation/dream_machine_video_generation_driver.py
@@ -13,6 +13,7 @@ from griptape.utils import import_optional_dependency
 
 if TYPE_CHECKING:
     from lumaai import LumaAI
+
 logger = logging.getLogger(__name__)
 
 
--- a/griptape/drivers/video_generation/dream_machine_video_generation_driver.py
+++ b/griptape/drivers/video_generation/dream_machine_video_generation_driver.py
@@ -1,0 +1,55 @@
+from __future__ import annotations
+
+import logging
+import time
+from typing import TYPE_CHECKING, Any
+
+import requests
+from attrs import Factory, define, field
+
+from griptape.artifacts import VideoArtifact
+from griptape.drivers import BaseVideoGenerationDriver
+from griptape.utils import import_optional_dependency
+
+if TYPE_CHECKING:
+    from lumaai import LumaAI
+logger = logging.getLogger(__name__)
+
+
+@define
+class DreamMachineVideoGenerationDriver(BaseVideoGenerationDriver):
+    api_key: str = field(kw_only=True, metadata={"serializable": True})
+    client: LumaAI = field(
+        default=Factory(
+            lambda self: import_optional_dependency("lumaai").LumaAI(auth_token=self.api_key), takes_self=True
+        ),
+        kw_only=True,
+    )
+    params: dict[str, Any] = field(default={}, kw_only=True, metadata={"serializable": True})
+
+    def try_text_to_video(self, prompt: str) -> VideoArtifact:
+        response = self.client.generations.create(prompt=prompt, **self.params)
+        generation = response
+        status = generation.state
+        while status in ["dreaming", "queued"]:
+            time.sleep(5)
+            if not generation.id:
+                raise Exception("Generation ID not found in the response")
+
+            generation = self.client.generations.get(generation.id)
+            status = generation.state
+        if status == "completed":
+            video_url = generation.assets.video
+            if not video_url:
+                raise Exception("Video URL not found in the generation response")
+            video_binary = self._download_video(video_url)
+            return VideoArtifact(
+                value=video_binary,
+            )
+        else:
+            raise Exception(f"Video generation failed with status: {status}")
+
+    def _download_video(self, video_url: str) -> bytes:
+        response = requests.get(video_url)
+        response.raise_for_status()
+        return response.content
--- a/griptape/mixins/artifact_file_output_mixin.py
+++ b/griptape/mixins/artifact_file_output_mixin.py
@@ -7,7 +7,7 @@ from typing import TYPE_CHECKING, Optional
 from attrs import Attribute, define, field
 
 if TYPE_CHECKING:
-    from griptape.artifacts import BaseArtifact
+    from griptape.artifacts import BaseArtifact, VideoArtifact
 
 
 @define(slots=False)
@@ -36,6 +36,19 @@ class ArtifactFileOutputMixin:
             outfile = self.output_file
         elif self.output_dir:
             outfile = os.path.join(self.output_dir, artifact.name)
+        else:
+            raise ValueError("No output_file or output_dir specified.")
+
+        if os.path.dirname(outfile):
+            os.makedirs(os.path.dirname(outfile), exist_ok=True)
+
+        Path(outfile).write_bytes(artifact.to_bytes())
+
+    def save_video_artifact(self, artifact: VideoArtifact) -> None:
+        if self.output_file:
+            outfile = self.output_file
+        elif self.output_dir:
+            outfile = os.path.join(self.output_dir, artifact.name + ".mp4")
         else:
             raise ValueError("No output_file or output_dir specified.")
 
--- a/griptape/tasks/prompt_video_generation_task.py
+++ b/griptape/tasks/prompt_video_generation_task.py
@@ -1,0 +1,59 @@
+from __future__ import annotations
+
+from typing import Callable
+
+from attrs import Factory, define, field
+
+from griptape.artifacts import TextArtifact, VideoArtifact
+from griptape.drivers import DreamMachineVideoGenerationDriver
+from griptape.tasks import BaseTask, BaseVideoGenerationTask
+from griptape.utils import J2
+
+
+@define
+class PromptVideoGenerationTask(BaseVideoGenerationTask):
+    """Used to generate a video from a text prompt.
+
+    Accepts prompt as input in one of the following formats:
+    - template string
+    - TextArtifact
+    - Callable that returns a TextArtifact.
+
+    Attributes:
+        video_generation_driver: The driver used to generate the video.
+        negative_prompts: List of negatively-weighted prompts applied to the text prompt, if supported by the driver.
+        output_dir: If provided, the generated video will be written to disk in output_dir.
+        output_file: If provided, the generated video will be written to disk as output_file.
+    """
+
+    api_key: str = field(kw_only=True)
+    DEFAULT_INPUT_TEMPLATE = "{{ args[0] }}"
+
+    _input: str | TextArtifact | Callable[[BaseTask], TextArtifact] = field(
+        default=DEFAULT_INPUT_TEMPLATE, alias="input"
+    )
+    video_generation_driver: DreamMachineVideoGenerationDriver = field(
+        default=Factory(lambda self: DreamMachineVideoGenerationDriver(api_key=self.api_key), takes_self=True),
+        kw_only=True,
+    )
+
+    @property
+    def input(self) -> TextArtifact:
+        if isinstance(self._input, TextArtifact):
+            return self._input
+        elif isinstance(self._input, Callable):
+            return self._input(self)
+        else:
+            return TextArtifact(J2().render_from_string(self._input, **self.full_context))
+
+    @input.setter
+    def input(self, value: TextArtifact) -> None:
+        self._input = value
+
+    def run(self) -> VideoArtifact:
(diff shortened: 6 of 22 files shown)


griptape/artifacts/video_artifact.py
```diff
@@ -0,0 +1,29 @@
...
+
+@define
+class VideoArtifact(BlobArtifact):
+    """Stores video binary data and relevant metadata.
+
+    Attributes:
+        value: The video binary data.
+        mime_type: The video MIME type.
+        resolution: The resolution of the video (e.g., 1920x1080).
+        duration: Duration of the video in seconds.
```
SavagePencil, line 16: which of these do you feel confident you can have in the initial version (for example, it sounded like a target duration may not be something easy to get)? Are there other params that the providers expose that we should be taking into consideration? For example, I would anticipate framerate to be a popular parameter so that we can have everyone's favorite NTSC 59.94!

griptape/artifacts/video_artifact.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    aspect_ratio: tuple[int, int] = field(default=(16, 9), kw_only=True)
+
+    @property
+    def mime_type(self) -> str:
+        return "video/mp4"  # Or make this flexible based on the video format
```
SavagePencil, line 23: sounds like our loaders support .ogg and .webm, can we point to the same loc for both or is there a reason to keep them separate?

griptape/artifacts/video_artifact.py
```diff
@@ -0,0 +1,29 @@
...
+    """Stores video binary data and relevant metadata.
+
+    Attributes:
+        value: The video binary data.
+        mime_type: The video MIME type.
+        resolution: The resolution of the video (e.g., 1920x1080).
+        duration: Duration of the video in seconds.
+    """
+
+    aspect_ratio: tuple[int, int] = field(default=(16, 9), kw_only=True)
```
SavagePencil, line 19: open question: do providers enumerate possible aspect ratios, or let you do weirdo things

griptape/artifacts/video_artifact.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @property
+    def mime_type(self) -> str:
+        return "video/mp4"  # Or make this flexible based on the video format
+
+    def get_aspect_ratio(self) -> tuple[int, int]:
+        return self.aspect_ratio
```
vachillo, line 26: dont need this, `aspect_ratio` can be directly accessed

griptape/artifacts/video_artifact.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def to_text(self) -> str:
+        raise NotImplementedError("VideoArtifact cannot be converted to text.")
```
vachillo, line 29: use an approach similar to AudioArtifact for this

SavagePencil, line 29: other not-so-texty artifacts generate a string that describes the parameters of the artifact, is that what we want here, too?

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
```diff
@@ -0,0 +1,56 @@
...
+
+@define
+class DreamMachineVideoGenerationDriver(BaseVideoGenerationDriver):
+    api_key: str = field(kw_only=True, metadata={"serializable": True})
+    client: LumaAI = field(
+        default=Factory(
+            lambda self: import_optional_dependency("lumaai").LumaAI(auth_token=self.api_key), takes_self=True
+        ),
+        kw_only=True,
+    )
```
vachillo, line 28: use the `lazy_property` approach for this

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
```diff
@@ -0,0 +1,56 @@
...
+    params: dict[str, Any] = field(default={}, kw_only=True, metadata={"serializable": True})
```
vachillo, line 29: use `factory=dict` here, this will return a single mutable object which is no good

griptape/drivers/__init__.py
```diff
@@ -242,4 +245,6 @@
     "OpenTelemetryObservabilityDriver",
     "GriptapeCloudObservabilityDriver",
     "DatadogObservabilityDriver",
+    "DreamMachineVideoGenerationDriver",
+    "BaseVideoGenerationDriver",
```
SavagePencil, line 249: am I the only one at this damned company that likes my lists alphabetized? Not for this PR, but dang

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
```diff
@@ -0,0 +1,56 @@
...
+        response = self.client.generations.create(prompt=prompt, **self.params)
+        generation = response
+        status = generation.state
+        while status in ["dreaming", "queued"]:
+            time.sleep(5)
+            if not generation.id:
+                raise Exception("Generation ID not found in the response")
+
+            generation = self.client.generations.get(generation.id)
+            status = generation.state
```
vachillo, line 41: try to use the `tenacity` library for this here. i know this is the same approach as the cloud driver but we should try to clean this pattern up

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
```diff
@@ -0,0 +1,56 @@
...
+        if status == "completed":
+            video_url = generation.assets.video
+            if not video_url:
+                raise Exception("Video URL not found in the generation response")
```
vachillo, line 45: when would this happen?

griptape/drivers/video_generation/base_video_generation_driver.py
```diff
@@ -0,0 +1,36 @@
...
+    def run_text_to_video(self, prompt: str) -> VideoArtifact:
+        for attempt in self.retrying():
+            with attempt:
+                self.before_run(prompt)
+                result = self.try_text_to_video(prompt)
+                self.after_run()
+
+                return result
+        else:
+            raise Exception("Failed to run text to video generation")
```
SavagePencil, line 33: is this the pattern we use elsewhere? generic `Exception` with no details?

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
```diff
@@ -0,0 +1,56 @@
...
+        default=Factory(
+            lambda self: import_optional_dependency("lumaai").LumaAI(auth_token=self.api_key), takes_self=True
+        ),
+        kw_only=True,
+    )
+    params: dict[str, Any] = field(default={}, kw_only=True, metadata={"serializable": True})
+
+    def try_text_to_video(self, prompt: str) -> VideoArtifact:
+        response = self.client.generations.create(prompt=prompt, **self.params)
+        generation = response
```
SavagePencil, line 33: instead of assigning to an alias, why not assign directly to `generation`?

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
```diff
@@ -0,0 +1,56 @@
...
+        status = generation.state
+        while status in ["dreaming", "queued"]:
+            time.sleep(5)
+            if not generation.id:
+                raise Exception("Generation ID not found in the response")
```
SavagePencil, line 38: Why is this in the while loop? Is this something that should be caught either before or after the dreaming begins?

griptape/mixins/artifact_file_output_mixin.py
```diff
@@ -43,3 +43,16 @@ def _write_to_file(self, artifact: BaseArtifact) -> None:
...
 
         Path(outfile).write_bytes(artifact.to_bytes())
+
+    def save_video_artifact(self, artifact: VideoArtifact) -> None:
+        if self.output_file:
+            outfile = self.output_file
+        elif self.output_dir:
+            outfile = os.path.join(self.output_dir, artifact.name + ".mp4")
+        else:
+            raise ValueError("No output_file or output_dir specified.")
```
vachillo, line 53: why this method? `_write_to_file` should be fine

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
SavagePencil, line 45: If no URL shows up, is that indicative of something bad, like it "completed" but was somehow a failure? Can we convey what the situation is to the user?

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
```diff
@@ -0,0 +1,56 @@
...
+
+            generation = self.client.generations.get(generation.id)
+            status = generation.state
+        if status == "completed":
+            video_url = generation.assets.video
+            if not video_url:
+                raise Exception("Video URL not found in the generation response")
+            video_binary = self._download_video(video_url)
```
SavagePencil, line 46: downloading feels like it could wrong in a lot of other ways (retries, timeouts, etc.)

griptape/tasks/prompt_video_generation_task.py
```diff
@@ -0,0 +1,59 @@
...
+from attrs import Factory, define, field
+
+from griptape.artifacts import TextArtifact, VideoArtifact
+from griptape.drivers import DreamMachineVideoGenerationDriver
+from griptape.tasks import BaseTask, BaseVideoGenerationTask
+from griptape.utils import J2
+
+
+@define
+class PromptVideoGenerationTask(BaseVideoGenerationTask):
```
vachillo, line 14: i dont think these need to be two different classes. the `input` can be different types instead

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
```diff
@@ -0,0 +1,56 @@
...
+            return VideoArtifact(
+                value=video_binary,
+            )
+        else:
+            raise Exception(f"Video generation failed with status: {status}")
+
+    def _download_video(self, video_url: str) -> bytes:
```
SavagePencil, line 53: videos are big and take a long time to come down. Do we...
1. ...already have this functionality somewhere else in the framework?
2. ...need to take into consideration fails on memory, disk, retries, timeout?

griptape/tools/prompt_video_generation/tool.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+
+    driver: DreamMachineVideoGenerationDriver = field(kw_only=True)
+
+    @activity(
+        config={
+            "description": "Generates a video from text prompts.",
+            "schema": Schema(
+                {
+                    Literal("prompt", description=BaseVideoGenerationTool.PROMPT_DESCRIPTION): str,
```
vachillo, line 33: this can be inlined

griptape/tools/prompt_video_generation/tool.py
```diff
@@ -0,0 +1,46 @@
...
+from griptape.tools import BaseVideoGenerationTool
+from griptape.utils.decorators import activity
+
+if TYPE_CHECKING:
+    from griptape.artifacts import ErrorArtifact, VideoArtifact
+    from griptape.drivers import DreamMachineVideoGenerationDriver
+
+
+@define
+class PromptVideoGenerationTool(BaseVideoGenerationTool):
```
vachillo, line 17: this doesnt need to be two classes, just inherit from `BaseTool` and make `VideoGenerationTool`

griptape/tools/prompt_video_generation/tool.py
```diff
@@ -0,0 +1,46 @@
...
+        config={
+            "description": "Generates a video from text prompts.",
+            "schema": Schema(
+                {
+                    Literal("prompt", description=BaseVideoGenerationTool.PROMPT_DESCRIPTION): str,
+                }
+            ),
+        },
+    )
+    def generate_video(self, params: dict[str, dict[str, str]]) -> VideoArtifact | ErrorArtifact:
```
vachillo, line 38: this is just typed as `params: dict`

griptape/artifacts/video_artifact.py
william-price01: This was more of a testing method, while getting it to actually work.

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
william-price01: Again, for testing purposes, pyright was giving me a hard time for not checking if the id exists.

griptape/mixins/artifact_file_output_mixin.py
william-price01: For some reason, it would call the to_text method inside of the base artifact class which the Video Artifact is unable to be converted to text in the same way.

vachillo: you need to override `to_bytes` on the vide artifact

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
vachillo: id be fine with putting this functionality directly on `VideoArtifact` as a "lazy load" type feature. the alternative is creating a persistence driver for everything, which we dont have, but we have the `ArtifactFileOutputMixin` available.

griptape/artifacts/video_artifact.py
william-price01: I'm going to somehow include most of the things  [Jason](https://griptape-ai.slack.com/archives/C05T3HLBH52/p1727295829173599?thread_ts=1727292898.313379&cid=C05T3HLBH52) mentioned, I don't believe that there is a way to bring audio with these videos, so probably no subtitles for initial version.

griptape/artifacts/video_artifact.py
dylanholmes: We should already have the same mime-type umbrella problem for image/audio, if its not in the framework already, check with collin on the preferred approach.

Also, should this be an actual attribute instead of a `@property`?

griptape/drivers/__init__.py
dylanholmes: I agree with james, but don't reorder this list in this PR. (Generally mixing refactoring or formatting with features makes it difficult for reviewers to see what is actually changing)

dylanholmes (commented):
Didn't quite get a full review in yet, but overall the approach seems fine. You are following existing patterns (e.g. image generation) which I like.

I'll try to review a little more later today

griptape/mixins/artifact_file_output_mixin.py
william-price01: okay, that makes sense! Thank you!

griptape/artifacts/video_artifact.py
collindutter, line 23: Add a `format` field similar to `AudioArtifact` and use that here.

collindutter: `@property` is correct, just needs to be updated to something like [this](https://github.com/griptape-ai/griptape/blob/53bc38b5162ef927127a7f838d2c74c88cc0919d/griptape/artifacts/audio_artifact.py#L18-L20).

griptape/drivers/__init__.py
collindutter: `BaseVideoGenerationDriver` should be placed before `DreamMachineVideoGenerationDriver` for circular dependency (and alphabetical) reasons.

griptape/drivers/video_generation/base_video_generation_driver.py
collindutter: This also caught my eye, but we do it in `BasePromptDriver` which is probably where this came from. `RuntimeError` would probably be a better candidate here.

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
```diff
@@ -0,0 +1,56 @@
...
+        kw_only=True,
+    )
+    params: dict[str, Any] = field(default={}, kw_only=True, metadata={"serializable": True})
+
+    def try_text_to_video(self, prompt: str) -> VideoArtifact:
+        response = self.client.generations.create(prompt=prompt, **self.params)
+        generation = response
+        status = generation.state
+        while status in ["dreaming", "queued"]:
+            time.sleep(5)
```
collindutter, line 36: Sleep time should be configurable as something like `poll_interval`

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
collindutter, line 38: Flip this logic to be something like:
```python
if generation.id:
...
else:
raise ValueError

File: griptape/drivers/video_generation/dream_machine_video_generation_driver.py (hunk 9 above)
Comment (collindutter) [inline griptape/drivers/video_generation/dream_machine_video_generation_driver.py:41]: Can get rid of `status` -- just use `generation.state`

Comment (collindutter) [reply griptape/drivers/video_generation/dream_machine_video_generation_driver.py:41]: I'm actually not sure about this. Retrying (tenacity) feels different than polling (what we're doing here).

File: griptape/drivers/video_generation/dream_machine_video_generation_driver.py [hunk 22]
```diff
@@ -0,0 +1,56 @@
...
+            if not generation.id:
+                raise Exception("Generation ID not found in the response")
+
+            generation = self.client.generations.get(generation.id)
+            status = generation.state
+        if status == "completed":
```
collindutter, line 42: Does the exa library provide any types for this so we can do something like `status == exa.COMPLETED`

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
collindutter, line 45: Flip conditional as mentioned above. Also raise a `ValueError`

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
collindutter: Agreed that the Driver should probably not be the one to do this. I don't know about putting it in `VideoArtifact` either -- feels like too much responsibility.

In-fact, we probably shouldn't make any assumptions that the user wants us to download it. Maybe they're fine receiving a URL that they watch in their browser. Maybe this should fall onto a Loader?

collindutter (changes requested):
Great start! Have not looked at the entire thing but left some initial feedback.

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
vachillo: tenacity has built in conditions for `if_condition` or `if_not_condition`. its not just for retrying on errors

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
vachillo: could be the same idea as a lazy load. `VideoArtifact` contains a reference to the video somewhere.

griptape/drivers/video_generation/dream_machine_video_generation_driver.py
collindutter: I don't doubt it can be used for non-exception things, but all of the examples are for exception retrying which makes me think that's its primary purpose.

Do you have an example of how tenacity might be used for polling?

vachillo: something like
```
from tenacity import retry, retry_if_result, stop_after_attempt, wait_fixed

@retry(
    retry=retry_if_result(lambda result: result is None),
    stop=stop_after_attempt(3),
    wait=wait_fixed(5),
)
def call_api(url_to_poll: str) -> Optional[str]:
    response = requests.get(url_to_poll)
    if response.status_code != 200:
        return None
    return response.text
```

collindutter:
This functionality is experimental enough where it should be initially implemented as a [griptape extension](https://github.com/griptape-ai/tool-template) once https://github.com/griptape-ai/tool-template/pull/3 goes in

PR #1204, opened 2024-09-26
Change Stucture.output behavior/return type

collindutter:

### Changed
- `Structure.output`'s type is now `BaseArtifact` and raises an exception if the output is `None`.
Cuts out `None` checks post Structure run
NA

### DIFF
--- a/griptape/structures/pipeline.py
+++ b/griptape/structures/pipeline.py
@@ -6,7 +6,6 @@ from attrs import define
 
 from griptape.artifacts import ErrorArtifact
 from griptape.common import observable
-from griptape.memory.structure import Run
 from griptape.structures import Structure
 
 if TYPE_CHECKING:
@@ -52,11 +51,6 @@ class Pipeline(Structure):
     @observable
     def try_run(self, *args) -> Pipeline:
         self.__run_from_task(self.input_task)
-
-        if self.conversation_memory and self.output_task.output is not None:
-            run = Run(input=self.input_task.input, output=self.output_task.output)
-
-            self.conversation_memory.add_run(run)
 
         return self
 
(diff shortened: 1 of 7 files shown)


griptape/structures/pipeline.py
```diff
@@ -53,8 +53,8 @@ def insert_task(self, parent_task: BaseTask, task: BaseTask) -> BaseTask:
     def try_run(self, *args) -> Pipeline:
         self.__run_from_task(self.input_task)
 
-        if self.conversation_memory and self.output is not None:
-            run = Run(input=self.input_task.input, output=self.output)
+        if self.conversation_memory and self.output_task.output is not None:
```
cjkindel, line 56: Is there any desire to consolidate the conversation memory logic of all the structures (L56-59 here, L117-120 in workflow)? Noticed the same code when tinkering with a custom Agent locally and now this PR is surfacing it again.

collindutter: Sure thing, it would consolidate nicely into `Structure.after_run`. Would you mind making that PR?

collindutter: Actually since there's no change in behavior I can just cram it into this one.

cjkindel (approved):

griptape/structures/structure.py
```diff
@@ -163,6 +165,11 @@ def before_run(self, args: Any) -> None:
 
     @observable
     def after_run(self) -> None:
+        if self.conversation_memory and self.output_task.output is not None:
```
vachillo, line 168: do you need the second `is not None` check now?

collindutter: There were some tests that relied on `output_task.output` being `None` which is still a valid state. Didn't feel like refactoring those.

PR #1205, opened 2024-09-26
Add Ruleset Drivers

vachillo:

## Describe your changes
adding `RulesetDriver`s to pull in read-only rules from external sources, namely Griptape Cloud

### DIFF
--- a/docs/griptape-framework/drivers/ruleset-drivers.md
+++ b/docs/griptape-framework/drivers/ruleset-drivers.md
@@ -1,0 +1,26 @@
+---
+search:
+  boost: 2
+---
+
+## Overview
+
+Ruleset Drivers can be used to load rules in from external sources. 
+
+## Ruleset Drivers
+
+### Local
+
+The [LocalRulesetDriver](../../reference/griptape/drivers/ruleset/local_ruleset_driver.md) allows you to load a Ruleset from a local JSON file. The `persist_dir` parameter is used to specify a local directory where one or more Ruleset files are located. If no `persist_dir` parameter is given, the `.load` method is a no-op.
+
+```python
+--8<-- "docs/griptape-framework/drivers/src/local_ruleset_driver.py"
+```
+
+### Griptape Cloud
+
+The [GriptapeCloudRulesetDriver](../../reference/griptape/drivers/ruleset/griptape_cloud_ruleset_driver.md) allows you to load a Griptape Cloud Ruleset resource. `Ruleset.name` is used to try and find a Griptape Cloud Ruleset with that alias.
+
+```python
+--8<-- "docs/griptape-framework/drivers/src/griptape_cloud_ruleset_driver.py"
+```
--- a/docs/griptape-framework/drivers/src/griptape_cloud_ruleset_driver.py
+++ b/docs/griptape-framework/drivers/src/griptape_cloud_ruleset_driver.py
@@ -1,0 +1,7 @@
+from griptape.drivers import GriptapeCloudRulesetDriver
+from griptape.rules import Ruleset
+
+rulset = Ruleset(
+    name="my_griptape_cloud_ruleset_alias",
+    ruleset_driver=GriptapeCloudRulesetDriver(),
+)
--- a/docs/griptape-framework/drivers/src/local_ruleset_driver.py
+++ b/docs/griptape-framework/drivers/src/local_ruleset_driver.py
@@ -1,0 +1,19 @@
+import json
+import os
+from pathlib import Path
+
+from griptape.drivers import LocalRulesetDriver
+from griptape.rules import Ruleset
+
+ruleset_dir = "path/to/ruleset/dir"
+ruleset_name = "my_local_ruleset.json"
+ruleset_path = Path(os.path.join(ruleset_dir, ruleset_name))
+
+os.makedirs(ruleset_dir, exist_ok=True)
+
+ruleset_path.write_text(json.dumps({"rules": [{"value": "Always talk like a pirate."}]}))
+
+ruleset = Ruleset(
+    name=ruleset_name,
+    ruleset_driver=LocalRulesetDriver(persist_dir=ruleset_dir),
+)
--- a/griptape/drivers/ruleset/base_ruleset_driver.py
+++ b/griptape/drivers/ruleset/base_ruleset_driver.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Any
+
+from attrs import define, field
+
+from griptape.mixins.serializable_mixin import SerializableMixin
+
+if TYPE_CHECKING:
+    from griptape.rules import BaseRule
+
+
+@define
+class BaseRulesetDriver(SerializableMixin, ABC):
+    """Base class for ruleset drivers.
+
+    Attributes:
+        raise_not_found: Whether to raise an error if the ruleset is not found. Defaults to True.
+    """
+
+    raise_not_found: bool = field(default=True, kw_only=True, metadata={"serializable": True})
+
+    @abstractmethod
+    def load(self, ruleset_name: str) -> tuple[list[BaseRule], dict[str, Any]]: ...
+
+    def _from_ruleset_dict(self, params_dict: dict[str, Any]) -> tuple[list[BaseRule], dict[str, Any]]:
+        return [
+            self._get_rule(rule["value"], rule.get("meta", {})) for rule in params_dict.get("rules", [])
+        ], params_dict.get("meta", {})
+
+    def _get_rule(self, value: Any, meta: dict[str, Any]) -> BaseRule:
+        from griptape.rules import JsonSchemaRule, Rule
+
+        return JsonSchemaRule(value=value, meta=meta) if isinstance(value, dict) else Rule(value=str(value), meta=meta)
--- a/griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
+++ b/griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
@@ -1,0 +1,89 @@
+from __future__ import annotations
+
+import os
+from typing import TYPE_CHECKING, Any, Optional
+from urllib.parse import urljoin
+
+import requests
+from attrs import Attribute, Factory, define, field
+
+from griptape.drivers import BaseRulesetDriver
+from griptape.utils import dict_merge
+
+if TYPE_CHECKING:
+    from griptape.rules import BaseRule
+
+
+@define(kw_only=True)
+class GriptapeCloudRulesetDriver(BaseRulesetDriver):
+    """A driver for storing conversation memory in the Griptape Cloud.
+
+    Attributes:
+        ruleset_id: The ID of the Thread to store the conversation memory in. If not provided, the driver will attempt to
+            retrieve the ID from the environment variable `GT_CLOUD_THREAD_ID`. If that is not set, a new Thread will be
+            created.
+        base_url: The base URL of the Griptape Cloud API. Defaults to the value of the environment variable
+            `GT_CLOUD_BASE_URL` or `https://cloud.griptape.ai`.
+        api_key: The API key to use for authenticating with the Griptape Cloud API. If not provided, the driver will
+            attempt to retrieve the API key from the environment variable `GT_CLOUD_API_KEY`.
+
+    Raises:
+        ValueError: If `api_key` is not provided.
+    """
+
+    ruleset_id: str = field(
+        default=None,
+        metadata={"serializable": True},
+    )
+    base_url: str = field(
+        default=Factory(lambda: os.getenv("GT_CLOUD_BASE_URL", "https://cloud.griptape.ai")),
+    )
+    api_key: Optional[str] = field(default=Factory(lambda: os.getenv("GT_CLOUD_API_KEY")))
+    headers: dict = field(
+        default=Factory(lambda self: {"Authorization": f"Bearer {self.api_key}"}, takes_self=True),
+        init=False,
+    )
+
+    @api_key.validator  # pyright: ignore[reportAttributeAccessIssue]
+    def validate_api_key(self, _: Attribute, value: Optional[str]) -> str:
+        if value is None:
+            raise ValueError(f"{self.__class__.__name__} requires an API key")
+        return value
+
+    def load(self, ruleset_name: str) -> tuple[list[BaseRule], dict[str, Any]]:
+        """Load the ruleset from Griptape Cloud, using the ruleset name as an alias if ruleset_id is not provided."""
+        ruleset = None
+
+        if self.ruleset_id is not None:
+            res = self._call_api("get", f"/rulesets/{self.ruleset_id}", raise_for_status=False)
+            if res.status_code == 200:
+                ruleset = res.json()
+
+        # use name as 'alias' to get ruleset
+        if ruleset is None:
+            res = self._call_api("get", f"/rulesets?alias={ruleset_name}").json()
+            if res.get("rulesets"):
+                ruleset = res["rulesets"][0]
+
+        # no ruleset by name or ruleset_id
+        if ruleset is None:
+            if self.raise_not_found:
+                raise ValueError(f"No ruleset found with alias: {ruleset_name} or ruleset_id: {self.ruleset_id}")
+            return [], {}
+
+        rules = self._call_api("get", f"/rules?ruleset_id={ruleset['ruleset_id']}").json().get("rules", [])
+
+        for rule in rules:
+            rule["metadata"] = dict_merge(rule.get("metadata", {}), {"griptape_cloud_rule_id": rule["rule_id"]})
+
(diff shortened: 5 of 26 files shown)


griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @api_key.validator  # pyright: ignore[reportAttributeAccessIssue]
+    def validate_api_key(self, _: Attribute, value: Optional[str]) -> str:
+        if value is None:
+            raise ValueError(f"{self.__class__.__name__} requires an API key")
+        return value
+
+    def load(self, name: Optional[str] = None) -> tuple[list[BaseRule], dict[str, Any]]:
+        if name is None and self.ruleset_id is None:
+            raise ValueError("No ruleset name or ruleset_id provided")
```
vachillo, line 55: should this raise an error or just return `[], {}`?

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # if ruleset not found, try to get ruleset by ruleset_id
+        if ruleset is None and self.ruleset_id is not None:
+            res = self._call_api("get", f"/rulesets/{self.ruleset_id}", raise_for_status=False)
+            if res.status_code == 200:
+                ruleset = res.json()
+
+        # no ruleset by name or ruleset_id, raise error
+        if ruleset is None:
+            raise ValueError(f"No ruleset found with alias: {name} or ruleset_id: {self.ruleset_id}")
```
vachillo, line 73: should this raise an error or just return `[], {}`?

griptape/drivers/ruleset/local_ruleset_driver.py
```diff
@@ -0,0 +1,32 @@
...
+            file_name is not None
+            and os.path.exists(file_name)
+            and (loaded_str := Path(file_name).read_text()) is not None
+        ):
+            try:
+                return self._from_ruleset_dict(json.loads(loaded_str))
+            except Exception as e:
+                raise ValueError(f"Unable to load data from {file_name}") from e
+
+        return [], {}
```
vachillo, line 32: should this even exist? is there value in loading in rules from a file? it also has to be a specific format that is not really clear with how a `Rule` and `Ruleset` currently exist

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
cjkindel: Agree on error. Could the logic be:

```
if name:
 #Fetch by name
elif ruleset:
 #Fetch by ruleset_id
else:
 raise ValueError?
```

I would be surprised if I provided a name and I got back a ruleset that didn't have that name but matched some default ruleset_id I set earlier in my code

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
```diff
@@ -0,0 +1,90 @@
...
+    Raises:
+        ValueError: If `api_key` is not provided.
+    """
+
+    ruleset_id: str = field(
+        default=None,
+        metadata={"serializable": True},
+    )
+    base_url: str = field(
+        default=Factory(lambda: os.getenv("GT_CLOUD_BASE_URL", "https://cloud.griptape.ai")),
```
cjkindel, line 39: We should apply this pattern to all of our GriptapeCloud drivers (GT_Cloud mixin?). Looks like some use GT_CLOUD_BASE_URL, some just default to cloud.griptape.ai (Not this PR)

griptape/drivers/ruleset/local_ruleset_driver.py
cjkindel: I think value is there. Documentation will be necessary here, maybe even a helper fn that generates a file/json in correct format.

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
vachillo: this issue goes away with a proper SDK

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
vachillo: yeah the current logic chain is: fetch by name -> if name (alias) doesnt exist, fetch by ruleset_id -> if ruleset_id doesnt exist, fail

cjkindel: I would expect it to be: if name was provided, fetch by name/alias. If it wasn't provided, fetch by ruleset id. If either of those came up empty, raise.

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
vachillo: starting using this, i think i'm in favor of not raising an error. if its being used in an environment where the ruleset you are looking for may or may not exist, you wouldnt want to error out. for example, mapping a ruleset to a user id of some kind, or message thread. a ruleset doesn't necessarily exist for those things every time.

vachillo: i added a paramter `raise_not_found` so this is togglable

cjkindel: "It's Easier to Ask Forgiveness than Permission" is the pythonic pattern I think we should use. As a user, I would rather have to write a try/catch with handling for a non-existent ruleset than not know when a ruleset I thought existed does exist

vachillo: dont disagree but that gets complicated when its a griptape primitive. are you thinking the parameter should default to True?

cjkindel: Is that because Structures/chain of thought don't handle when a driver/tool raises an exception?

cjkindel: Happy to go with whatever team wants, but the following seems easy/clean to me:
```
rulesets = []

try:
  matts_cool_ruleset = Ruleset(name="filename.txt", ruleset_driver=LocalRulesetDriver())
except RulesetNotFound as e:
  logger.warning("matts_cool_ruleset doesn't exist!"
else:
  rulesets.append(matts_cool_ruleset)

agent = Agent(rulesets=rulesets)
```

griptape/drivers/ruleset/base_ruleset_driver.py
```diff
@@ -0,0 +1,29 @@
...
+    @abstractmethod
+    def load(self, name: Optional[str] = None) -> tuple[list[BaseRule], dict[str, Any]]: ...
+
+    def _from_ruleset_dict(self, params_dict: dict[str, Any]) -> tuple[list[BaseRule], dict[str, Any]]:
+        return [
+            self._get_rule(rule["value"], rule.get("meta", {})) for rule in params_dict.get("rules", [])
+        ], params_dict.get("meta", {})
+
+    def _get_rule(self, value: Any, meta: dict[str, Any]) -> BaseRule:
+        from griptape.rules import JsonSchemaRule, Rule
```
collindutter, line 27: Confirming: circular dependency?

griptape/drivers/ruleset/base_ruleset_driver.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @abstractmethod
+    def load(self, name: Optional[str] = None) -> tuple[list[BaseRule], dict[str, Any]]: ...
+
+    def _from_ruleset_dict(self, params_dict: dict[str, Any]) -> tuple[list[BaseRule], dict[str, Any]]:
+        return [
+            self._get_rule(rule["value"], rule.get("meta", {})) for rule in params_dict.get("rules", [])
+        ], params_dict.get("meta", {})
+
+    def _get_rule(self, value: Any, meta: dict[str, Any]) -> BaseRule:
```
collindutter, line 26: Let's tighten up `Any` to `dict | str`.

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
```diff
@@ -0,0 +1,94 @@
...
+    @api_key.validator  # pyright: ignore[reportAttributeAccessIssue]
+    def validate_api_key(self, _: Attribute, value: Optional[str]) -> str:
+        if value is None:
+            raise ValueError(f"{self.__class__.__name__} requires an API key")
+        return value
+
+    def load(self, name: Optional[str] = None) -> tuple[list[BaseRule], dict[str, Any]]:
+        if name is None and self.ruleset_id is None:
+            if self.raise_not_found:
+                raise ValueError("No ruleset name or ruleset_id provided")
```
collindutter, line 56: Let's make this error affirmative. Something like: "A Ruleset `name` or `ruleset_id` is required."

griptape/drivers/ruleset/base_ruleset_driver.py
```diff
@@ -0,0 +1,29 @@
...
+if TYPE_CHECKING:
+    from griptape.rules import BaseRule
+
+
+@define
+class BaseRulesetDriver(SerializableMixin, ABC):
+    raise_not_found: bool = field(default=False, kw_only=True, metadata={"serializable": True})
+
+    @abstractmethod
+    def load(self, name: Optional[str] = None) -> tuple[list[BaseRule], dict[str, Any]]: ...
```
collindutter, line 19: Can you please add a docstring?

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
```diff
@@ -0,0 +1,94 @@
...
+        init=False,
+    )
+
+    @api_key.validator  # pyright: ignore[reportAttributeAccessIssue]
+    def validate_api_key(self, _: Attribute, value: Optional[str]) -> str:
+        if value is None:
+            raise ValueError(f"{self.__class__.__name__} requires an API key")
+        return value
+
+    def load(self, name: Optional[str] = None) -> tuple[list[BaseRule], dict[str, Any]]:
```
collindutter, line 53: Feels kind of weird that `ruleset_id` is a field but `name` is an argument. Can we make them both kwargs?

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
```diff
@@ -0,0 +1,94 @@
...
+        if name is not None:
+            res = self._call_api("get", f"/rulesets?alias={name}").json()
+            if res.get("rulesets"):
+                ruleset = res["rulesets"][0]
+
+        # if ruleset not found, try to get ruleset by ruleset_id
+        if ruleset is None and self.ruleset_id is not None:
+            res = self._call_api("get", f"/rulesets/{self.ruleset_id}", raise_for_status=False)
+            if res.status_code == 200:
+                ruleset = res.json()
```
collindutter, line 71: Can drop the `if ruleset is None`, right? Unless the user provided both `ruleset_id` and `alias`?

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
```diff
@@ -0,0 +1,94 @@
...
+        # no ruleset by name or ruleset_id
+        if ruleset is None:
+            if self.raise_not_found:
+                raise ValueError(f"No ruleset found with alias: {name} or ruleset_id: {self.ruleset_id}")
+            return [], {}
+
+        rules = self._call_api("get", f"/rules?ruleset_id={ruleset['ruleset_id']}").json().get("rules", [])
+
+        for rule in rules:
+            rule["metadata"] = dict_merge(rule.get("metadata", {}), {"rule_id": rule["rule_id"]})
```
collindutter, line 82: Feels like we may want to indicate that `rule_id` is Griptape Cloud's resource id.

collindutter (changes requested):
Nice work, just a couple minor public API things. 

And docs + changelog.

griptape/drivers/ruleset/base_ruleset_driver.py
vachillo: `BaseRule` is typed `value: `Any`.

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
vachillo: the idea is that you could set `ruleset_id` on `GriptapeCloudDriver` init because its a specific field to griptape cloud. vs name is coming from the `Ruleset` which doesnt care about what the driver is doing with the name, its just giving it as a parameter. two different ways to potentially get a ruleset.

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
vachillo: situation where they provided `name (alias)` and we couldn't find a ruleset with that alias. if we dropped the `None` check, and a ruleset is found with the given alias but they also gave a ruleset_id, then it would be overwritten. 

i would say to throw an error if both are given, but that kinda breaks how rulesets currently are expected to work.

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
collindutter: It still feels a bit odd to have multiple places you can pass resource identification. Should this just be `def load()` and add `alias` to the Cloud Driver fields? I know we'd lose the convenience of using `Ruleset.name`, but it's not that bad.

vachillo: offline discussion

griptape/drivers/ruleset/griptape_cloud_ruleset_driver.py
vachillo: i switched `raise_not_found` to default to `True` to be more explicit.

docs/griptape-framework/drivers/ruleset-drivers.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Overview
+
+Ruleset Drivers can be used to load rules in from external sources. 
+
+## Ruleset Drivers
+
+### Local
+
+The [LocalRulesetDriver](../../reference/griptape/drivers/ruleset/local_ruleset_driver.md) allows you to load a Ruleset from a local JSON file. If no `persist_file` parameter is given, the `.load` method is a no-op.
```
collindutter, line 14: Should talk about `persist_dir`

docs/griptape-framework/drivers/src/griptape_cloud_ruleset_driver.py
```diff
@@ -0,0 +1,7 @@
+from griptape.drivers import GriptapeCloudRulesetDriver
+from griptape.rules import Ruleset
+
+rulset = Ruleset(
+    name="my_griptape_cloud_ruleset_alias",
```
collindutter, line 5: Can you create a real ruleset in the service account that we run integ tests against?

docs/griptape-framework/drivers/src/local_ruleset_driver.py
```diff
@@ -0,0 +1,7 @@
+from griptape.drivers import LocalRulesetDriver
+from griptape.rules import Ruleset
+
+ruleset = Ruleset(
+    name="my_local_ruleset.json",
+    ruleset_driver=LocalRulesetDriver(persist_dir="/path/to/ruleset/dir", raise_not_found=False),
```
collindutter, line 6: Would like this to run as well. Can you add some set up file system logic before we load in?

collindutter (changes requested):
Looks good aside from some docs tweaks

collindutter (approved):

docs/griptape-framework/drivers/src/local_ruleset_driver.py
```diff
@@ -16,9 +16,9 @@
 {
     "rules": [
         {
-            "value": "Always talk like a pirate.",
-        },
-    ],
+            "value": "Always talk like a pirate."
+        }
+    ]
```
collindutter, line 21: Can we just do `f.write(Ruleset(...).to_json())`?

vachillo: we can but the point is that the schema doesnt have to be a serialized ruleset

collindutter: Ok that's fine (though we should probably make them so). Can you at least just do `f.write(json.dumps({...whatever schema...}))`

griptape/drivers/memory/conversation/griptape_cloud_conversation_memory_driver.py
```diff
@@ -37,6 +37,10 @@ class GriptapeCloudConversationMemoryDriver(BaseConversationMemoryDriver):
         default=None,
         metadata={"serializable": True},
     )
+    alias: str = field(
+        default=None,
+        metadata={"serializable": True},
+    )
```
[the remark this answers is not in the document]
vachillo: removed for another PR

PR #1218, opened 2024-10-02
Add media tags to markdownify exclude list

cjkindel:

## Describe your changes
With the MarkdownifyWebScraperDriver returning TextArtifact, media content can be excluded from the default output. This will reduce the noise and size of webscraper output


collindutter (changes requested):
Looks good! Can you please update the changelog?

collindutter (approved):

Issue #1155, opened 2024-09-09
prompt_task.py prompt_stack should check weather self.structure is None


Chuan-Zh:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Describe the bug**
If use prompt_task without a structure, access to self.structure might raise None exception.

**To Reproduce**
A minimal, reproducible code example.
``` python
text_artifact = TextArtifact(
    "Respond to the user's following question 'who are you' in the language '{{preferred_language}}' and tone '{{tone}}'."
)

prompt_task = PromptTask(
    text_artifact,
    context={"preferred_language": "ENGLISH", "tone": "PLAYFUL"},
)

print(prompt_task.prompt_stack.to_json())
```

collindutter:
Thanks for the issue @Chuan-Zh! Would an improved exception be sufficient for your use-case, or are you looking to use Tasks without a Structure?

Chuan-Zh:
An improved exception is ok.

> Thanks for the issue @Chuan-Zh! Would an improved exception be sufficient for your use-case, or are you looking to use Tasks without a Structure?

Issue #1154, opened 2024-09-07
For base_task, task.run() will work with context if no structure is provided.


Chuan-Zh:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
``` python
prompt_task = PromptTask(
    "Respond to the user's following question 'who are you' in the language '{{preferred_language}}' and tone '{{tone}}'.",
    context={"preferred_language": "ENGLISH", "tone": "PLAYFUL"},
)
```
if just create a task like this, the context will not filled in when prompt_task.run().

**Describe the solution you'd like**
the context for prompt_task work without a  structure.

in base_task.py
``` python 
    @property
    def full_context(self) -> dict[str, Any]:
        if self.structure:
            structure_context = self.structure.context(self)

            structure_context.update(self.context)

            return structure_context
        else:
            return self.context
```

**Describe alternatives you've considered**
None

**Additional context**
None

collindutter:
Are you able to provide a fully reproducible example? With this example, I run into #1155 first:

```python
from griptape.tasks import PromptTask

prompt_task = PromptTask(
    "Respond to the user's following question 'who are you' in the language '{{preferred_language}}' and tone '{{tone}}'.",
    context={"preferred_language": "ENGLISH", "tone": "PLAYFUL"},
)

prompt_task.run()
```

```
Traceback (most recent call last):
  File "/Users/collindutter/Documents/griptape/griptape-playground/src/issues/1154.py", line 8, in <module>
    prompt_task.run()
  File "/Users/collindutter/Documents/griptape/griptape/griptape/tasks/prompt_task.py", line 81, in run
    message = self.prompt_driver.run(self.prompt_stack)
                                     ^^^^^^^^^^^^^^^^^
  File "/Users/collindutter/Documents/griptape/griptape/griptape/tasks/prompt_task.py", line 48, in prompt_stack
    memory = self.structure.conversation_memory
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'conversation_memory'
```

Thanks!

Chuan-Zh:
Could not because I also run into the error, so i fix the exception myself and continued.
I just want to find the inner working of Task, if it is not supported to work alone, I could use a structure.

PR #1223, opened 2024-10-03
Refactor RuleMixin

collindutter:

## Describe your changes
This PR was originally to solve https://github.com/griptape-ai/griptape/issues/1155 but it revealed an odd quirk of `RuleMixin`. `RuleMixin.structure` was added so that framework components that implement this mixin could inherit rules/rulesets from a Structure. In practice this is only really relevant in Tasks that want to inherit from their Structure. But Tasks already have a `structure` field which lead to some weirdness duplicate fields. 

This PR removes `RuleMixin.structure` and moves the inheritance logic to PromptTask. Additionally, inherited `rules` will be merged into the Structure's "Default Ruleset" instead of having a weird "Additional Ruleset". It also removes the restriction on passing both `rules` and `rulesets`. If you pass `rules`, they will just be added to a `Default Ruleset`.
Closes https://github.com/griptape-ai/griptape/issues/1155
Unblocks https://github.com/griptape-ai/griptape/issues/1154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **BREAKING**: `JsonExtractionEngine.template_schema` is now required.
 - **BREAKING**: `CsvExtractionEngine.column_names` is now required.
 - `StructureRunTask` now inherits from `PromptTask`.
+- **BREAKING**: Renamed`RuleMixin.all_rulesets` to `RuleMixin.rulesets`.
 - `JsonExtractionEngine.extract_artifacts` now returns a `ListArtifact[JsonArtifact]`.
 - `CsvExtractionEngine.extract_artifacts` now returns a `ListArtifact[CsvRowArtifact]`.
 - Remove `manifest.yml` requirements for custom tool creation.
@@ -74,6 +75,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Anthropic native Tool calling.
 - Empty `ActionsSubtask.thought` being logged.
+- `RuleMixin` no longer prevents setting `rulesets` _and_ `rules` at the same time.
+- `PromptTask` will merge in its Structure's Rulesets and Rules.
+- `PromptTask` not checking whether Structure was set before building Prompt Stack.
 
 ## [0.32.0] - 2024-09-17
 
--- a/griptape/tasks/base_task.py
+++ b/griptape/tasks/base_task.py
@@ -78,7 +78,7 @@ class BaseTask(FuturesExecutorMixin, ABC):
 
     @property
     def meta_memories(self) -> list[BaseMetaEntry]:
-        if self.structure and self.structure.meta_memory:
+        if self.structure is not None and self.structure.meta_memory:
             if self.max_meta_memory_entries:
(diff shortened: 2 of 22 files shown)


griptape/tasks/base_task.py
```diff
@@ -191,7 +191,7 @@ def run(self) -> BaseArtifact: ...
 
     @property
     def full_context(self) -> dict[str, Any]:
-        if self.structure:
+        if self.structure is not None:
```
vachillo, line 194: i know this is a small fix for this property, but why wouldnt this just return `self.context` if `self.structure` is None?

CHANGELOG.md
```diff
@@ -70,6 +70,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Anthropic native Tool calling.
 - Empty `ActionsSubtask.thought` being logged.
+- `RuleMixin` no longer prevents setting `rulesets` _and_ `rules` at the same time.
+- `PromptTask` will merge in its Structure's Rulesets and Rules.
+- `PromptTask` not checking whether Structure was set before building Prompt Stack.
```
vachillo, line 75: probably need a breaking change for `all_rulesets -> rulesets`

griptape/tasks/prompt_task.py
```diff
@@ -32,6 +33,22 @@ class PromptTask(RuleMixin, BaseTask):
...
     )
 
+    @property
+    def rulesets(self) -> list:
+        default_rules = self.rules
+        rulesets = self._rulesets
+
+        if self.structure is not None:
+            if self.structure._rulesets:
+                rulesets = self.structure._rulesets + self._rulesets
```
vachillo, line 43: its sort of odd to workaround the logic that is already in the mixin for the `ruleset` property right? can this be refactored at all to be less implentation-dependant? 

seems like we dont want two "default ruleset"s coming from the structure and task but maybe thats ok since they're defined two places?

CHANGELOG.md
collindutter: I had left it out since it felt more like an internal change, but I can add it back.

griptape/tasks/base_task.py
collindutter: That's in the [next PR](https://github.com/griptape-ai/griptape/pull/1224/files)

griptape/tasks/prompt_task.py
collindutter: > its sort of odd to workaround the logic that is already in the mixin for the ruleset property right?

I don't find it that odd -- we're overriding a field to add some context specific behavior. I could use `super().rulesets` to cut down on some duplicate code but then we need to figure out how to merge the two.

> seems like we dont want two "default ruleset"s coming from the structure and task but maybe thats ok since they're defined two places?

I don't think we should create two rulesets. A "Default Ruleset" is already ambiguous enough, having a "Default Structure Ruleset" and a "Default Task Ruleset" feels unnecessarily confusing for the LLM.

CHANGELOG.md
vachillo: might be worth keeping just because it was a public property.

griptape/tasks/prompt_task.py
vachillo: should the default name just be `Rules` or something then?

vachillo (approved):

PR #1250, opened 2024-10-09
Prompt Driver Debug Logs

collindutter:

### Added
- `LoggingConfig.handlers` for configuring logging handlers.
- `LoggingConfig.handlers_formatter` for configuring a formatter for all handlers.
- `LoggingConfig.propagate` for configuring whether log messages should be propagated to parent loggers.
- `LoggingConfig.level` for configuring the logging level.
- `griptape.configs.logging.JsonFormatter` for formatting logs as JSON.
- Request/response debug logging to all Prompt Drivers.
### Changed
- `_DefaultsConfig.logging_config` and `Defaults.drivers_config` are now lazily instantiated.

Closes https://github.com/griptape-ai/griptape/issues/267

### DIFF
--- a/griptape/configs/logging/json_formatter.py
+++ b/griptape/configs/logging/json_formatter.py
@@ -13,5 +13,7 @@ class JsonFormatter(logging.Formatter):
         super().__init__()
 
     def format(self, record: Any) -> str:
-        record.msg = json.dumps(record.msg, indent=self.indent)
+        if isinstance(record.msg, dict):
+            record.msg = json.dumps(record.msg, indent=self.indent)
+
         return super().format(record)
--- a/griptape/configs/logging/json_formatter.py
+++ b/griptape/configs/logging/json_formatter.py
@@ -1,0 +1,17 @@
+import json
+import logging
+from typing import Any
+
+from attrs import define, field
+
+
+@define
+class JsonFormatter(logging.Formatter):
+    indent: int = field(default=2, kw_only=True)
+
+    def __attrs_pre_init__(self) -> None:
+        super().__init__()
+
+    def format(self, record: Any) -> str:
+        record.msg = json.dumps(record.msg, indent=self.indent)
+        return super().format(record)
--- a/griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
+++ b/griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from typing import TYPE_CHECKING, Any
 
 from attrs import Factory, define, field
@@ -28,6 +29,7 @@ from griptape.common import (
     ToolAction,
     observable,
 )
+from griptape.configs import Defaults
 from griptape.drivers import BasePromptDriver
 from griptape.tokenizers import AmazonBedrockTokenizer, BaseTokenizer
 from griptape.utils import import_optional_dependency
@@ -40,6 +42,8 @@ if TYPE_CHECKING:
 
     from griptape.common import PromptStack
     from griptape.tools import BaseTool
+
+logger = logging.getLogger(Defaults.logging_config.logger_name)
 
 
 @define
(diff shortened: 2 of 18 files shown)


griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
```diff
@@ -41,6 +43,8 @@
     from griptape.common import PromptStack
     from griptape.tools import BaseTool
 
+logger = logging.getLogger(Defaults.logging_config.logger_name)
+
```
vachillo, line 47: instantiating the loggers here blows up the lazy logging config, as these get set on import

collindutter: I think that's fine, the lazy loading was mostly to get around circular dependencies. Should I remove `lazy_property` on `logging_config`?

vachillo: well i think the loggers need to be instantiated later. as implemented, if you update any properties on the logging config via `Defaults`, nothing will actually get updated because the config gets set on import

vachillo: unless you do all the logging config stuff before importing anything, which seems ugly

collindutter: Or you create a new `LoggingConfig` as I do in the example. Not _great_ but I think it's pretty common practice to have loggers declared up top.

vachillo: does that example work? the logger has already been fetched on import, updating the logging config afterwards wouldnt do anything i think

collindutter: It does work, but I'm realizing it's only because creating a new `LoggingConfig()` calls the `post_init` logic.

I guess only real solution here is to move all instances of logger into the methods.

collindutter: I'm now remembering that we intentionally kept functionality like this out of `LoggingConfig` to avoid unnecessary abstraction.

What about reverting `LoggingConfig` and changing example to:

```python
import logging

from griptape.configs import Defaults
from griptape.configs.logging import JsonFormatter

logger = logging.getLogger(Defaults.logging_config.logger_name)
logger.setLevel(logging.DEBUG)
logger.handlers[0].setFormatter(JsonFormatter())

agent = Agent()

agent.run("Hello world!")
```

vachillo: i kinda like the idea of exposing a `logger` parameter on the config and calling `config.logger` every time, but i know thats not very pythonic

collindutter: That feels out of place for the `Logging*Config*` though. Tbh this class probably should not exist and logger name should just be a constant somewhere else.

vachillo:
```
```
maybe im misunderstanding how fetching loggers works, but I thought the logger gets fetched on import and any subsequent changes to the logger would not get picked up.

vachillo: > That feels out of place for the `Logging*Config*` though. Tbh this class probably should not exist and logger name should just be a constant somewhere else.

```
from griptape.config import LOGGER_NAME
```

collindutter: > maybe im misunderstanding how fetching loggers works, but I thought the logger gets fetched on import and any subsequent changes to the logger would not get picked up.

`logger = logging.getLogger(...)` always refers to the same griptape logger. So changing values like debug level, formatter, etc post init is fine.

kyleroche (approved):
tested. works great.

griptape/configs/logging/json_formatter.py
```diff
@@ -0,0 +1,19 @@
...
+
+from attrs import define, field
+
+
+@define
+class JsonFormatter(logging.Formatter):
+    indent: int = field(default=2, kw_only=True)
+
+    def __attrs_pre_init__(self) -> None:
+        super().__init__()
```
vachillo, line 13: does this need to get set on the filters too?

vachillo (approved):

griptape/configs/logging/json_formatter.py
collindutter: Not sure, I just ran into errors without it.

PR #1258, opened 2024-10-14
Prompt the LLM to retry/fix actions when using native tools

collindutter:

## Describe your changes
I've noticed that when using native tool calling the LLM is much more likely to give up early. This may be because we don't instruct the LLM to retry/fix as many times as needed like we do for non native tool calling. This change moves around the prompt a bit to include this wording for both kinds of tool calling.
NA


vasinov (approved):
Nice prompt refactor!

Issue #285, opened 2023-09-26
Normalize `Callable` naming convention


vasinov:
Multiple Griptape classes have `Callable` properties. We use the following naming convention for them:

1. `_func`. For example `SnowflakeSqlDriver.connection_func`.
2. `_fn`. For example, `LocalVectorStoreDriver.relatedness_fn`.
3. Method-like. For example, `BasePromptDriver.prompt_stack_to_string`.

The team decided to normalize all callable properties to follow the `_fn` naming convention.

collindutter:
After revisiting the naming conventions, we decided against using the `_fn` suffix, as it felt unpythonic and inconsistent with naming conventions for other field types. For instance, we don't use `_str` for string fields, so there's no need to treat callables differently.

For callable fields and methods, the following guidelines should be used:

1. **Action Functions**: Functions that perform an action should be named using a verb or verb phrase that clearly describes what the function does.
   - `SnowflakeSqlDriver.connection_func` → `SnowflakeSqlDriver.create_connection`
   - `LocalVectorStoreDriver.relatedness_fn` → `LocalVectorStoreDriver.calculate_relatedness`
   - `BasePromptDriver.prompt_stack_to_string` → `BasePromptDriver.convert_prompt_stack_to_string`

2. **Boolean Functions**: Functions that return a boolean should be prefixed with predicates such as `is_`, `has_`, or `can_` to indicate a state or condition.

3. **Event Handlers**: Functions that react to events should be prefixed with `on_`. These functions typically don’t return a value but are triggered in response to an event. If a function transforms an event or returns a value based on it, it should follow the first rule (e.g., `process_event`).

Keep in mind that these are **guidelines**, not strict rules, and can be adjusted based on context.

collindutter:
Some team members have noted that these naming conventions may be more appropriate for class methods than class fields. For example, `create_connection` could be misinterpreted as a boolean that indicates whether a connection should be created.

To avoid this confusion, we are standardizing class field naming to be noun-based. Below are some of the conventions I landed on in #1275.

*_factory: Used for functions that create a new instance of an object.
*_provider: Used for functions that supply a value. These functions may create new instances, as long as the value represents the same resource or value.
*_formatter: Formats an object or value, typically into a string or another representation.
*_callback: Used for functions that react to or handle a specific event or condition.
*_handler: Used for functions responsible for actively managing or processing an event or task.
*_calculator: Used for functions that compute or derive a specific value.
[...]

PR #1263, opened 2024-10-15
Add RunnableMixin, implement in Structures, Tasks, Tools

### LINKED ISSUE #416: Add before and after hooks for masking data in prompts
Some users would like to mask and unmask data that goes out to external APIs as part the LLM prompt (e.g., customer PII). To solve this problem I propose adding the following to the framework:

1. Add `BasePromptStackProcessor` class that contains `before_run` and `after_run` methods. Let's think of better names.
2. Add `prompt_stack_processors: list[PromptStackProcessork]` property that contains a list of optional prompt stack processors.
3. Run `PromptStackProcessor.before_run` in `BasePromptDriver.before_run` and PromptStackProcessor.after_run` in `BasePromptDriver.after_run` in all prompt processors.
4. Add `AmazonComprehendPiiProcessor` that uses Amazon Comprehend to filter out PII data.
5. Add `PromptDriverPiiProcessor` that can use an arbitrary prompt driver to filter out PII data.

### LINKED ISSUE #924: Add `RunnableMixin` for pre/post processing on arbitrary run execution

there is a lot of different areas of the framework that could benefit from pre/post processing hooks on data. 

implement a `RunnableMixin`:
```
class RunnableMixin:
process_before_run_fn: Callable[[Any], Any]
process_after_run_fn: Callable[[Any], Any]

def before_run(self, *args, **kwargs) -> None (or Any?): ....
def run(self, *args, **kwargs) -> Any: ...
def after_run(self, *args, **kwargs) -> None (or Any?): ...
```

**Additional context**

collindutter:

### Added
- `RunnableMixin` which adds `on_before_run` and `on_after_run` hooks.
### Changed
- **BREAKING**: Renamed `BaseTask.run` to `BaseTask.try_run`.
- **BREAKING**: Renamed `BaseTask.execute` to `BaseTask.run`.
- **BREAKING**: Renamed `BaseTask.can_execute` to `BaseTool.can_run`.
- **BREAKING**: Renamed `BaseTool.run` to `BaseTool.try_run`.
- **BREAKING**: Renamed `BaseTool.execute` to `BaseTool.run`.
- Implemented `RunnableMixin` in `Structure`, `BaseTask`, and `BaseTool`.
Closes #924 
Closes #416

### DIFF
--- a/griptape/mixins/runnable_mixin.py
+++ b/griptape/mixins/runnable_mixin.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Generic, Optional, TypeVar, cast
+
+from attrs import define, field
+
+# Generics magic that allows us to reference the type of the class that is implementing the mixin
+T = TypeVar("T", bound="RunnableMixin")
+
+
+@define()
+class RunnableMixin(ABC, Generic[T]):
+    """Mixin for classes that can be "run".
+
+    Implementing classes should pass themselves as the generic type to ensure that the correct type is used in the callbacks.
+
+    Attributes:
+        on_before_run: Optional callback that is called at the very beginning of the `run` method.
+        on_after_run: Optional callback that is called at the very end of the `run` method.
+    """
+
+    on_before_run: Optional[Callable[[T], None]] = field(kw_only=True, default=None)
+    on_after_run: Optional[Callable[[T], None]] = field(kw_only=True, default=None)
+
+    def before_run(self, *args, **kwargs) -> Any:
+        if self.on_before_run is not None:
+            self.on_before_run(cast(T, self))
+
+    @abstractmethod
+    def run(self, *args, **kwargs) -> Any: ...
+
+    def after_run(self, *args, **kwargs) -> Any:
+        if self.on_after_run is not None:
+            self.on_after_run(cast(T, self))
--- a/griptape/structures/agent.py
+++ b/griptape/structures/agent.py
@@ -74,6 +74,6 @@ class Agent(Structure):
 
     @observable
     def try_run(self, *args) -> Agent:
-        self.task.execute()
+        self.task.run()
 
         return self
--- a/griptape/structures/structure.py
+++ b/griptape/structures/structure.py
@@ -12,6 +12,7 @@ from griptape.memory import TaskMemory
 from griptape.memory.meta import MetaMemory
 from griptape.memory.structure import ConversationMemory, Run
 from griptape.mixins.rule_mixin import RuleMixin
+from griptape.mixins.runnable_mixin import RunnableMixin
 from griptape.mixins.serializable_mixin import SerializableMixin
 
 if TYPE_CHECKING:
@@ -21,7 +22,7 @@ if TYPE_CHECKING:
 
 
 @define
-class Structure(ABC, RuleMixin, SerializableMixin):
+class Structure(RuleMixin, SerializableMixin, RunnableMixin["Structure"], ABC):
     id: str = field(default=Factory(lambda: uuid.uuid4().hex), kw_only=True, metadata={"serializable": True})
     _tasks: list[Union[BaseTask, list[BaseTask]]] = field(
         factory=list, kw_only=True, alias="tasks", metadata={"serializable": True}
@@ -139,6 +140,7 @@ class Structure(ABC, RuleMixin, SerializableMixin):
 
     @observable
     def before_run(self, args: Any) -> None:
+        super().before_run(args)
         self._execution_args = args
 
         [task.reset() for task in self.tasks]
@@ -155,6 +157,7 @@ class Structure(ABC, RuleMixin, SerializableMixin):
 
     @observable
     def after_run(self) -> None:
+        super().after_run()
         if self.conversation_memory and self.output_task.output is not None:
             run = Run(input=self.input_task.input, output=self.output_task.output)
 
--- a/tests/mocks/mock_image_generation_task.py
+++ b/tests/mocks/mock_image_generation_task.py
@@ -1,4 +1,4 @@
-from attrs import define, field
+from attrs import Factory, define, field
 
 from griptape.artifacts import ImageArtifact, TextArtifact
 from griptape.tasks import BaseImageGenerationTask
@@ -6,7 +6,7 @@ from griptape.tasks import BaseImageGenerationTask
 
 @define
 class MockImageGenerationTask(BaseImageGenerationTask):
-    _input: TextArtifact = field(default="input")
+    _input: TextArtifact = field(default=Factory(lambda: TextArtifact("input")))
 
     @property
     def input(self) -> TextArtifact:
@@ -16,5 +16,5 @@ class MockImageGenerationTask(BaseImageGenerationTask):
     def input(self, value: str) -> None:
         self._input = TextArtifact(value)
 
-    def run(self) -> ImageArtifact:
+    def try_run(self) -> ImageArtifact:
         return ImageArtifact(value=b"image data", format="png", width=512, height=512)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Events `BaseChunkEvent`, `TextChunkEvent`, `ActionChunkEvent`.
 - `wrapt` dependency for more robust decorators.
 - `BasePromptDriver.extra_params` for passing extra parameters not explicitly declared by the Driver.
+- `RunnableMixin` which adds `on_before_run` and `on_after_run` hooks.
 
 ### Changed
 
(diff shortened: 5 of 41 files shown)


vachillo:
doesnt `Structure.before_run()` need to call `super`?

collindutter:
Ah good catch, added to `Structure` and `BaseTool`.

Also, I'd like to come up with a clear plan for https://github.com/griptape-ai/griptape/issues/285 before introducing two new ones.

vachillo:
> Ah good catch, added to `Structure` and `BaseTool`.
> 
> Also, I'd like to come up with a clear plan for #285 before introducing two new ones.

last i remember, we decided on `_fn` for everything. however there is some nuance between a provided function that returns some value vs just runs and returns `None`.

griptape/structures/agent.py
```diff
@@ -74,6 +74,6 @@ def add_tasks(self, *tasks: BaseTask | list[BaseTask]) -> list[BaseTask]:
 
     @observable
     def try_run(self, *args) -> Agent:
-        self.task.execute()
+        self.task.run()
```
dylanholmes, line 77: I like this consistency

griptape/structures/structure.py
```diff
@@ -20,7 +21,7 @@
 
 
 @define
-class Structure(ABC, RuleMixin):
+class Structure(RuleMixin, RunnableMixin["Structure"], ABC):
```
dylanholmes, line 24: Dang, didn't realize you need to explicitly specify that generic argument. Still nice reuse though.

griptape/structures/structure.py
```diff
@@ -133,8 +134,8 @@ def resolve_relationships(self) -> None:
                 if task.id not in child.parent_ids:
                     child.parent_ids.append(task.id)
 
-    @observable
     def before_run(self, args: Any) -> None:
+        super().before_run(args)
```
dylanholmes, line 138: Want to make sure I understand (1) the implications of moving `@observable` from here to the parent (like will this still instrument the child class like before?) and (2) why did you get the `TypeError` originally which prompted the change?

griptape/mixins/futures_executor_mixin.py
```diff
@@ -1,14 +1,13 @@
...
-from abc import ABC
 from concurrent import futures
 from typing import Callable, Optional
 
 from attrs import Factory, define, field
 
 
 @define(slots=False, kw_only=True)
-class FuturesExecutorMixin(ABC):
+class FuturesExecutorMixin:
```
dylanholmes, line 10: Did you remove `ABC` for "style" (like because this class has no abstract methods) or for function (because you got some error)?

tests/mocks/mock_image_generation_task.py
```diff
@@ -6,7 +6,7 @@
 
 @define
 class MockImageGenerationTask(BaseImageGenerationTask):
-    _input: TextArtifact = field(default="input")
+    _input: TextArtifact = field(default=TextArtifact("input"))
```
dylanholmes, line 9: I know this is just for unit tests, but you should still use a factory here, since `TextArtifact` is not a primitive and has state that can be mutated.

tests/unit/drivers/observability/test_open_telemetry_observability_driver.py
```diff
@@ -163,6 +163,8 @@ def test_observability_agent(self, driver, mock_span_exporter):
                 ExpectedSpan(name="Agent.run()", parent="main", status_code=StatusCode.OK),
                 ExpectedSpan(name="Agent.before_run()", parent="Agent.run()", status_code=StatusCode.OK),
                 ExpectedSpan(name="Agent.try_run()", parent="Agent.run()", status_code=StatusCode.OK),
+                ExpectedSpan(name="PromptTask.before_run()", parent="Agent.try_run()", status_code=StatusCode.OK),
+                ExpectedSpan(name="PromptTask.after_run()", parent="Agent.try_run()", status_code=StatusCode.OK),
```
dylanholmes, line 167: Well, this gives me some extra confidence that the move of `@observable` kept the same functionality.

Thank you unit tests 🙏 .

griptape/structures/structure.py
dylanholmes: For (1), I think the unit tests have me satisfied half satisfied. I'd like one empirical test to ensure that the child behavior is fully wrapped (do exceptions propogate? is timing accurate?).

For (2), I have a hunch its because of the use of generics, we have `self: T` in the parent class. When you call `super().before_run(args)`, is the type of `self`, really `T`? I wouldn't expect so. I have a funny feeling about `self: T`

griptape/structures/structure.py
collindutter: @dylanholmes was able to get rid of this

griptape/structures/structure.py
collindutter: Woohoo we solved this offline by being explicit about which parent we're calling

griptape/structures/structure.py
```diff
@@ -139,6 +140,7 @@ def resolve_relationships(self) -> None:
 
     @observable
     def before_run(self, args: Any) -> None:
+        RunnableMixin.before_run(self, args)
```
vachillo, line 143: why doesnt `super().before_run(args)` work here? seems kinda wonky to have to call it like this.

griptape/tasks/base_text_input_task.py
```diff
@@ -24,7 +25,7 @@ class BaseTextInputTask(RuleMixin, BaseTask, ABC):
         alias="input",
     )
 
-    @property
+    @cached_property
```
vachillo, line 28: is there concern here with having this cached but having something like `self.full_context` change? those changes wont get picked up

griptape/structures/structure.py
collindutter: There was some wacky multiple inheritance stuff going on. Although this syntax is a bit clunkier, I think it's much more explicit as to which parent we're calling into.

griptape/mixins/runnable_mixin.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import Any, Callable, Generic, Optional, TypeVar, cast
+
+from attrs import define, field
+
+# Generics magic that allows us to reference the type of the class that is implementing the mixin
+T = TypeVar("T", bound="RunnableMixin")
+
+
+@define()
+class RunnableMixin(ABC, Generic[T]):
```
dylanholmes, line 13: Maybe worth a comment that when used as a parent class, `T` must be the the child class.

For example, this will work:
```python
class Foo(RunnableMixin["Foo"]):
   ...
```

But this will result in weird errors:
```python
class Foo(RunnableMixin[Bar]):
   ...
```

griptape/structures/structure.py
```diff
@@ -21,7 +22,7 @@
 
 
 @define
-class Structure(ABC, RuleMixin, SerializableMixin):
+class Structure(RuleMixin, SerializableMixin, RunnableMixin, ABC):
```
dylanholmes, line 25: Can you make `ABC` the first one again? (all style)

griptape/tasks/base_text_input_task.py
dylanholmes: I have this question too. What is expected in a situation like this? (excuse any usage mistakes, focus on the general point of mutating full_context in between input() calls):

```python
t = Task("foo: {{ foo }}")
t.full_context["foo"] = "foo"
x = t.input()

assert x == 'foo: foo'

t.full_context["foo"] = "bar"
y = t.input()

# after mutating full_context, should we expect this?
assert y == 'foo: foo'
# or this?
assert y == 'foo: bar' 
```

griptape/structures/structure.py
vachillo: with this syntax, there is no inheritance happening though. we are bypassing it completely

griptape/tasks/base_text_input_task.py
vachillo: yep exactly. the property is cached so its always going to print out "foo: foo"

griptape/structures/structure.py
vachillo: it seems like `@observable` is the root cause with `super` not working

griptape/structures/structure.py
collindutter: Unfortunately not. Because `RunnableMixin` implements ABC we get this from pyright:
```
Cannot create consistent method ordering
```

griptape/tasks/base_text_input_task.py
collindutter: Ah I see. Yes, it will print out `foo: foo`. The cache does get invalidated when Task is reset, but I can see this leading to some weird and difficult bugs. Such as:

```python
def test_input_context(self):
        pipeline = Pipeline(
            tasks=[
                PromptTask(
                    "foo",
                    prompt_driver=MockPromptDriver(),
                    on_before_run=lambda task: print(task.children[0].input),
                ),
                PromptTask("{{ parent_output }}", prompt_driver=MockPromptDriver()),
            ]
        )

        pipeline.run()

        assert pipeline.tasks[1].input.value == "mock output 1"
```

Lemme see if I can revert and still get the behavior I'm looking to achieve.

collindutter: Reverted and we're good 👍

griptape/structures/structure.py
collindutter: Agreed that observable seems to be the root cause -- somehow it is mangling the method signature. I think it is a complex issue that will probably require introducing something like https://wrapt.readthedocs.io/en/master/ to resolve.

This solution is still doing everything [by the books,](https://www.geeksforgeeks.org/python-call-parent-class-method/) just a bit lower level than super. If we can resolve the observability issues in a separate PR I'm all for reverting to super(), but I don't think it should block this PR.

griptape/mixins/runnable_mixin.py
collindutter: Added docstring to mixin.

griptape/structures/structure.py
dylanholmes: I would really like to dig into this some more while we have the chance.

collindutter: Fix implemented [here](https://github.com/griptape-ai/griptape/pull/1273). Still unclear what wrapt does that fixes this over `functools.wraps`.

dylanholmes: I was in favor of wrapt from the beginning out of fear, but I'm also in favor of yagni. Now we get both.

Let's not forget to add a task to learn how wrapt does its magic! (and add unit tests for ensuring presence of magic)

dylanholmes (changes requested):
Try `super()` again

docs/griptape-tools/official-tools/prompt-summary-tool.md
```diff
@@ -4,7 +4,7 @@ The [PromptSummaryTool](../../reference/griptape/tools/prompt_summary/tool.md) e
 --8<-- "docs/griptape-tools/official-tools/src/prompt_summary_tool_1.py"
 ```
 
-```
+````
```
vachillo, line 7: wut

griptape/structures/structure.py
```diff
@@ -139,6 +140,7 @@ def resolve_relationships(self) -> None:
 
     @observable
     def before_run(self, args: Any) -> None:
+        super().before_run(args)
```
vachillo, line 143: awesome

vachillo (dismissed):

griptape/structures/structure.py
dylanholmes: super

dylanholmes (dismissed):
🐑 after wut

collindutter:
Resolved merge conflicts

dylanholmes (approved):

PR #1283, opened 2024-10-24
Fix/contextvars

collindutter:

### Added
- `griptape.utils.decorators.with_context` decorator for running functions with the current `contextvars` context.
### Changed
- Using `EventListener`s as a context manager now sets thread-local `EventListener`s on the Event Bus instead of adding/removing them.
### Fixed
- Issues when using `EventListener` as a context manager in a multi-threaded environment.

NA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `wrapt` dependency for more robust decorators.
 - `BasePromptDriver.extra_params` for passing extra parameters not explicitly declared by the Driver.
 - `RunnableMixin` which adds `on_before_run` and `on_after_run` hooks.
+- `griptape.utils.with_contextvars` utility for running functions with the current `contextvars` context.
 
 ### Changed
 
@@ -61,6 +62,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `EventListener.event_types` will now listen on child types of any provided type.
 - Only install Tool dependencies if the Tool provides a `requirements.txt` and the dependencies are not already met.
 - Implemented `RunnableMixin` in `Structure`, `BaseTask`, and `BaseTool`.
+- `EventBus`'s Event Listeners are now thread/coroutine-local. Event Listeners from the spawning thread will be automatically copied when using concurrent griptape features like Workflows.
 
 ### Fixed
 
@@ -69,6 +71,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Models occasionally hallucinating `memory_name` and `artifact_namespace` into Tool schemas when using `ToolkitTask`.
 - Models occasionally providing overly succinct final answers when using `ToolkitTask`.
 - Exception getting raised in `FuturesExecutorMixin.__del__`.
+- Issues when using `EventListener` as a context manager in a multi-threaded environment.
 
 ## \[0.33.1\] - 2024-10-11
 
--- a/griptape/engines/rag/stages/response_rag_stage.py
+++ b/griptape/engines/rag/stages/response_rag_stage.py
@@ -7,6 +7,7 @@ from attrs import define, field
 
 from griptape import utils
 from griptape.engines.rag.stages import BaseRagStage
+from griptape.utils import with_contextvars
 
 if TYPE_CHECKING:
     from griptape.engines.rag import RagContext
@@ -32,7 +33,7 @@ class ResponseRagStage(BaseRagStage):
         logging.info("ResponseRagStage: running %s retrieval modules in parallel", len(self.response_modules))
 
         results = utils.execute_futures_list(
-            [self.futures_executor.submit(r.run, context) for r in self.response_modules]
+            [self.futures_executor.submit(with_contextvars(r.run), context) for r in self.response_modules]
         )
 
         context.outputs = results
--- a/griptape/events/event_bus.py
+++ b/griptape/events/event_bus.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-import threading
+from contextvars import ContextVar
 from typing import TYPE_CHECKING
 
-from attrs import Factory, define, field
+from attrs import define
 
 from griptape.mixins.singleton_mixin import SingletonMixin
 
@@ -11,14 +11,21 @@ if TYPE_CHECKING:
     from griptape.events import BaseEvent, EventListener
 
 
+# Context Vars must be declared at the top module level.
+# Also, in-place modifications do not trigger the context var's `set` method
+# so we must reassign the context var with the new value when adding or removing event listeners.
+_event_listeners: ContextVar[list[EventListener]] = ContextVar("event_listeners", default=[])
+
+
 @define
 class _EventBus(SingletonMixin):
-    _event_listeners: list[EventListener] = field(factory=list, kw_only=True, alias="_event_listeners")
-    _thread_lock: threading.Lock = field(default=Factory(lambda: threading.Lock()), alias="_thread_lock")
-
     @property
     def event_listeners(self) -> list[EventListener]:
-        return self._event_listeners
+        return _event_listeners.get()
+
+    @event_listeners.setter
+    def event_listeners(self, event_listeners: list[EventListener]) -> None:
+        _event_listeners.set(event_listeners)
 
     def add_event_listeners(self, event_listeners: list[EventListener]) -> list[EventListener]:
         return [self.add_event_listener(event_listener) for event_listener in event_listeners]
@@ -28,24 +35,21 @@ class _EventBus(SingletonMixin):
             self.remove_event_listener(event_listener)
 
     def add_event_listener(self, event_listener: EventListener) -> EventListener:
-        with self._thread_lock:
-            if event_listener not in self._event_listeners:
-                self._event_listeners.append(event_listener)
+        if event_listener not in self.event_listeners:
+            self.event_listeners = self.event_listeners + [event_listener]
 
         return event_listener
 
     def remove_event_listener(self, event_listener: EventListener) -> None:
-        with self._thread_lock:
-            if event_listener in self._event_listeners:
-                self._event_listeners.remove(event_listener)
+        if event_listener in self.event_listeners:
+            self.event_listeners = [listener for listener in self.event_listeners if listener != event_listener]
 
     def publish_event(self, event: BaseEvent, *, flush: bool = False) -> None:
-        for event_listener in self._event_listeners:
+        for event_listener in self.event_listeners:
             event_listener.publish_event(event, flush=flush)
 
     def clear_event_listeners(self) -> None:
-        with self._thread_lock:
-            self._event_listeners.clear()
+        self.event_listeners = []
 
 
 EventBus = _EventBus()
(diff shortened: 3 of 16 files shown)


CHANGELOG.md
```diff
@@ -51,13 +52,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Models in `ToolkitTask` with native tool calling no longer need to provide their final answer as `Answer:`.
 - `EventListener.event_types` will now listen on child types of any provided type.
 - Only install Tool dependencies if the Tool provides a `requirements.txt` and the dependencies are not already met.
+- Using `EventListener`s as a context manager now sets thread-local `EventListener`s on the Event Bus instead of adding/removing them.
```
dylanholmes, line 55: Maybe say "Context-local" (or something like that) instead? It's better than thread-local cause contextvars work as expected for coroutines  too (which are in the same thread)

collindutter: Updated wording slightly

griptape/events/event_listener.py
```diff
@@ -12,6 +13,10 @@
 
 T = TypeVar("T", bound=BaseEvent)
 
+_active_listeners: contextvars.ContextVar[list[EventListener]] = contextvars.ContextVar(
+    "active_event_listeners", default=[]
+)
+
```
dylanholmes, line 19: Just curious, why did you choose to put `_active_listeners` in in `event_listener.py` as opposed to `event_bus.py`? I would imagine if you did it the other way this file would be almost unchanged.

I don't have a strong opinion, just want to hear the reasoning

griptape/utils/decorators.py
```diff
@@ -15,6 +17,16 @@
 )
 
 
+def with_context(wrapped: Callable) -> Callable:
+    ctx = contextvars.copy_context()
+
+    @wrapt.decorator
```
dylanholmes, line 23: Good. 😄

griptape/utils/decorators.py
```diff
@@ -15,6 +17,16 @@
 )
 
 
+def with_context(wrapped: Callable) -> Callable:
+    ctx = contextvars.copy_context()
```
dylanholmes, line 21: So this is not griptape-context specific then, is it? Should/can it be?

tests/unit/events/test_event_listener.py
```diff
@@ -168,22 +168,18 @@ def event_handler(event: BaseEvent):
         mock_event_listener_driver.publish_event.assert_called_once_with({"event": mock_event.to_dict()})
 
     def test_context_manager(self):
```
dylanholmes, line 170: I would like to see a test that shows event listeners are not propagated to new threads or thread pool jobs without the `with_context` util.

CHANGELOG.md
dylanholmes: Could update the PR description too for thoroughness

griptape/utils/decorators.py
dylanholmes: Like what if client code is using contextvars for something else and doesn't want it propagated? (Like could be a pretty weird/difficult to debug side effect if it did happen... like maybe another library is doing something with contextvars and the customer doesn't even realize....)

griptape/engines/rag/stages/response_rag_stage.py
```diff
@@ -32,7 +33,7 @@ def run(self, context: RagContext) -> RagContext:
         logging.info("ResponseRagStage: running %s retrieval modules in parallel", len(self.response_modules))
 
         results = utils.execute_futures_list(
-            [self.futures_executor.submit(r.run, context) for r in self.response_modules]
+            [self.futures_executor.submit(with_context(r.run), context) for r in self.response_modules]
```
dylanholmes, line 36: Unfortunate naming collision for readability, makes me do a double take.

griptape/events/event_listener.py
```diff
@@ -33,18 +38,17 @@ class EventListener(Generic[T]):
     _last_event_listeners: Optional[list[EventListener]] = field(default=None)
 
     def __enter__(self) -> EventListener:
-        from griptape.events import EventBus
-
-        EventBus.add_event_listener(self)
-
+        # Add this listener to the active listeners stack
```
dylanholmes, line 41: I'd prefer to remove this comment (stack is a little triggering).

tests/unit/utils/test_decorators.py
```diff
@@ -0,0 +1,37 @@
...
+        @with_context
+        def decorated_function(vals: list) -> None:
+            vals.append(context_var.get())
+
+        return_values = []
+        thread = threading.Thread(target=decorated_function, args=(return_values,))
+        thread.start()
+        thread.join()
+
+        assert return_values == ["test"]
```
dylanholmes, line 21: You should also test your test by showing an example without the decorator (same for the other test too).

tests/unit/tasks/test_base_task.py
```diff
@@ -112,9 +109,11 @@ def test_children_property_no_structure(self, task):
         assert len(parent.children) == 3
 
     def test_execute_publish_events(self, task):
-        task.execute()
+        handler = Mock()
+        with EventListener(handler):
+            task.execute()
 
-        assert EventBus.event_listeners[0].handler.call_count == 2
+            assert handler.call_count == 2
```
dylanholmes, line 116: This line doesn't need to be indented, right? (Like it should be the same inside or outside the context manager scope, right?)

tests/unit/tasks/test_base_task.py
```diff
@@ -14,11 +13,9 @@
 class TestBaseTask:
     @pytest.fixture()
     def task(self):
-        EventBus.add_event_listeners([EventListener(handler=Mock())])
```
dylanholmes, line 17: Wait, you can still do this one, too, right?

griptape/events/event_bus.py
```diff
@@ -18,7 +20,11 @@ class _EventBus(SingletonMixin):
 
     @property
     def event_listeners(self) -> list[EventListener]:
-        return self._event_listeners
+        # Check if there are active listeners in the context
+        active_listeners = _active_listeners.get()
+
+        # Use active listeners if any exist, otherwise use global listeners
+        return active_listeners if active_listeners else self._event_listeners
```
dylanholmes, line 27: I do not like this logic

dylanholmes: What if I want to clear the event listeners from with a thread/coroutine? (I actually had a use case for this before)

griptape/events/event_bus.py
```diff
@@ -11,14 +12,20 @@
     from griptape.events import BaseEvent, EventListener
 
 
+_event_listeners: contextvars.ContextVar[list[EventListener]] = contextvars.ContextVar("event_listeners", default=[])
+
+
 @define
 class _EventBus(SingletonMixin):
-    _event_listeners: list[EventListener] = field(factory=list, kw_only=True, alias="_event_listeners")
     _thread_lock: threading.Lock = field(default=Factory(lambda: threading.Lock()), alias="_thread_lock")
```
dylanholmes, line 20: We don't need this lock anymore do we?

griptape/events/event_bus.py
```diff
@@ -11,14 +12,20 @@
     from griptape.events import BaseEvent, EventListener
 
 
+_event_listeners: contextvars.ContextVar[list[EventListener]] = contextvars.ContextVar("event_listeners", default=[])
```
dylanholmes, line 15: Just curious, could this have been a field on the singleton?

dylanholmes (commented):
This is awesome; so clean -- just a couple more questions and then I just want to look through the tests

griptape/events/event_bus.py
collindutter: Great catch!

griptape/events/event_bus.py
collindutter: https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar

> Context Variables should be created at the top module level and never in closures. [Context](https://docs.python.org/3/library/contextvars.html#contextvars.Context) objects hold strong references to context variables which prevents context variables from being properly garbage collected.

dylanholmes: I almost feel like we should have a comment about this... but I leave that decision to you

griptape/events/event_bus.py
```diff
@@ -1,24 +1,28 @@
...
-    _thread_lock: threading.Lock = field(default=Factory(lambda: threading.Lock()), alias="_thread_lock")
-
     @property
     def event_listeners(self) -> list[EventListener]:
-        return self._event_listeners
+        return _event_listeners.get()
+
+    @event_listeners.setter
+    def event_listeners(self, event_listeners: list[EventListener]) -> None:
+        _event_listeners.set(event_listeners)
```
dylanholmes, line 25: Just for my own understanding, what would be the difference between setting the whole list versus updating the contents?

In other words what if this looked like:
```py
self.event_listeners.clear()
self.event_listeners.extend(event_listeners)
```

Like what if some class previously stored a reference to the return value from `EventBus.event_listeners`, then the setter was called? If you set the whole list then the stored reference will contain event listeners no longer in the event bus, is that the desired behavior?

Is there a restriction on mutating the value returned by `_event_listeners.get()` because of `contextvars`?

griptape/events/event_bus.py
```diff
@@ -28,24 +32,21 @@ def remove_event_listeners(self, event_listeners: list[EventListener]) -> None:
             self.remove_event_listener(event_listener)
 
     def add_event_listener(self, event_listener: EventListener) -> EventListener:
-        with self._thread_lock:
-            if event_listener not in self._event_listeners:
-                self._event_listeners.append(event_listener)
+        if event_listener not in self.event_listeners:
+            self.event_listeners = self.event_listeners + [event_listener]
```
dylanholmes, line 36: Just curious, why don't we use a `set` instead of a `list` so we get uniqueness for free? Doesn't seem like we are relying on order for anything.

tests/unit/events/test_event_listener.py
```diff
@@ -185,6 +186,28 @@ def test_context_manager_multiple(self):
 
         assert EventBus.event_listeners == [e1]
 
+    def test_threaded(self):
```
dylanholmes, line 189: This is great!

Could you also add a test for not using `copy_context_vars`?
And can you add the same with/without tests, but for creating a thread directly instead of a pool?
(And if possible, could you do the same with/without tests for coroutines?)

I know this is kind of neurotic, but I want tests for each of these situations so that we double check we don't have any invalid assumptions.

dylanholmes (changes requested):
Really cool!

2 Questions and 1 Ask

griptape/events/event_bus.py
collindutter: I think just an initial oversight. We can update but I'd prefer to do so outside of this PR since there's enough complexity as is.

griptape/events/event_bus.py
collindutter: The important thing is that `_event_listeners.set` is called at some point. While we could probably keep the in-place modifications like `append/remove/clear`, we'd need to call `set` later in order for the context variable to appear in `copy_context`. I can add a comment that explains this behavior.

dylanholmes: Thanks for the comment! Giving 🐑 , but will want to discuss offline

dylanholmes (approved):
Great work!

Will want to discuss the `set` behavior later, but I think its good enough for now 🍻

PR #1289, opened 2024-10-25
Add default for getting run_id from metadata

vachillo:

### DIFF
--- a/griptape/drivers/memory/conversation/griptape_cloud_conversation_memory_driver.py
+++ b/griptape/drivers/memory/conversation/griptape_cloud_conversation_memory_driver.py
@@ -128,7 +128,7 @@ class GriptapeCloudConversationMemoryDriver(BaseConversationMemoryDriver):
 
         runs = [
             Run(
-                **({"id": message["metadata"]} if "metadata" in message else {}),
+                **({"id": message["metadata"].pop("run_id", None)} if "run_id" in message.get("metadata") else {}),
                 meta=message["metadata"],
                 input=BaseArtifact.from_json(message["input"]),
                 output=BaseArtifact.from_json(message["output"]),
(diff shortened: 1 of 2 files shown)


griptape/drivers/memory/conversation/griptape_cloud_conversation_memory_driver.py
```diff
@@ -128,7 +128,7 @@ def load(self) -> tuple[list[Run], dict[str, Any]]:
 
         runs = [
             Run(
-                id=m["metadata"].pop("run_id"),
+                id=m["metadata"].pop("run_id", uuid.uuid4().hex),
```
collindutter, line 131: When would this occur? Seems odd to invent a cloud resource id.

vachillo: it would occur if that metadata key doesnt exist on the cloud Message, like if the message was inserted manually.

collindutter: Oh I see, `Run.id` already [has a default ](https://github.com/griptape-ai/griptape/blob/d69d2fc47083f2719b7b0f19a4fb3696bfdfd443/griptape/memory/structure/run.py#L16) so should just avoid setting `id` here if `run_id` is not present.

griptape/drivers/memory/conversation/griptape_cloud_conversation_memory_driver.py
```diff
@@ -126,15 +126,18 @@ def load(self) -> tuple[list[Run], dict[str, Any]]:
...
+            run_id = m["metadata"].pop("run_id", None)
+            run = Run(
                 meta=m["metadata"],
                 input=BaseArtifact.from_json(m["input"]),
                 output=BaseArtifact.from_json(m["output"]),
             )
-            for m in messages_response.get("messages", [])
-        ]
+            if run_id is not None:
+                run.id = run_id
```
collindutter, line 138: There could be subtle issues with assigning `id` after creation (future `attrs_post_init` logic). We should still do it at time of instantiation. Either with something like:

```python
        runs = [
            Run(
                **({"id": message["metadata"]} if "metadata" in message else {}),
                meta=message["metadata"],
                input=BaseArtifact.from_json(message["input"]),
                output=BaseArtifact.from_json(message["output"]),
            )
            for message in messages_response.get("messages", [])
        ]
```
Or by separating the param creation and run creation into two separate steps.

vachillo: updated

collindutter (approved):

PR #1303, opened 2024-10-30
Remove `docker` and `stringcase` from core dependencies

collindutter:

### Changed
- **BREAKING**: Removed `stringcase` and `docker` from core dependencies. `ComputerTool` will now install this on the fly.
NA

### DIFF
--- a/griptape/tools/computer/tool.py
+++ b/griptape/tools/computer/tool.py
@@ -8,10 +8,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Optional
 
 import docker
-import stringcase
 from attrs import Attribute, Factory, define, field
-from docker.errors import NotFound
-from docker.models.containers import Container
 from schema import Literal, Schema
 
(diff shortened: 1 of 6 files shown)


griptape/tools/computer/tool.py
```diff
@@ -97,6 +94,8 @@ def execute_command(self, params: dict) -> BaseArtifact:
         return self.execute_command_in_container(command)
 
     def execute_command_in_container(self, command: str) -> BaseArtifact:
+        from docker.models.containers import Container
+
```
vachillo, line 98: should these be wrapped with `import_optional_dependency`?

collindutter: We don't use `import_optional_dependency` for Tools since it's expected that the Tool will install them.

collindutter: Not saying this is ideal, but it's consistent with how other Tools work

vachillo (approved):

PR #1340, opened 2024-11-12
Replace Image Query Drivers with Prompt Drivers

collindutter:

## Describe your changes
Image Query Drivers were added early on before Prompt Drivers supported image inputs. Now they provide no value other than some syntactic niceties. This PR removes them and improves some syntax with Prompt Drivers.

Note that Image Query Tool has been kept since we don't have a Prompt Tool (though maybe we should...separate discussion).

### Added
- `PromptStack.from_artifact` factory method for creating a Prompt Stack with a user message from an Artifact.
### Changed
- **BREAKING**: Removed all `ImageQueryDriver`s, use `PromptDriver`s instead.
- **BREAKING**: Removed `ImageQueryTask`, use `PromptTask` instead.
- **BREAKING**: Updated `ImageQueryTool.image_query_driver` to `ImageQueryTool.prompt_driver`.
- `BasePromptDriver.run` can now accept an Artifact in addition to a Prompt Stack.
NA

----
📚 Documentation preview 📚: https://griptape--1340.org.readthedocs.build//1340/


docs/griptape-tools/official-tools/src/image_query_tool_1.py
```diff
@@ -1,13 +1,12 @@
...
+from griptape.drivers import OpenAiChatPromptDriver
 from griptape.structures import Agent
 from griptape.tools import ImageQueryTool
 
-# Create an Image Query Driver.
-driver = OpenAiImageQueryDriver(model="gpt-4o")
+driver = OpenAiChatPromptDriver(model="gpt-4o")
 
 # Create an Image Query Tool configured to use the engine.
 tool = ImageQueryTool(
```
emjay07, line 8: I'm curious about the use case when an `ImageQueryTool` is still useful? if the prompt is "describe this image", the model should be able to do that now without using a tool.

README.md
```diff
@@ -40,7 +40,6 @@ Drivers facilitate interactions with external resources and services:
 - 🔢 **Embedding Drivers** generate vector embeddings from textual inputs.
 - 💾 **Vector Store Drivers** manage the storage and retrieval of embeddings.
 - 🎨 **Image Generation Drivers** create images from text descriptions.
-- 🔎 **Image Query Drivers** query images from text queries.
```
emjay07, line 43: Up to you on this, but may be good to add a line in the prompt drivers that they can now handle multi-modal queries or something so it doesn't seem like we don't support it at all.

emjay07 (commented):
I think it'd be good to copy the examples from the migration.md to the `PromptDriver` or `PromptTask` docs or a recipe.

docs/griptape-tools/official-tools/src/image_query_tool_1.py
collindutter: There is still value in having a Tool that can query images from the file system/task memory. Long term this functionality should maybe be baked into the File Manager Tool but that would require refactors outside the scope of this PR.

README.md
dylanholmes: Maybe this list could focus on functionality (like verbs), then mention which driver gives it to you 🤷 .

dylanholmes (approved):
lgtm

PR #1342, opened 2024-11-13
Add griptape cloud tool tool

collindutter:

### Added
- `GriptapeCloudToolTool` for running Griptape Cloud hosted Tools.
NA


collindutter:
Not marking as ready for review until naming is finalized and feature is deployed.

.github/workflows/docs-integration-tests.yml
```diff
@@ -111,6 +111,7 @@ jobs:
       WEBHOOK_URL: ${{ secrets.INTEG_WEBHOOK_URL }}
       AMAZON_SQS_QUEUE_URL: ${{ secrets.INTEG_AMAZON_SQS_QUEUE_URL }}
       GT_CLOUD_STRUCTURE_RUN_ID: ${{ secrets.INTEG_GT_CLOUD_STRUCTURE_RUN_ID }}
+      GT_CLOUD_TOOL_ID: ${{ secrets.INTEG_GT_CLOUD_TOOL_ID }}
```
collindutter, line 114: @zachgiordano will need your help getting this

griptape/tools/griptape_cloud_tool/tool.py
```diff
@@ -0,0 +1,122 @@
...
+
+        if self.name == self.__class__.__name__:
+            self.name = tool_name
+
+        for activity_name, (description, activity_schema) in activity_schemas.items():
+            activity_handler = self.__create_activity_handler(activity_name, description, activity_schema)
+
+            setattr(self, activity_name, MethodType(activity_handler, self))
+
+    def __get_schema(self) -> dict:
```
dylanholmes, line 41: Why are we using leading double underscores? Are we trying to prevent users from overriding this method in subclasses?

As I understand it, adding leading double underscores will cause the class member's name to be mangled. In this case it will change `__get_schema` to `_GriptapeCloudToolTool__get_schema`. So my question is why?


> Or is this just related to `attrs`?

collindutter: I guess I was feeling overly private about these fields. Marking them as protected is more consistent with our usual behavior, I'll do that.

PR #1367, opened 2024-11-20
Add GriptapeCloudToolDriver

collindutter:

## Describe your changes
Alternate implementation to the one in [#1342](https://github.com/griptape-ai/griptape/pull/1342).

### Added
- New class of Drivers, Tool Drivers, for augmenting Tools with additional functionality.
- `GriptapeCloudToolDriver` for interacting with Griptape Cloud's Tool API.
- `BaseTool.tool_driver` for setting a Tool Driver on a Tool. The Tool will call `tool_driver.initialize_tool` after Tool initialization.

I'm not in love how we explain these Drivers in the docs, they're kind of hard to understand without the full picture of the Tools Cloud service.
Closes https://github.com/griptape-ai/griptape/pull/1342

----
📚 Documentation preview 📚: https://griptape--1367.org.readthedocs.build//1367/


.github/workflows/docs-integration-tests.yml
```diff
@@ -68,6 +68,7 @@ jobs:
       GT_CLOUD_STRUCTURE_ID: ${{ vars.INTEG_GT_CLOUD_STRUCTURE_ID }}
       GT_CLOUD_STRUCTURE_RUN_ID: ${{ vars.INTEG_GT_CLOUD_STRUCTURE_RUN_ID }}
       GT_CLOUD_THREAD_ID: ${{ vars.INTEG_GT_CLOUD_THREAD_ID }}
+      GT_CLOUD_TOOL_ID: ${{ vars.INTEG_GT_CLOUD_TOOL_ID }}
```
collindutter, line 71: @zachgiordano what would it take to get one of these in Tony?

griptape/utils/decorators.py
```diff
@@ -58,8 +59,8 @@ def lazy_attr(self: Any, value: Any) -> None:
 
 
 def _build_kwargs(func: Callable, params: dict) -> dict:
-    func_params = inspect.signature(func).parameters.copy()
-    func_params.pop("self")
+    func_params = cast(OrderedDict, inspect.signature(func).parameters.copy())
```
collindutter, line 62: Pop the first item, not the one named "self" (which is just a convention)

griptape/drivers/tool/base_tool_driver.py
```diff
@@ -0,0 +1,23 @@
...
+from attrs import define
+
+if TYPE_CHECKING:
+    from griptape.tools.base_tool import BaseTool
+
+
+@define
+class BaseToolDriver(ABC):
+    @abstractmethod
+    def initialize_tool(self, tool: BaseTool) -> None:
```
vachillo, line 15: might be useful to have a field like `raise_on_same_activity` so you can control if you want the local activities to be overwritten by the remote ones?

collindutter: When would you ever not want the local activities overwritten? What value would the Driver add in that case?

vachillo: if you wanted to add local activities in addition to the remote ones, just to be extra sure there is no naming collisions. it would be more of a self-describing param too - the fact that it exists and defaults to False makes it clear what the behavior of the driver is.

griptape/drivers/tool/griptape_cloud_tool_driver.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+
+    base_url: str = field(default=Factory(lambda: os.getenv("GT_CLOUD_BASE_URL", "https://cloud.griptape.ai")))
+    api_key: str = field(default=Factory(lambda: os.environ["GT_CLOUD_API_KEY"]))
+    tool_id: str = field(kw_only=True)
+    headers: dict = field(
+        default=Factory(lambda self: {"Authorization": f"Bearer {self.api_key}"}, takes_self=True),
+        init=False,
+    )
+    raise_on_same_activity: bool = field(default=True, kw_only=True)
```
vachillo, line 39: maybe this should default to `False`? or named something like `overwrite_activties`? basically i think the default behavior should be overwriting the existing locally-defined activities but there should be a way to toggle that off. sorry not super clear

collindutter: Oh yeah, duh. I still think `raise_on_same_activity` is a good name though.

vachillo: it is no more

vachillo (approved):
~really cool~ Tools

zachgiordano (approved):
Tools

PR #1371, opened 2024-11-21
Add converstion memory per task run

collindutter:

### Added
- `PromptTask.conversation_memory` for setting the Conversation Memory on a Prompt Task.

NA

### DIFF
--- a/griptape/tasks/prompt_task.py
+++ b/griptape/tasks/prompt_task.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Callable, Optional, Union
+from typing import TYPE_CHECKING, Callable, Literal, Optional, Union
 
-from attrs import Factory, define, field
+from attrs import NOTHING, Factory, define, field
 
 from griptape.artifacts import BaseArtifact, ListArtifact, TextArtifact
 from griptape.common import PromptStack
 from griptape.configs import Defaults
+from griptape.memory.structure import Run
 from griptape.mixins.rule_mixin import RuleMixin
 from griptape.rules import Ruleset
 from griptape.tasks import BaseTask
@@ -15,6 +16,8 @@ from griptape.utils import J2
 
 if TYPE_CHECKING:
     from griptape.drivers import BasePromptDriver
+    from griptape.memory.structure.base_conversation_memory import BaseConversationMemory
+    from griptape.structures import Structure
 
 logger = logging.getLogger(Defaults.logging_config.logger_name)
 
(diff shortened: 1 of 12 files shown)


griptape/tasks/prompt_task.py
```diff
@@ -93,12 +97,24 @@ def after_run(self) -> None:
...
     def try_run(self) -> BaseArtifact:
         message = self.prompt_driver.run(self.prompt_stack)
 
         return message.to_artifact()
 
+    def preprocess(self, structure: Structure) -> BaseTask:
+        super().preprocess(structure)
+
+        if self.conversation_memory is None and structure.conversation_memory is not None:
+            self.conversation_memory = structure.conversation_memory
```
vasinov, line 114: Won't this prevent the user from forcing task `conversation_memory` to `None`?

collindutter: Ah...right. This is where python's lack of a [sentinel value](https://peps.python.org/pep-0661/) is particularly challenging.

We need to differentiate between an unset value and an explicitly `None` value.

vachillo: [can we steal `attrs.NOTHING`?](https://www.attrs.org/en/stable/api.html#attrs.NOTHING)

collindutter: Oh nice, yes that's exactly what we need.

collindutter:
@vachillo @dylanholmes re-request.

griptape/tasks/prompt_task.py
```diff
@@ -28,6 +31,9 @@ class PromptTask(RuleMixin, BaseTask):
         default=Factory(lambda self: self.default_generate_system_template, takes_self=True),
         kw_only=True,
     )
+    conversation_memory: Union[Optional[BaseConversationMemory], Literal[NOTHING]] = field(  # pyright: ignore[reportInvalidTypeForm] TODO: Replace with [NothingType](https://github.com/python-attrs/attrs/pull/1358)
+        default=Factory(lambda: NOTHING), kw_only=True
+    )
```
vachillo, line 36: we can steal the `TypeAlias` implementation from that PR if that would make pyright happier

griptape/tasks/prompt_task.py
```diff
@@ -98,12 +104,33 @@ def after_run(self) -> None:
...
         message = self.prompt_driver.run(self.prompt_stack)
 
         return message.to_artifact()
 
+    def preprocess(self, structure: Structure) -> BaseTask:
+        super().preprocess(structure)
+
+        if self.conversation_memory is NOTHING:
+            if structure.conversation_memory is not None:
+                self.conversation_memory = structure.conversation_memory
```
vachillo, line 128: double checking, the `NOTHING` addition is just for this check here?

vachillo (approved):

griptape/tasks/prompt_task.py
collindutter: I don't see how their PR is making pyright happy since it still uses `Literal[Nothing]`.

griptape/tasks/prompt_task.py
collindutter: Yes. Since we need to support the user not setting conversation memory (`NOTHING`) _and_ the user explicitly disabling conversation memory (`None`).

PR #1377, opened 2024-11-22
Fix `with_contextvars` not properly wrapping functions in some cases.

collindutter:

### Fixed
- `with_contextvars` not properly wrapping functions in some cases.

The old implementation had some issues with wrapping some gradio functions, presumably because `functools.partial` fixes the arguments. This new implementation has a bit less magic and fixes the issue.

While it strongly resembles a decorator, that is a whole rabbit hole I don't want to go back down.
Closes upcoming issue


vachillo (approved):
any insight into why `partial` is messing this up or does this _just work_

collindutter:
> any insight into why partial is messing this up

I think gradio relies on inspecting function signatures and `partial` changes it enough to break things.

dylanholmes (commented):
Would like to understand why, but let's get it in.

PR #1386, opened 2024-11-26
Fix kwarg injection default values

vachillo:

## Describe your changes
the activity decorator currently overwrites default values with `None` if they are not passed to the activity.
context from below:

currently if you defined a schema with 3 params: `foo` , `bar`, `baz`, and your activity method was defined as:
```
@activity(...)
def activity_method(foo: str, bar: str, baz: str = "default"): ...
```
and the input looked like:
```
{
  "values": {
    "foo": "some_foo",
    "bar": "some_bar"
  },
}
```
the decorator would set `baz=None` rather than keeping the default that was defined on the method signature

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,7 +93,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `MarqoVectorStoreDriver.query` failing when `include_metadata` is `True`.
 - `with_contextvars` not properly wrapping functions in some cases.
 - Crash when calling `ToolkitTask.run()` directly.
-- Support default values using kwarg injection on `@activity` methods.
+- `@activity` decorator overwriting injected kwargs with default values as `None`.
 
 ## [0.34.3] - 2024-11-13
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -93,6 +93,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `MarqoVectorStoreDriver.query` failing when `include_metadata` is `True`.
 - `with_contextvars` not properly wrapping functions in some cases.
 - Crash when calling `ToolkitTask.run()` directly.
+- Support default values using kwarg injection on `@activity` methods.
```
collindutter, line 96: Is this meant to be under Fixed?

collindutter (commented):
Can you provide some context in the PR description?

CHANGELOG.md
vachillo: I think so. currently if you defined a schema with 3 params: `foo` , `bar`, `baz`, and your activity method was defined as:
```
def activity_method(foo: str, bar: str, baz: str = "default"): ...
```
and the input looked like:
```
{
  "values": {
    "foo": "some_foo",
    "bar": "some_bar"
  },
}
```
the decorator would set `baz=None` rather than keeping the default that was defined on the method signature

collindutter: Got it. Can you rephrase the changelog entry with an implicit "Fixed"?

collindutter (approved):

PR #1396, opened 2024-12-02
Add EvalEngine

collindutter:

### Added
- `EvalEngine` for evaluating the performance of an LLM's output against a given input.

Deeper integrations into the framework (`EvalSubtask`) will come in a future PR.
NA

----
📚 Documentation preview 📚: https://griptape--1396.org.readthedocs.build//1396/


docs/griptape-framework/engines/src/eval_engines_1.py
```diff
@@ -0,0 +1,22 @@
...
+        EvalEngine.Param.EXPECTED_OUTPUT,
+        EvalEngine.Param.ACTUAL_OUTPUT,
+    ],
+)
+
+score, reason = engine.evaluate(
+    input="If you have a red house made of red bricks, a blue house made of blue bricks, what is a greenhouse made of?",
+    expected_output="Glass",
+    actual_output="Glass",
+)
```
dylanholmes, line 18: Without looking at the implementation (not hypothetical -- I haven't), it looks weird that I need to say which parameters I'm going to pass into a method in the constructor. Why can't this be implied given the parameters that I actually pass into evaluate?

griptape/engines/eval/eval_engine.py
```diff
@@ -0,0 +1,144 @@
...
+        if EvalEngine.Param.INPUT not in value:
+            raise ValueError("Input is required in evaluation_params")
+
+        if EvalEngine.Param.ACTUAL_OUTPUT not in value:
+            raise ValueError("Actual Output is required in evaluation_params")
+
+    def __attrs_post_init__(self) -> None:
+        if self.evaluation_steps is None:
+            with validators.disabled():
+                self.evaluation_steps = self._generate_steps()
```
dylanholmes, line 84: We're invoking an LLM when constructing the eval engine? Not sure how I feel about this.

griptape/templates/engines/eval/results/system.j2
```diff
@@ -0,0 +1,18 @@
...
+{{ text }}
+
+**
+IMPORTANT: Please make sure to only return in JSON format, with the "score" and "reason" key. No words or explanation is needed.
+
+Example JSON:
+{
+    "score": 0,
+    "reason": "The text does not follow the evaluation steps provided."
+}
```
dylanholmes, line 17: Why can't you use the json schema rule?

griptape/engines/eval/eval_engine.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def evaluate(self, **kwargs) -> tuple[float, str]:
+        if not all(key.upper() in EvalEngine.Param.__members__ for key in kwargs):
+            raise ValueError("All keys in kwargs must be a member of EvalEngine.Param")
+
+        if EvalEngine.Param.INPUT.name.lower() not in kwargs:
+            raise ValueError("Input is required for evaluation")
+        if EvalEngine.Param.ACTUAL_OUTPUT.name.lower() not in kwargs:
+            raise ValueError("Actual Output is required for evaluation")
```
vachillo, line 93: agreed with @dylanholmes, seems like we're validating the kwargs here anyway, why the extra enum for the parameters?

griptape/engines/eval/eval_engine.py
vachillo: agreed, this seems like it can be lazily instantiated via a `property` or just directly called during `_generate_results`

docs/griptape-framework/engines/src/eval_engines_1.py
collindutter: I think we can, updated the implementation.

griptape/engines/eval/eval_engine.py
```diff
@@ -0,0 +1,137 @@
...
+            raise ValueError("can't have both evaluation_steps and criteria specified")
+
+        if not value:
+            raise ValueError("evaluation_steps must not be empty")
+
+    def evaluate(self, input: str, actual_output: str, **kwargs) -> tuple[float, str]:  # noqa: A002
+        evaluation_params = {
+            key.replace("_", " ").title(): value
+            for key, value in {"input": input, "actual_output": actual_output, **kwargs}.items()
+        }
```
dylanholmes, line 79: Why do you need to transform the key this way? Why pass kwargs as params? (don't see it used in test or example)

griptape/engines/eval/eval_engine.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if self.evaluation_steps is None:
+            with validators.disabled():
```
dylanholmes, line 82: What does`validators.disabled()` do?

griptape/engines/eval/eval_engine.py
```diff
@@ -0,0 +1,137 @@
...
+        if not value:
+            raise ValueError("evaluation_steps must not be empty")
+
+    def evaluate(self, input: str, actual_output: str, **kwargs) -> tuple[float, str]:  # noqa: A002
+        evaluation_params = {
+            key.replace("_", " ").title(): value
+            for key, value in {"input": input, "actual_output": actual_output, **kwargs}.items()
+        }
+
+        if self.evaluation_steps is None:
```
dylanholmes, line 81: Should it be an error to pass both evaluation steps and criteria?

dylanholmes: What if you just passed in one str or list of strings and then had another option to upsample? (Or even a separate utility that could upsample outside of this class?)

griptape/engines/eval/eval_engine.py
```diff
@@ -0,0 +1,137 @@
...
+        result = self.prompt_driver.run(
+            PromptStack(
+                messages=[
+                    Message(system_prompt, role=Message.SYSTEM_ROLE),
+                    Message(user_prompt, role=Message.USER_ROLE),
+                ],
+            ),
+        ).to_artifact()
+
+        parsed_result = json.loads(result.value)
```
dylanholmes, line 104: Why do you need to do `to_artifact()` then `.value`? Can't you just do `.value` or `to_text()` on the returned `Message`?

griptape/engines/eval/eval_engine.py
```diff
@@ -0,0 +1,137 @@
...
+
+        score = float(parsed_result["score"]) / 10
```
dylanholmes, line 128: Why `/ 10`?

griptape/engines/eval/eval_engine.py
```diff
@@ -0,0 +1,137 @@
...
+        reason = parsed_result["reason"]
+
+        return score, reason
+
+    def __generate_evaluation_params_text(self, evaluation_params: dict[str, str]) -> str:
+        return ", ".join(param for param in evaluation_params)
```
dylanholmes, line 134: Same as `", ".join(evalutation_params.keys())`? Almost feel like it'd be more readable to inline even if adds duplication since you really need to know its implementation to know whether or not you can reuse it (and it's really short anyway).

griptape/templates/engines/eval/results/system.j2
```diff
@@ -0,0 +1,10 @@
+Given the evaluation steps, return a JSON with two keys:
+1) a `score` key ranging from 0 - 10, with 10 being that it follows the criteria outlined in the steps and 0 being that it does not.
```
dylanholmes, line 2: Why not `0.0` - `1.0`? and avoid the division afterwords? (which is temporally coupled to this specific prompt template)

griptape/engines/eval/eval_engine.py
collindutter: Disables validators 😄

Will add a comment explaining why.

griptape/engines/eval/eval_engine.py
collindutter: Probably don't need to, just wanted to make it look less pythonic to the LLM. As if a user was explaining it in plain english.

griptape/engines/eval/eval_engine.py
collindutter: > https://github.com/griptape-ai/griptape/blob/441f55311c58e6ebe228a72638bf7480420253d2/griptape/engines/eval/eval_engine.py#L56-L57

Yes, and we check for this.

> What if you just passed in one str or list of strings and then had another option to upsample?

Could you please expand a bit more on this?

griptape/engines/eval/eval_engine.py
collindutter: `to_text()` works, will update

griptape/engines/eval/eval_engine.py
collindutter: I'd rather the LLM not have to deal with floats at all. Seems like there could be more ambiguity around level of precision. I can add a comment explaining this.

griptape/templates/engines/eval/results/system.j2
collindutter: Answer [here](https://github.com/griptape-ai/griptape/pull/1396#discussion_r1889012269)

griptape/engines/eval/eval_engine.py
dylanholmes: Like only have one param for text input, then a flag for whether or not to upsample that text automatically.

dylanholmes (approved):
Expanded on that comment but it's optional

griptape/engines/eval/eval_engine.py
collindutter: After giving this more thought, I believe `evaluation_steps` are fundamentally different from upsampled `criteria`. And writing a single criterion feels easier than writing a single evaluation step and then having it upsampled into 3-4. My preference is to release this feature as-is and gather feedback. If users find having both fields confusing, we can revisit and simplify in Griptape 2.0.

PR #1404, opened 2024-12-05
Add branch task

### LINKED ISSUE #904: Add support for branching in `Workflow`s

collindutter:

### Added
- `BranchTask` for selecting which Tasks (if any) to run based on a condition.
- Support for `BranchTask` in `StructureVisualizer`.

Closes #904

### DIFF
--- a/griptape/tasks/base_task.py
+++ b/griptape/tasks/base_task.py
@@ -28,6 +28,7 @@ class BaseTask(FuturesExecutorMixin, SerializableMixin, RunnableMixin["BaseTask"
         PENDING = 1
         RUNNING = 2
         FINISHED = 3
+        SKIPPED = 4
 
     id: str = field(default=Factory(lambda: uuid.uuid4().hex), kw_only=True, metadata={"serializable": True})
     state: State = field(default=State.PENDING, kw_only=True, metadata={"serializable": True})
@@ -136,6 +137,9 @@ class BaseTask(FuturesExecutorMixin, SerializableMixin, RunnableMixin["BaseTask"
     def is_running(self) -> bool:
         return self.state == BaseTask.State.RUNNING
 
+    def is_skipped(self) -> bool:
+        return self.state == BaseTask.State.SKIPPED
+
     def before_run(self) -> None:
         super().before_run()
         if self.structure is not None:
@@ -181,7 +185,19 @@ class BaseTask(FuturesExecutorMixin, SerializableMixin, RunnableMixin["BaseTask"
             )
 
     def can_run(self) -> bool:
-        return self.state == BaseTask.State.PENDING and all(parent.is_finished() for parent in self.parents)
+        # If this Task has been skipped or is not pending, it should not run
+        if self.is_skipped() or not self.is_pending():
+            return False
+
+        # If this Task has parents, and _all_ of them are skipped, it should not run
+        if self.parents and all(parent.is_skipped() for parent in self.parents):
+            self.state = BaseTask.State.SKIPPED
+            return False
+
+        # If _all_ this Task's unskipped parents are finished, it should run
+        unskipped_parents = [parent for parent in self.parents if not parent.is_skipped()]
+
+        return all(parent.is_finished() for parent in unskipped_parents)
 
     def reset(self) -> BaseTask:
         self.state = BaseTask.State.PENDING
--- a/griptape/tasks/branch_task.py
+++ b/griptape/tasks/branch_task.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+from typing import Callable, Union
+
+from attrs import define, field
+
+from griptape.artifacts import InfoArtifact, ListArtifact
+from griptape.tasks import BaseTask, CodeExecutionTask
+
+
+@define
+class BranchTask(CodeExecutionTask):
+    on_run: Callable[[BranchTask], Union[InfoArtifact, ListArtifact[InfoArtifact]]] = field(kw_only=True)
+
+    def try_run(self) -> InfoArtifact | ListArtifact[InfoArtifact]:
+        result = self.on_run(self)
+
+        if isinstance(result, ListArtifact):
+            branch_task_ids = {artifact.value for artifact in result}
+        else:
+            branch_task_ids = {result.value}
+
+        if not all(branch_task_id in self.child_ids for branch_task_id in branch_task_ids):
+            raise ValueError(f"Branch task returned invalid child task id {branch_task_ids}")
+
+        if self.structure is not None:
+            children_to_skip = [child for child in self.children if child.id not in branch_task_ids]
+            for child in children_to_skip:
+                child.state = BaseTask.State.SKIPPED
+
+        return result
--- a/griptape/utils/structure_visualizer.py
+++ b/griptape/utils/structure_visualizer.py
@@ -41,12 +41,15 @@ class StructureVisualizer:
 
     def __render_task(self, task: BaseTask) -> str:
         from griptape.drivers import LocalStructureRunDriver
-        from griptape.tasks import StructureRunTask
+        from griptape.tasks import BranchTask, StructureRunTask
 
         parts = []
         if task.children:
             children = " & ".join([f"{self.build_node_id(child)}" for child in task.children])
-            parts.append(f"{self.build_node_id(task)}--> {children};")
+            if isinstance(task, BranchTask):
+                parts.append(f"{self.build_node_id(task)}{{ {self.build_node_id(task)} }}-.-> {children};")
+            else:
+                parts.append(f"{self.build_node_id(task)}--> {children};")
         else:
             parts.append(f"{self.build_node_id(task)};")
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `PromptTask.conversation_memory` for setting the Conversation Memory on a Prompt Task.
 - `Structure.conversation_memory_strategy` for setting whether Conversation Memory Runs should be created on a per-Structure or per-Task basis. Default is `per_structure`.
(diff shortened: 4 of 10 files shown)


griptape/tasks/base_task.py
```diff
@@ -181,7 +185,18 @@ def after_run(self) -> None:
...
 
     def can_run(self) -> bool:
-        return self.state == BaseTask.State.PENDING and all(parent.is_finished() for parent in self.parents)
+        # If this Task has been skipped or is not pending, it should not run
+        if self.is_skipped() or not self.is_pending():
+            return False
+
+        # If this Task has parents, and _all_ of them are skipped, it should not run
+        if self.parents and all(parent.is_skipped() for parent in self.parents):
+            return False
```
vachillo, line 194: need to set `self.state = BaseTask.State.SKIPPED` here?

griptape/tasks/branch_task.py
```diff
@@ -0,0 +1,30 @@
...
+from attrs import define, field
+
+from griptape.artifacts import InfoArtifact, ListArtifact
+from griptape.tasks import BaseTask, CodeExecutionTask
+
+
+@define
+class BranchTask(CodeExecutionTask):
+    on_run: Callable[[BranchTask], Union[InfoArtifact, ListArtifact[InfoArtifact]]] = field(kw_only=True)
+
```
vachillo, line 14: should we set a final "chosen" task_id on the class?

griptape/tasks/branch_task.py
```diff
@@ -0,0 +1,30 @@
...
+            branch_task_ids = {artifact.value for artifact in result}
+        else:
+            branch_task_ids = {result.value}
+
+        if not all(branch_task_id in self.child_ids for branch_task_id in branch_task_ids):
+            raise ValueError(f"Branch task returned invalid child task id {branch_task_ids}")
+
+        for child_id in self.child_ids:
+            if self.structure and child_id not in branch_task_ids:
+                self.structure.find_task(child_id).state = BaseTask.State.SKIPPED
```
vachillo, line 28: should this traverse the rest of the children to see if they need to be skipped? or will `BaseTask` take care of that?

griptape/tasks/base_task.py
collindutter: Depends what we want the behavior to be. Should a Task of entirely skipped parents be `SKIPPED` or `PENDING`?

I can see the argument for `SKIPPED` but this feels like the wrong place to do it.

vachillo: yeah modifying state here feels odd. might need to traverse down the graph as part of the `BranchTask` then

griptape/tasks/branch_task.py
collindutter: That is the output of the task, are you envisioning something else?

griptape/tasks/base_task.py
```diff
@@ -181,7 +185,18 @@ def after_run(self) -> None:
...
+
+        # If _all_ this Task's unskipped parents are finished, it should run
+        unskipped_parents = [parent for parent in self.parents if not parent.is_skipped()]
```
collindutter, line 197: I'm not sure this should be the default behavior. A Task may rely on the output of a parent and not know that it was skipped.

Airflow defaults to [all_success](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#trigger-rules) which I think we should follow.

What do you think @vachillo?

griptape/tasks/base_task.py
```diff
@@ -181,7 +185,18 @@ def after_run(self) -> None:
...
+
+        return all(parent.is_finished() for parent in unskipped_parents)
```
collindutter, line 199: I'm not sure this should be the default behavior. A Task may rely on the output of a parent and not know that it was skipped.

Airflow defaults to [all_success](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#trigger-rules) which I think we should follow.

What do you think @vachillo?

```suggestion
    def can_run(self) -> bool:
        # If this Task has been skipped or is not pending, it should not run
        if self.is_skipped() or not self.is_pending():
            return False

        # If this Task has parents, and _any_ of them are skipped, it should not run
        if self.parents and any(parent.is_skipped() for parent in self.parents):
            self.state = BaseTask.State.SKIPPED
            return False

        return all(parent.is_finished() for parent in self.parents)
```

collindutter, line 199: I'm not sure this should be the default behavior. A Task may rely on the output of a parent and not know that it was skipped.

Airflow defaults to [all_success](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#trigger-rules) which I think we should follow.

What do you think @vachillo?

```suggestion
        # If this Task has been skipped or is not pending, it should not run
        if self.is_skipped() or not self.is_pending():
            return False

        # If this Task has parents, and _any_ of them are skipped, it should not run
        if self.parents and any(parent.is_skipped() for parent in self.parents):
            self.state = BaseTask.State.SKIPPED
            return False

        return all(parent.is_finished() for parent in self.parents)
```

griptape/tasks/base_task.py
collindutter: I'm leaning towards this being fine. It's so much easier to implement it here than elsewhere.

griptape/tasks/branch_task.py
vachillo: yeah i see now. maybe as a property we can grab the str value of the output? it just seems like you should be able to directly reference it, considering thats the whole point of the task.

collindutter: Agreed, but that seems like an improvement for all Artifacts, not just ones returned from this Task.

vachillo: no just the task has that property I mean

griptape/tasks/base_task.py
vachillo: i can see the argument either way, but to me it would seem weird if a task has two parents, one gets skipped, and then task (and all subsequent tasks) get skipped. 

The best "real use case" i can come up for "Task with 2 parents where one can be skipped" is a workflow where one of the parents is augmenting the task with some optional data. I wouldnt want it to be skipped in that case.

> A Task may rely on the output of a parent and not know that it was skipped.

imo a user advanced enough to write a workflow where this is a possibility would make sure their code could handle it. and if not, I think a better developer experience is a failure rather than skipping.

griptape/tasks/branch_task.py
collindutter: Yeah and I'm proposing that the improvement be made to all Artifacts/Tasks. Feels a bit weird to introduce some syntactic sugar for only one type of Task.

griptape/tasks/base_task.py
collindutter: Having a non-`all_success` default could lead to some really tricky-to-debug behavior where Tasks are receiving partial inputs, but not necessarily failing. With `all_success` as the default, I think it'd be much more obvious that something is wrong since entire Task branches are not running. 

> The best "real use case" i can come up for "Task with 2 parents where one can be skipped" is a workflow where one of the parents is augmenting the task with some optional data. I wouldnt want it to be skipped in that case.

Totally valid use case, I think it would be well suited for a `all_done` trigger rule. 

There are going to be use-cases for all sorts of trigger rules, and ultimately I think we just need to implement more of them so users can choose their behavior. The important thing here is landing on a sensible default for now.

Should I prioritize implementing those before Branch Task is merged?

griptape/tasks/branch_task.py
collindutter: I think this is covered [here](https://github.com/griptape-ai/griptape/pull/1404#discussion_r1873660280), right?

griptape/utils/structure_visualizer.py
```diff
@@ -41,12 +41,15 @@ def __render_tasks(self, tasks: list[BaseTask]) -> str:
...
         from griptape.drivers import LocalStructureRunDriver
-        from griptape.tasks import StructureRunTask
+        from griptape.tasks import BranchTask, StructureRunTask
 
         parts = []
         if task.children:
             children = " & ".join([f"{self.build_node_id(child)}" for child in task.children])
-            parts.append(f"{self.build_node_id(task)}--> {children};")
+            if isinstance(task, BranchTask):
+                parts.append(f"{self.build_node_id(task)}{{ {self.build_node_id(task)} }}--> {children};")
```
vachillo, line 50: can we add the nice [dotted line like this](https://github.com/griptape-ai/griptape/pull/913/files#diff-bebfb804f6ed2a4f064203cffa668c2b0638e75060c386a7a5bfba3f3384c11cR46)?

griptape/tasks/branch_task.py
vachillo: yep should be

griptape/tasks/base_task.py
vachillo: still need to set it?

griptape/utils/structure_visualizer.py
```diff
@@ -41,12 +41,15 @@ def __render_tasks(self, tasks: list[BaseTask]) -> str:
...
         from griptape.drivers import LocalStructureRunDriver
-        from griptape.tasks import StructureRunTask
+        from griptape.tasks import BranchTask, StructureRunTask
 
         parts = []
         if task.children:
             children = " & ".join([f"{self.build_node_id(child)}" for child in task.children])
-            parts.append(f"{self.build_node_id(task)}--> {children};")
+            if isinstance(task, BranchTask):
+                parts.append(f"{self.build_node_id(task)}{{ {self.build_node_id(task)} }}-.-> {children};")
```
vachillo, line 50: actually should we represent it differently depending on if the structure has already run and the tasks have already been chosen?

griptape/tasks/base_task.py
collindutter: Oops

PR #1427, opened 2024-12-12
Change conversation_memory_strategy from enum to string

collindutter:

## Describe your changes
Change `Structure.conversation_memory_strategy` from clunky enum to string literal based on feedback from @shhlife. Both are type safe and new version is easier for users.

I think enums are better suited for internal properties like `BaseTask.state`.
NA

### DIFF
--- a/griptape/structures/structure.py
+++ b/griptape/structures/structure.py
@@ -2,8 +2,7 @@ from __future__ import annotations
 
 import uuid
 from abc import ABC, abstractmethod
-from enum import Enum
-from typing import TYPE_CHECKING, Any, Optional, Union
+from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 
 from attrs import Factory, define, field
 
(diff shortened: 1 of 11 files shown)


griptape/structures/structure.py
```diff
@@ -37,8 +32,8 @@ class ConversationMemoryStrategy(Enum):
         kw_only=True,
         metadata={"serializable": True},
     )
-    conversation_memory_strategy: ConversationMemoryStrategy = field(
-        default=ConversationMemoryStrategy.PER_STRUCTURE, kw_only=True, metadata={"serializable": True}
+    conversation_memory_strategy: Literal["per_structure", "per_task"] = field(
+        default="per_structure", kw_only=True, metadata={"serializable": True}
```
vachillo, line 36: do we need a validator that checks the value is correct?

collindutter: I'd rather rely on the type checker for this.

vachillo (approved):

PR #1428, opened 2024-12-12
Update release branch format

collindutter:

## Describe your changes
Release branches no longer have patch versions (e.g. `release/1.0`, `release/1.1`). 
NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,24 @@
-.PHONY: version
-version: ## Bump version and push to release branch.
-	@poetry version $(v)
+.PHONY: version/patch
+version/patch: ## Bump patch version and push to release branch.
+	@poetry version patch
+	@make version/commit
+	@git push origin release/v$$(poetry version -s | sed 's/\.[0-9]*$$//') # Trim the patch version
+
+.PHONY: version/minor
+version/minor: ## Bump minor version.
+	@poetry version minor
+	@make version/commit
+
+.PHONY: version/major
+version/major: ## Bump major version.
+	@poetry version major
+	@make version/commit
+
+.PHONY: version/commit
+version/commit: ## Commit version bump.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -3,8 +3,8 @@ version: ## Bump version and push to release branch.
 	@poetry version $(v)
 	@git add pyproject.toml
 	@git commit -m "Version bump v$$(poetry version -s)"
-	@git push origin release/v$$(poetry version -s)
-
+	@git push origin release/v$$(poetry version -s | sed 's/\.[0-9]*$$//') # Release branches don't have patch versions.
+
```
vachillo, line 7: i thought we were not doing release branches now, and tagging main directly?

collindutter: The challenge is that we need release branches for patches so that we don't release new functionality early (even if it's feature flagged away).

It's ugly but I think the process looks like this:
Patch releases:
1. Checkout to `release/vX.Y`
1. Cherry pick patch onto `release/X.Y.Z`
1. Run `make version=vX.Y.Z` on `release/X.Y.Z`
1. Run `make publish`

Major/minor releases:
1. Checkout to `main`
1. Run `make version=vX.Y.Z` on `main`
1. Check out to `release/vX.Y` (technically not necessary)
1. Rebase onto `main` (technically not necessary)
1. Run `make publish` (technically not necessary)

Ultimately I'm trying to avoid having to merge release branches back into main to get the version bump. Should we just keep release branches only for patches and then drop the redundant steps for major/minor?

vachillo: yeah we should just tag off main for releases, and bump the version in main to the _next_ unreleased version.

vachillo: for patches, we can branch off the release tag and push fixes there

vachillo (approved):

PR #1434, opened 2024-12-12
Add StructureVisualizer params

### LINKED ISSUE #1420: Add a `theme` parameter to `StructureVisualizer`

Currently StructureVisualizer outputs a light theme by default. it'd be great to be able to pass it a the dark theme. I've been hacking it with:
```python
# Visualize the workflow
url = StructureVisualizer(workflow).to_url()
theme = "dark"
bgColor = "2b2b2b"
full_url = f"{url}?theme={theme}&bgColor={bgColor}"
webbrowser.open(full_url)
```

but it'd be awesome to be able to do:
```python
url = StructureVisualizer(workflow, theme="dark").to_url()
webbrowser.open(url)
```

Note - currenly the `dark` theme doesn't actually do the background, just the nodes - that's why I set the bgColor manually before. Ideally it sets it itself.

collindutter:

### Added
- `StructureVisualizer.base_url` for setting the base URL on the url generated by `StructureVisualizer.to_url()`.
- `StructureVisualizer.query_params` for setting query parameters on the url generated by `StructureVisualizer.to_url()`.
Closes #1420


griptape/utils/structure_visualizer.py
```diff
@@ -34,7 +38,14 @@ def to_url(self) -> str:
         graph_bytes = graph.encode("utf-8")
         base64_string = base64.b64encode(graph_bytes).decode("utf-8")
 
-        return f"https://mermaid.ink/svg/{base64_string}"
+        url = urllib.parse.urljoin(self.base_url, f"svg/{base64_string}")
+        req = PreparedRequest()
+        req.prepare_url(url, self.query_params)
+
+        if req.url is None:
+            raise ValueError("Failed to generate the URL")
```
collindutter, line 46: I can't for the life of me figure out how to trigger this, but `pyright` says it's `str | None` so 🤷

griptape/utils/structure_visualizer.py
```diff
@@ -17,6 +19,8 @@ class StructureVisualizer:
     structure: Structure = field()
     header: str = field(default="graph TD;", kw_only=True)
     build_node_id: Callable[[BaseTask], str] = field(default=lambda task: task.id.title(), kw_only=True)
+    query_params: dict[str, str] = field(factory=dict, kw_only=True)
+    base_url: str = field(default="https://mermaid.ink", kw_only=True)
```
vachillo, line 23: should these just be passed directly to the `to_url()` method? they arent really necessary for the state of the object, only used for generating a url

collindutter: That argument could be made for most of the parameters here, right? If anything I'd say that _structure_ belongs on `to_url()` with the idea being that you'd create one `StructureVisualizer` for many Structures. But the work required to change that in a non-breaking way feels like it outweighs the value of a silly util.

vachillo: eh maybe. there could be other `to_` methods that need the other stuff, but not these. it also seems reasonable to generate different URLs with different params without updating the property every time.

vachillo (approved):
will still leave this here

PR #1238, opened 2024-10-08
Callable Schema

collindutter:
- [x] I have read and agree to the contributing guidelines for [submitting new pull requests](https://github.com/griptape-ai/griptape?tab=readme-ov-file#submitting-pull-requests).

### Added
- Ability to pass callable to `activity.schema` for dynamic schema generation.
## Issue ticket number and link
NA

PR #1443, opened 2024-12-13
Add Structured Output

### LINKED ISSUE #1468: Structured Output

Griptape does not support the various structured output features provided by LLMs. Some examples:
- https://platform.openai.com/docs/guides/structured-outputs
- https://ollama.com/blog/structured-outputs
- https://ai.google.dev/gemini-api/docs/structured-output?lang=python
- https://huggingface.co/learn/cookbook/en/structured_generation#-constrained-decoding

Some mechanism to supply a schema and have it used by these services. Maybe leveraging the existing `JsonSchemaRule`.

Ideally supporting three tiers of structured output support:
- Native
- Tool Based
- System Prompt Based (`JsonSchemaRule` today)

Using `JsonSchemaRule` and being extra polite to the LLM.

collindutter:

### Added
- Structured Output support for all Prompt Drivers.
- `PromptTask.output_schema` for setting an output schema to be used with Structured Output.
- `Agent.output_schema` for setting an output schema to be used on the Agent's Prompt Task.
- `BasePromptDriver.structured_output_strategy` for changing the Structured Output strategy between `native`, `tool`, and `rule`.
Closes #1468


collindutter:
@dylanholmes @vachillo still working through docs and some final cleanup, but the PR is certainly developed/large enough for first pass of reviews.

griptape/tasks/prompt_task.py
```diff
@@ -321,3 +296,60 @@ def _process_task_input(
...
+        schemas = [rule.value for rule in json_schema_rules if isinstance(rule.value, Schema)]
+
+        if len(json_schema_rules) != len(schemas):
+            warnings.warn(
+                "Not all provided `JsonSchemaRule`s include a `schema.Schema` instance. These will be ignored with `use_native_structured_output`.",
+                stacklevel=2,
+            )
+
+        if schemas:
+            stack.output_schema = schemas[0] if len(schemas) == 1 else Schema(Or(*schemas))
```
collindutter, line 318: Really don't like this method. Very open to suggestions on improvements.

griptape/tasks/prompt_task.py
```diff
@@ -92,54 +92,30 @@ def prompt_stack(self) -> PromptStack:
...
+            self._add_native_schema_to_prompt_stack(stack, rulesets)
+
+        # Ensure there is at least one Ruleset that has non-empty `rules`.
+        if any(len(ruleset.rules) for ruleset in rulesets):
+            system_artifacts.append(TextArtifact(J2("rulesets/rulesets.j2").render(rulesets=rulesets)))
+
+        # Ensure there is at least one system Artifact that has a non-empty value.
+        has_system_artifacts = any(system_artifact.value for system_artifact in system_artifacts)
+        if has_system_artifacts:
+            stack.add_system_message(ListArtifact(system_artifacts))
```
collindutter, line 107: By separating out the rendering of Rules from `generate_system_template` it solves #1467 but only in the context of Prompt Task. There are plenty of other places where we use Rules ([RagEngine](https://github.com/griptape-ai/griptape/blob/d07d699bc22e23ff464036972d3a5cfcbb622316/griptape/engines/rag/modules/response/prompt_response_rag_module.py#L59-L60)) that would not benefit from this change.

Open to discussion on how we can solve this globally.

docs/griptape-framework/drivers/prompt-drivers.md
```diff
@@ -25,6 +25,37 @@ You can pass images to the Driver if the model supports it:
...
+## Structured Output
+
+Some LLMs provide functionality often referred to as "Structured Output". This means instructing the LLM to output data in a particular format, usually JSON. This can be useful for forcing the LLM to output in a parsable format that can be used by downstream systems.
+
+Structured output can be enabled or disabled for a Prompt Driver by setting the [use_native_structured_output](../../reference/griptape/drivers.md#griptape.drivers.BasePromptDriver.use_native_structured_output).
+
+If `use_native_structured_output=True`, you can change _how_ the output is structured by setting the [native_structured_output_strategy](../../reference/griptape/drivers.md#griptape.drivers.BasePromptDriver.native_structured_output_strategy) to one of:
+
+- `native`: The Driver will use the LLM's structured output functionality provided by the API.
+- `tool`: Griptape will pass a special Tool, [StructuredOutputTool](../../reference/griptape/tools/structured_output_tool.md) and try to force the LLM to use a Tool.
```
dylanholmes, line 37: In what cases would you use the `tool` strategy? Can it result in multiple "tool" calls/outputs for a single LLM call?

docs/griptape-framework/drivers/prompt-drivers.md
```diff
@@ -25,6 +25,37 @@ You can pass images to the Driver if the model supports it:
 --8<-- "docs/griptape-framework/drivers/src/prompt_drivers_images.py"
 ```
 
+## Structured Output
+
+Some LLMs provide functionality often referred to as "Structured Output". This means instructing the LLM to output data in a particular format, usually JSON. This can be useful for forcing the LLM to output in a parsable format that can be used by downstream systems.
+
+Structured output can be enabled or disabled for a Prompt Driver by setting the [use_native_structured_output](../../reference/griptape/drivers.md#griptape.drivers.BasePromptDriver.use_native_structured_output).
+
+If `use_native_structured_output=True`, you can change _how_ the output is structured by setting the [native_structured_output_strategy](../../reference/griptape/drivers.md#griptape.drivers.BasePromptDriver.native_structured_output_strategy) to one of:
```
dylanholmes, line 34: What is the default strategy? (Maybe mention it here?)

docs/griptape-framework/drivers/src/prompt_drivers_structured_output.py
```diff
@@ -0,0 +1,35 @@
...
+from griptape.structures import Pipeline
+from griptape.tasks import PromptTask
+
+pipeline = Pipeline(
+    tasks=[
+        PromptTask(
+            prompt_driver=OpenAiChatPromptDriver(
+                model="gpt-4o",
+                use_native_structured_output=True,
+                native_structured_output_strategy="native",
```
dylanholmes, line 15: Seems weird to see `use_native_...` then `native...strategy="native"`. Maybe just remove `native` from the param names?

docs/griptape-framework/drivers/src/prompt_drivers_structured_output_multi.py
```diff
@@ -0,0 +1,28 @@
...
+        PromptTask(
+            prompt_driver=OpenAiChatPromptDriver(
+                model="gpt-4o",
+                use_native_structured_output=True,
+                native_structured_output_strategy="tool",
+            ),
+            rules=[
+                JsonSchemaRule(schema.Schema({"color": "red"})),
+                JsonSchemaRule(schema.Schema({"color": "blue"})),
+            ],
```
dylanholmes, line 20: This is really hard for me to imagine what this would do (let alone decide to use it). How do the rules interact with the `tool` native output strategy?

docs/griptape-framework/drivers/prompt-drivers.md
dylanholmes: Actually why do I need to pass the boolean in addition to the strategy name? If I just pass the strategy name is that not sufficient to infer that it should be used? In other words can you just remove the `use_native_structured_output` output flag and rely on `native_structured_output_strategy` exclusively?

docs/griptape-framework/structures/rulesets.md
```diff
@@ -29,6 +29,9 @@ A [Ruleset](../../reference/griptape/rules/ruleset.md) can be used to define [Ru
 [JsonSchemaRule](../../reference/griptape/rules/json_schema_rule.md)s defines a structured format for the LLM's output by providing a JSON schema.
 This is particularly useful when you need the LLM to return well-formed data, such as JSON objects, with specific fields and data types.
 
+If the Prompt Driver supports [Structured Output](../drivers/prompt-drivers.md#structured-output), Griptape will use the schema provided to the `JsonSchemaRule` to ensure JSON output.
+If the Prompt Driver does not support Structured Output, Griptape will include the schema in the system prompt using [this template](https://github.com/griptape-ai/griptape/blob/main/griptape/templates/rules/json_schema.j2).
```
dylanholmes, line 33: I'd suggest not using `Griptape` to like to refer to the thing performing the logic. Instead how about getting more specific about what is responsible? Or if only the effect is important, then maybe use a passive voice, something like `the schema will be used to....`.

docs/griptape-framework/drivers/prompt-drivers.md
collindutter: > In what cases would you use the tool strategy? Can it result in multiple "tool" calls/outputs for a single LLM call?

Really only if `native` is not supported. Anthropic defaults to `tool`, for instance. And users might still value in having multiple strategies even when `native` is supported.

> Actually why do I need to pass the boolean in addition to the strategy name? 

I originally wanted to keep the API similar to native tools. So `use_native_structured_output` would toggle whether `JsonSchemaRule`s went into the system prompt or as the schema for structured output. But now we just have a single `output_schema` field which doesn't have a system-prompt fallback equivalent. So I'm fine with dropping the toggle.

docs/griptape-framework/drivers/src/prompt_drivers_structured_output.py
collindutter: Going to change to `structured_output_strategy` because there is no longer a non-native option as mentioned [here](https://github.com/griptape-ai/griptape/pull/1443/files#r1901336683).

docs/griptape-framework/drivers/src/prompt_drivers_structured_output_multi.py
collindutter: Example has been removed

docs/griptape-framework/drivers/prompt-drivers.md
collindutter: Actually... what do you think about having a `output_schema` get turned into a `JsonSchemaRule` if `use_native_structured_output=False`? This feels more in-line with native tools, and provides a consistent interface for json output even when an LLM does not natively support it.

collindutter: Reverted and added this, let me know what you think

docs/griptape-framework/structures/rulesets.md
collindutter: Removed

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Structure.run_stream()` for streaming Events from a Structure as an iterator.
 - Support for `GenericMessageContent` in `AnthropicPromptDriver` and `AmazonBedrockPromptDriver`.
 - Validators to `Agent` initialization.
+- `BasePromptDriver.use_native_structured_output` for enabling or disabling structured output.
+- `BasePromptDriver.structured_output_strategy` for changing the structured output strategy between `native` and `tool`.
```
dylanholmes, line 25: What happens for when value of `use_native_structured_output` contradicts with the strategy?

Specifically:
1. `use_native_structured_output=False, structured_output_strategy="native"`
2. `use_native_structured_output=True, structured_output_strategy="tool"`

docs/griptape-framework/drivers/prompt-drivers.md
```diff
@@ -25,6 +25,32 @@ You can pass images to the Driver if the model supports it:
...
+## Structured Output
+
+Some LLMs provide functionality often referred to as "Structured Output". This means instructing the LLM to output data in a particular format, usually JSON. This can be useful for forcing the LLM to output in a parsable format that can be used by downstream systems.
+
+Structured output can be enabled or disabled for a Prompt Driver by setting the [use_native_structured_output](../../reference/griptape/drivers.md#griptape.drivers.BasePromptDriver.use_native_structured_output).
+
+If `use_native_structured_output=True`, you can change _how_ the output is structured by setting the [structured_output_strategy](../../reference/griptape/drivers.md#griptape.drivers.BasePromptDriver.structured_output_strategy) to one of:
+
+- `native`: The Driver will use the LLM's structured output functionality provided by the API.
+- `tool`: Griptape will pass a special Tool, [StructuredOutputTool](../../reference/griptape/tools/structured_output/tool.md) and try to force the LLM to use a Tool.
```
dylanholmes, line 37: Including "native" in "use_native_structured_output" doesn't make sense to me if you use the "tool" strategy.

Renaming "use_native_structured_output" to "use_structured_output" would fix this imo.

docs/griptape-framework/drivers/prompt-drivers.md
```diff
@@ -25,6 +25,32 @@ You can pass images to the Driver if the model supports it:
...
+The easiest way to get started with structured output is by using a [PromptTask](../structures/tasks.md#prompt)'s [output_schema](../../reference/griptape/tasks/prompt_task.md#griptape.tasks.PromptTask.output_schema) parameter.
+
+```python
+--8<-- "docs/griptape-framework/drivers/src/prompt_drivers_structured_output.py"
+```
+
+If `use_native_structured_output=False`, the Task will add a [JsonSchemaRule](../structures/rulesets.md#json-schema-rule) to the Task's system prompt.
+
+!!! warning
+    Not every LLM supports `use_native_structured_output` or all `structured_output_strategy` options.
```
dylanholmes, line 52: What is the error behavior if it's not supported? Silently ignoring or an exception?

docs/griptape-framework/structures/rulesets.md
```diff
@@ -26,6 +26,10 @@ A [Ruleset](../../reference/griptape/rules/ruleset.md) can be used to define [Ru
 
 ### Json Schema
 
+!!! tip
+    [Structured Output](../drivers/prompt-drivers.md#structured-output) provides a more robust solution for having the LLM generate structured output.
+    And if an LLM does not natively support structured output, a `JsonSchemaRule` will automatically be added.
```
dylanholmes, line 31: If structured output is enabled and an LLM does not ...., then a Json...

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
```diff
@@ -55,9 +55,20 @@ class AmazonBedrockPromptDriver(BasePromptDriver):
         kw_only=True,
     )
     use_native_tools: bool = field(default=True, kw_only=True, metadata={"serializable": True})
+    use_native_structured_output: bool = field(default=True, kw_only=True, metadata={"serializable": True})
```
dylanholmes, line 58: It doesn't make sense to allow passing `use_native_structured_output=True` to this prompt driver  if it doesn't support native structured output. As I mentioned in another comment, I think this is just a naming problem.

docs/griptape-framework/drivers/prompt-drivers.md
```diff
@@ -25,6 +25,32 @@ You can pass images to the Driver if the model supports it:
 --8<-- "docs/griptape-framework/drivers/src/prompt_drivers_images.py"
 ```
 
+## Structured Output
```
dylanholmes, line 28: I still don't think there should be a strategy and a flag. You can just add one more strategy for the flag is false case. A single param would make for a simpler, less confusing API.

Additionally, I think you could go one step further and move output_schema into a strategy object that is passed in.

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
```diff
@@ -115,14 +125,27 @@ def _base_params(self, prompt_stack: PromptStack) -> dict:
...
+                "toolChoice": self.tool_choice,
+            }
+
+            if (
+                prompt_stack.output_schema is not None
+                and self.use_native_structured_output
+                and self.structured_output_strategy == "tool"
+            ):
+                self._add_structured_output_tool(prompt_stack)
+                params["toolConfig"]["toolChoice"] = {"any": {}}
```
dylanholmes, line 143: Are we overriding the tool choice passed in by the user? Does that mean that tool choice is ignored when using structured output?

griptape/drivers/prompt/base_prompt_driver.py
```diff
@@ -122,6 +126,16 @@ def try_run(self, prompt_stack: PromptStack) -> Message: ...
     @abstractmethod
     def try_stream(self, prompt_stack: PromptStack) -> Iterator[DeltaMessage]: ...
 
+    def _add_structured_output_tool(self, prompt_stack: PromptStack) -> None:
+        from griptape.tools.structured_output.tool import StructuredOutputTool
+
+        if prompt_stack.output_schema is None:
+            raise ValueError("PromptStack must have an output schema to use structured output.")
```
dylanholmes, line 133: If output_schema was on a strategy object, then this check wouldn't even be needed -- it'd be free by-design.

griptape/drivers/prompt/base_prompt_driver.py
```diff
@@ -122,6 +126,16 @@ def try_run(self, prompt_stack: PromptStack) -> Message: ...
     @abstractmethod
     def try_stream(self, prompt_stack: PromptStack) -> Iterator[DeltaMessage]: ...
 
+    def _add_structured_output_tool(self, prompt_stack: PromptStack) -> None:
```
dylanholmes, line 129: I know it'd be verbose, but I'd prefer `_add_structured_output_tool_if_absent`. I had to do a double take when encountering its usage.

griptape/drivers/prompt/huggingface_hub_prompt_driver.py
```diff
@@ -51,11 +55,23 @@ def client(self) -> InferenceClient:
             token=self.api_token,
         )
 
+    @structured_output_strategy.validator  # pyright: ignore[reportAttributeAccessIssue, reportOptionalMemberAccess]
+    def validate_structured_output_strategy(self, attribute: Attribute, value: str) -> str:
+        if value == "tool":
+            raise ValueError("HuggingFaceHubPromptDriver does not support `tool` structured output mode.")
```
dylanholmes, line 61: Is this because it just doesn't work very well?

griptape/drivers/prompt/ollama_prompt_driver.py
```diff
@@ -79,7 +80,7 @@ def try_run(self, prompt_stack: PromptStack) -> Message:
         params = self._base_params(prompt_stack)
         logger.debug(params)
         response = self.client.chat(**params)
-        logger.debug(response)
+        logger.debug(response.model_dump())
```
dylanholmes, line 83: You only need `model_dump()` if the response is structure output (e.g. a pydantic model), right? Should there be a conditional here?

CHANGELOG.md
collindutter: If `use_native_structured_output=False`, `native_structured_output_strategy` will have no effect.

docs/griptape-framework/drivers/prompt-drivers.md
collindutter: That's a good point. Will remove "native".

docs/griptape-framework/drivers/prompt-drivers.md
collindutter: There are [validators](https://github.com/griptape-ai/griptape/blob/541fc543aec165100650467c7695c899e791e65b/griptape/drivers/prompt/anthropic_prompt_driver.py?plain=1#L82-L87) on the Prompt Drivers that do not support certain strategies.

docs/griptape-framework/structures/rulesets.md
collindutter: This will only happen if `use_structured_output=False` and `output_schema` is provided. I will still reword for clarity.

collindutter: Actually, going to remove. There is already a [blurb](https://github.com/griptape-ai/griptape/blob/541fc543aec165100650467c7695c899e791e65b/docs/griptape-framework/drivers/prompt-drivers.md?plain=1#L49) on this behavior on the prompt drivers page.

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
collindutter: Fixed naming

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
collindutter: Yes

griptape/drivers/prompt/huggingface_hub_prompt_driver.py
collindutter: Because hugging face does not support native tools which is basically required for the `tool` structured output strategy.

griptape/drivers/prompt/ollama_prompt_driver.py
collindutter: This change is mostly unrelated to this PR, just an improvement I noticed while debugging.

docs/griptape-framework/drivers/prompt-drivers.md
collindutter: So maybe: `native`, `tool`, and `rule`?

griptape/tools/structured_output/tool.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@define
+class StructuredOutputTool(BaseTool):
+    output_schema: Schema = field(kw_only=True)
+
+    @activity(
+        config={
+            "description": "Used to provide the final response which ends this conversation.",
+            "schema": lambda self: self.output_schema,
```
vachillo, line 16: does this need to be `_self`? [reference](https://github.com/griptape-ai/griptape/pull/1238/files#diff-19e76960e5f1e031b1ce73aa3331cf15527d09ffc6742630c3c72e238334c3b8R59)

griptape/tasks/prompt_task.py
```diff
@@ -8,6 +8,7 @@
 
 from griptape import utils
 from griptape.artifacts import ActionArtifact, BaseArtifact, ErrorArtifact, ListArtifact, TextArtifact
+from griptape.artifacts.json_artifact import JsonArtifact
```
vachillo, line 11: nit: intentional to not add to top-level import?

griptape/common/prompt_stack/prompt_stack.py
```diff
@@ -24,13 +24,16 @@
...
+    from schema import Schema
+
     from griptape.tools import BaseTool
 
 
 @define
 class PromptStack(SerializableMixin):
     messages: list[Message] = field(factory=list, kw_only=True, metadata={"serializable": True})
     tools: list[BaseTool] = field(factory=list, kw_only=True)
+    output_schema: Optional[Schema] = field(default=None, kw_only=True)
```
vachillo, line 36: is this supposed to be unused in this class??

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
```diff
@@ -3,7 +3,7 @@
 import logging
 from typing import TYPE_CHECKING, Any
 
-from attrs import Factory, define, field
+from attrs import Attribute, Factory, define, field
```
vachillo, line 6: nit: looks like this should be under TYPE_CHECKING, but we dont do that anywhere else

griptape/drivers/prompt/anthropic_prompt_driver.py
```diff
@@ -110,23 +121,28 @@ def _base_params(self, prompt_stack: PromptStack) -> dict:
...
         }
 
+        if prompt_stack.tools and self.use_native_tools:
+            params["tool_choice"] = self.tool_choice
+
+            if prompt_stack.output_schema is not None and self.structured_output_strategy == "tool":
+                params["tool_choice"] = {"type": "any"}
+
+            params["tools"] = self.__to_anthropic_tools(prompt_stack.tools)
+
```
vachillo, line 143: i dont know enough about all of the driver-specific implementations here, so i am going to assume all the test cases thoroughly cover these changes.

griptape/drivers/prompt/base_prompt_driver.py
```diff
@@ -122,6 +129,34 @@ def try_run(self, prompt_stack: PromptStack) -> Message: ...
...
+            if self.structured_output_strategy == "tool":
+                structured_output_tool = StructuredOutputTool(output_schema=output_schema)
+                if structured_output_tool not in prompt_stack.tools:
+                    prompt_stack.tools.append(structured_output_tool)
+            elif self.structured_output_strategy == "rule":
+                output_artifact = TextArtifact(JsonSchemaRule(output_schema.json_schema("Output Schema")).to_text())
+                system_messages = prompt_stack.system_messages
+                if system_messages:
+                    last_system_message = prompt_stack.system_messages[-1]
+                    last_system_message.content.extend(
```
vachillo, line 145: if this logic lives on `BasePromptDriver`, why is `output_format` only on `PromptTask`?

griptape/tools/structured_output/tool.py
collindutter: Doesn't need to be, but will update for consistency.

griptape/tasks/prompt_task.py
collindutter: Not intentional, will update.

griptape/common/prompt_stack/prompt_stack.py
collindutter: It's used by the Prompt Drivers, much like `PromptStack.tools`.

I don't love having either of these fields on the Prompt Stack, but it's effectively our way of passing runtime args to a Prompt Driver.

griptape/drivers/prompt/amazon_bedrock_prompt_driver.py
collindutter: `ruff` is fine with this so I'm fine with this

griptape/drivers/prompt/anthropic_prompt_driver.py
collindutter: Trust

griptape/drivers/prompt/base_prompt_driver.py
collindutter: We can add to more Prompt Driver-based things in the future (`EvalEngine`), but `PromptTask` brings the most immediate value and this PR is already large enough.

griptape/common/prompt_stack/prompt_stack.py
vachillo: is it runtime though? it gets set on init. should `BasePromptDriver` have an `output_schema` property ?

collindutter: That would require creating a new Prompt Driver for each type of structured output. Tasks are intended to be created per use-case, not Drivers.

vachillo: makes sense

vachillo:
offline discussion: support a json schema for `output_schem`, rather than a `schema.Schema` object

dylanholmes (approved):

vachillo (approved):

PR #1451, opened 2024-12-16
Implement serialization in rulesets

collindutter:

### Changed
- Rulesets can now be serialized and deserialized.
NA

### DIFF
--- a/griptape/rules/base_rule.py
+++ b/griptape/rules/base_rule.py
@@ -5,10 +5,12 @@ from typing import Any
 
 from attrs import define, field
 
+from griptape.mixins.serializable_mixin import SerializableMixin
 
-@define(frozen=True)
-class BaseRule(ABC):
-    value: Any = field()
+
+@define()
+class BaseRule(ABC, SerializableMixin):
+    value: Any = field(metadata={"serializable": True})
     meta: dict[str, Any] = field(factory=dict, kw_only=True)
 
     def __str__(self) -> str:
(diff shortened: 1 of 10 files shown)


griptape/rules/base_rule.py
```diff
@@ -5,10 +5,12 @@
 
 from attrs import define, field
 
+from griptape.mixins.serializable_mixin import SerializableMixin
 
-@define(frozen=True)
```
collindutter, line 9: Unfortunate but unavoidable without massive refactors.

dylanholmes: Would love a little more info on the thought process. If the goal is to make `value` and `meta` immutable is there any other ways? Like override the setter?

griptape/rules/rule.py
```diff
@@ -5,9 +5,9 @@
 from griptape.rules import BaseRule
 
 
-@define(frozen=True)
+@define()
```
dylanholmes, line 8: Maybe remove parenthesis for consistency, obviously this is a really nit picky comment.

dylanholmes (approved):
No important comments.

collindutter:
> Would love a little more info on the thought process. If the goal is to make value and meta immutable is there any other ways? Like override the setter?

@dylanholmes The change comes from pyright's note that:
```A frozen class cannot inherit from a class that is not frozen [reportGeneralTypeIssues]```

Making `SerializableMixin` frozen would be a massive (maybe unfeasable) change.

griptape/rules/base_rule.py
collindutter: The change comes from pyright's note that:
```A frozen class cannot inherit from a class that is not frozen [reportGeneralTypeIssues]```

Making `SerializableMixin` frozen would be a massive (maybe unfeasable) change.

PR #1453, opened 2024-12-16
Remove TODO now that attrs is released

collindutter:

## Describe your changes
Remove TODO now that attrs is released.
Relevant https://github.com/griptape-ai/griptape/pull/1371/files#r1880960715

### DIFF
--- a/griptape/tasks/prompt_task.py
+++ b/griptape/tasks/prompt_task.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Callable, Literal, Optional, Union
+from typing import TYPE_CHECKING, Callable, Optional, Union
 
-from attrs import NOTHING, Factory, define, field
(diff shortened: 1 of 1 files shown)


griptape/tasks/prompt_task.py
```diff
@@ -79,7 +79,7 @@ def prompt_stack(self) -> PromptStack:
         if self.output:
             stack.add_assistant_message(self.output)
 
-        if memory is not None:
+        if memory is not None and isinstance(memory, BaseConversationMemory):
```
vachillo, line 82: nit: can this be 
```suggestion
        if isinstance(memory, BaseConversationMemory):
```
or 
```suggestion
        if memory is not None and memory is not NothingType:
```
?

collindutter: While it works, I think the first one kind of hides some behavior. Updated to a variation on the second.

vachillo (approved):

PR #1457, opened 2024-12-17
Support inserting conversation memory on tasks without structures

collindutter:

### Changed
- Support inserting conversation memory on tasks without structures
NA

### DIFF
--- a/griptape/tasks/prompt_task.py
+++ b/griptape/tasks/prompt_task.py
@@ -104,11 +104,9 @@ class PromptTask(RuleMixin, BaseTask):
             self.id,
             self.output.to_text() if self.output is not None else "",
         )
-        structure = self.structure
         conversation_memory = self.conversation_memory
         if (
-            structure is not None
-            and structure.conversation_memory_strategy == "per_task"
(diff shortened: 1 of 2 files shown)


griptape/tasks/prompt_task.py
```diff
@@ -107,8 +107,7 @@ def after_run(self) -> None:
         structure = self.structure
         conversation_memory = self.conversation_memory
         if (
-            structure is not None
-            and structure.conversation_memory_strategy == "per_task"
+            (structure is None or (structure is not None and structure.conversation_memory_strategy == "per_task"))
```
dylanholmes, line 110: Should be equivalent to:
```py
 (structure is None or structure.conversation_memory_strategy == "per_task")
 ```

If `structure is None` is true, then the rhs of the `or` won't even be checked.

dylanholmes (changes requested):

griptape/tasks/prompt_task.py
collindutter: Great call

dylanholmes (approved):
🐑

PR #1463, opened 2024-12-19
Deprecate ToolkitTask, add functionality to PromptTask

### LINKED ISSUE #1464: Merge PromptTask, ToolkitTask, and ToolTask
- [x] I have read and agree to the [[contributing guidelines](https://github.com/griptape-ai/griptape#contributing)](https://github.com/griptape-ai/griptape#contributing).

Users find it unclear when to use each of the three types of Prompt Tasks. For example, `ToolkitTask` can handle everything that `PromptTask` can, making it difficult to justify when to use `PromptTask`. Additionally, `ToolTask` returns tool results directly to the user, while `ToolkitTask` does not, further adding to the confusion.

Simplify the framework by merging all three task types into a single `PromptTask`. The `PromptTask` could include a configurable field to toggle between different tool-calling strategies, such as:

- **Direct Tool Results**: Tool results are returned directly to the user (current `ToolTask` behavior).  
- **LLM-Mediated Results**: Tool results are fed into the LLM until it generates a final response, which is then returned to the user (current `ToolkitTask` behavior).  
- **Combined Results**: Tool results are fed into the LLM until it generates a response, after which both the response and the tool results are returned to the user (currently unsupported).

One alternative is abstracting the logic into different types of Prompt Engines that users can swap between. However, this approach only adds to the complexity. 

Prompt Drivers are responsible for tools, it's reasonable to assume that Prompt Task can be too.

collindutter:

### Deprecated
- `ToolkitTask`. `PromptTask` is a drop-in replacement. 
Closes #1464


dylanholmes (approved):
🌈

Issue #1475, opened 2024-12-20
Document Artifact


collindutter:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
Some LLMs have the ability to upload documents as part of the request. 
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentBlock.html
- TODO: Add others (Anthropic?)

**Describe the solution you'd like**
A `DocumentArtifact` that allows me to pass a document to an LLM.

**Describe alternatives you've considered**
Passing a `GenericArtifact`.

PR #1477, opened 2024-12-20
Add support for GenericMessageContent to AmazonBedrockPromptDriver

collindutter:

### Added
- Support for `GenericMessageContent` in `AnthropicPromptDriver` and `AmazonBedrockPromptDriver`.
Relevant #1475

### DIFF
--- a/docs/examples/src/talk_to_a_document.py
+++ b/docs/examples/src/talk_to_a_document.py
@@ -1,0 +1,51 @@
+import base64
+import logging
+
+import requests
+
+from griptape.artifacts import GenericArtifact, TextArtifact
+from griptape.configs import Defaults
+from griptape.configs.logging import TruncateLoggingFilter
+from griptape.drivers import AnthropicPromptDriver
+from griptape.structures import Agent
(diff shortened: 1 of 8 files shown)


docs/examples/src/talk_to_a_document.py
```diff
@@ -0,0 +1,42 @@
...
+from griptape.artifacts import GenericArtifact, TextArtifact
+from griptape.drivers import AnthropicPromptDriver
+from griptape.structures import Agent
+
+
+@define
+class DocumentArtifact(GenericArtifact):
+    """Artifact for storing a document.
+
+    Subclassing `GenericArtifact` to avoid printing out `self.value` as `GenericArtifact` does.
```
dylanholmes, line 15: Feels a little weird to me that this is subclassing is needed for this reason only. Doesn't this mean that everyone using the feature will end up creating this subclass?

collindutter: Yeah I don't love it. Let me see if I can provide a more elegant solution to log truncating.

dylanholmes (approved):

PR #1480, opened 2024-12-20
Add events util

### LINKED ISSUE #1479: Explore More Intuitive Event Mechanism

Users find the callback-based nature of `EventBus` confusing. `Stream` provides a nice alternative, but only works for certain events.

A general purpose eventing solution that is as easy to use as `Stream`. Implementation can mostly be the same as `Stream` so I think most of the discussion is around what to name it and where it should live (util, on `Structure`, etc)

collindutter:

### Added
- `Structure.run_stream()` for streaming Events from a Structure as an iterator.

Closes #1479


docs/griptape-framework/misc/events.md
```diff
@@ -74,6 +74,18 @@ Handler 1 <class 'griptape.events.finish_structure_run_event.FinishStructureRunE
 Handler 2 <class 'griptape.events.finish_structure_run_event.FinishStructureRunEvent'>
 ```
 
+## Stream Iterator
+
+You can use `Structure.run_stream()` for streaming Events from the `Structure` in the form of an iterator.
```
dylanholmes, line 79: Would prefer a `EventStream` util like `Stream` that currently exists and punt more core API changes to the Structure api to a later time, but not going to block the PR based on that.

dylanholmes (approved):

docs/griptape-framework/misc/events.md
```diff
@@ -74,6 +74,18 @@ Handler 1 <class 'griptape.events.finish_structure_run_event.FinishStructureRunE
...
+
+!!! tip
+
+Set `stream=True` on your [Prompt Driver](../drivers/prompt-drivers.md) in order to receive completion chunk events.
```
dylanholmes, line 83: I would love to see this requirement be inferred based on Structure usage rather than needing it to be set in the constructor. Obviously out of scope, but writing it down until I have time to fill out an issue.

docs/griptape-framework/misc/events.md
collindutter: The downside of this approach is that `EventStream` would need to support anything that might need this functionality (Structures, Tasks, Drivers, Tools, etc).

I'm just as nervous about modifying `Structure`'s API, but this implementation does at least seem consistent with other similar frameworks.

docs/griptape-framework/misc/events.md
collindutter: Agreed, we need a better way for framework components to broadcast that they publish stream-like events.

PR #1494, opened 2024-12-30
Implement query_vector() for all vector_store_drivers

cjkindel:

## Describe your changes
Add a query_vector() function to all vector_store_driver classes. This facilitates querying with existing vectors instead of generating vectors at query time.

Closes https://github.com/griptape-ai/griptape/issues/1498


collindutter:
@cjkindel can you please create an issue with the original context of why this PR is being made?

griptape/drivers/vector/local_vector_store_driver.py
```diff
@@ -113,6 +111,18 @@ def query(
...
+        query: str,
+        *,
+        count: Optional[int] = None,
+        namespace: Optional[str] = None,
+        include_vectors: bool = False,
+        **kwargs,
+    ) -> list[BaseVectorStoreDriver.Entry]:
+        vector = self.embedding_driver.embed_string(query)
+        return self.query_vector(vector, count=count, namespace=namespace, include_vectors=include_vectors, **kwargs)
+
```
collindutter, line 125: Can we move this implementation up to the base class since it appears to be common across most implementations?

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -139,6 +139,17 @@ def load_entry(self, vector_id: str, *, namespace: Optional[str] = None) -> Opti
...
+    def query_vector(
+        self,
+        vector: list[float],
+        *,
+        count: Optional[int] = None,
+        namespace: Optional[str] = None,
+        include_vectors: bool = False,
+        **kwargs,
+    ) -> list[Entry]: ...
+
```
collindutter, line 152: Adding abstract method is technically a breaking change for any users who have created their own vector store driver.

What do you think about:
```suggestion
    def query_vector(
        self,
        vector: list[float],
        *,
        count: Optional[int] = None,
        namespace: Optional[str] = None,
        include_vectors: bool = False,
        **kwargs,
    ) -> list[Entry]:
      # TODO: Mark as abstract method for griptape 2.0
      raise NotImplementedError(f"{self.__class__.__name__} does not support vector query.")
````

File: griptape/drivers/vector/local_vector_store_driver.py (hunk 1 above)
Comment (cjkindel) [reply griptape/drivers/vector/local_vector_store_driver.py]: Done, left the classes that implement additional kwargs so that documentation will be complete.

File: docs/griptape-framework/misc/events.md [hunk 3]
```diff
@@ -79,7 +79,8 @@ Handler 2 <class 'griptape.events.finish_structure_run_event.FinishStructureRunE
 You can use `Structure.run_stream()` for streaming Events from the `Structure` in the form of an iterator.
 
 !!! tip
-  Set `stream=True` on your [Prompt Driver](../drivers/prompt-drivers.md) in order to receive completion chunk events.
+
+Set `stream=True` on your [Prompt Driver](../drivers/prompt-drivers.md) in order to receive completion chunk events.
```
cjkindel, line 83: I made sure I had latest formatter with `make install` and it still made this change so I think it is valid

collindutter: I've been battling this guy for weeks now. Let me try to fix on main before we merge this.

collindutter: Ah, can you add two tabs before `Set`? That should satisfy the formatter.

griptape/drivers/vector/qdrant_vector_store_driver.py
```diff
@@ -105,18 +105,16 @@ def query(
         """Query the Qdrant collection based on a query vector.
 
         Parameters:
-            query (str): Query string.
+            query (list[float]): Query vector.
```
collindutter, line 108: Docs failing due to missing `vector` in docstring.

collindutter (approved):
Looks good, can you please add a CHANGELOG entry before you merge?

PR #1577, opened 2025-01-14
Fix chunker dropping suffixes

### LINKED ISSUE #1544: Chunked text removing the separator character at end of chunk
When chunking text via a separator, the separator isn't included in the chunks.

```python
from griptape.chunkers import ChunkSeparator, TextChunker
from griptape.tokenizers import OpenAiTokenizer
from rich import print

text = "This is a paragraph of text. I'll count to three: one, two three."
tokenizer = OpenAiTokenizer(model="gpt-4o-mini")
max_tokens = 10
chunker = TextChunker(
    max_tokens=max_tokens, separators=[ChunkSeparator(".")], tokenizer=tokenizer
).chunk(text)

print(f"Max Tokens: {max_tokens}")
for i, chunk in enumerate(chunker):
    chunk_size = tokenizer.count_tokens(chunk.value)
    print(f"{chunk_size}: {chunk.value}")
```

This results in:

```bash
Max Tokens: 10
7: This is a paragraph of text.
9: I'll count to three: one, two three
```

But notice the last `.` is not included after the `three`.

collindutter:

### Fixed
- Chunker occasionally dropping suffix chunk separators.

Whew that was painful. The issue boiled down to `subchunks = list(filter(None, chunk.split(separator.value)))` not re-assembling nicely during `separator.value.join(subchunks)` because we've filtered out some information. However for _very specific_ texts, it is important that we filter out the empty subchunks before proceeding.
Closes #1544

### DIFF
--- a/tests/unit/chunkers/test_text_chunker.py
+++ b/tests/unit/chunkers/test_text_chunker.py
@@ -114,8 +114,8 @@ class TestTextChunker:
             TextChunker(max_tokens=-1)
 
     def test_tiny_max_tokens(self, chunker):
-        max_tokens = 10
-        chunker.separators = [ChunkSeparator(" ")]
+        max_tokens = 11
+        chunker.separators = [ChunkSeparator(".")]
         chunker.max_tokens = max_tokens
         text = "This is a paragraph of text. I'll count to three: one, two three."
         chunks = chunker.chunk(text)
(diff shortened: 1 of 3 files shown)


tests/unit/chunkers/test_text_chunker.py
```diff
@@ -114,15 +114,16 @@ def test_chunk_with_max_tokens(self, chunker):
             TextChunker(max_tokens=-1)
 
     def test_tiny_max_tokens(self, chunker):
-        max_tokens = 10
```
collindutter, line 117: Since the extra period is no longer being dropped we need to use one more token.

vachillo (approved):
im going to assume you now know way too much about this, so ill just approve. nice work

PR #1611, opened 2025-01-24
Add new DatetimeTool activities

### LINKED ISSUE #1609: DateTimeTool Relative Dates Don't Work Well
[DateTimeTool.get_relative_datetime](https://github.com/griptape-ai/griptape/blob/ddb40a941d1ba26e878bf8a29686f5ce0d47630d/griptape/tools/date_time/tool.py#L34) is very finicky. [dateparser](https://dateparser.readthedocs.io/en/latest/), the library we use, is not very clear about what wording is supported. It's not uncommon for the LLM to pass a value like `Date in 3 days` which the library fails to parse.

collindutter:

### Added
- `DateTimeTool.add_timedelta` and `DateTimeTool.get_datetime_diff` for basic datetime arithmetic.

### Changed

- Added `DateTime.get_relative_datetime` to `DateTimeTool.denylist`. May be removed in a future release.

Closes #1609 

----
📚 Documentation preview 📚: https://griptape--1611.org.readthedocs.build//1611/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,10 +23,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+
 - Provider-specific Driver namespaces (e.g., `griptape.drivers.prompt.openai`, `griptape.drivers.embedding.cohere`).
 
-## Added
+### Added
+
 - Tool streaming support to `OllamaPromptDriver`.
+- `DateTimeTool.add_timedelta` and `DateTimeTool.get_datetime_diff` for basic datetime arithmetic.
+
+### Changed
+
+- Added `DateTime.get_relative_datetime` to `DateTimeTool.denylist`. May be removed in a future release.
+
 ### Deprecated
 
 - `griptape.drivers` namespace. Use provider-specific namespaces instead.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 - Tool streaming support to `OllamaPromptDriver`.
+- `DateTimeTool.add_timedelta` and `DateTimeTool.get_datetime_diff` for basic datetime arithmetic.
+
+## Changed
+
+- Added `DateTime.get_relative_datetime` to `DateTimeTool.denylist`. May be removed in a future release.
+
```
vachillo, line 32: isnt this considering breaking still?

vachillo (approved):

CHANGELOG.md
collindutter: I don't think so. The python interface hasn't changed in a breaking way, and LLM be able to solve more Tasks now.

vachillo: hm. the interface hasn't changed. but the default values and the behavior is kind of changed when instantiating an empty `DateTimeTool`. but the LLM probably wouldnt even notice the difference 🤔 .

collindutter: Yeah it's kind of a weird one. If you have strong feelings about I can add it back. But I feel like the behavior will be weirder if there are multiple relative datetime activities.

vachillo: oh yeah for sure it would be weird. it's objectively better so I think it's fine, but interesting situation.

PR #1617, opened 2025-01-24
OpenAi Audio Inputs/Outputs

### LINKED ISSUE #1601: Support OpenAi Audio Inputs/Outputs
Support OpenAi's [audio input/output capabilities](https://platform.openai.com/docs/guides/audio).

collindutter:

Closes #1601


collindutter:
Still need docs, code is ready for review tough.

griptape/drivers/prompt/base_prompt_driver.py
```diff
@@ -177,6 +180,8 @@ def __process_stream(self, prompt_stack: PromptStack) -> Message:
                     delta_contents[content.index] = [content]
                 if isinstance(content, TextDeltaMessageContent):
                     EventBus.publish_event(TextChunkEvent(token=content.text, index=content.index))
+                elif isinstance(content, AudioDeltaMessageContent) and content.data is not None:
```
vachillo, line 183: what would be the scenario where data is `None`?

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -159,6 +170,8 @@ def _base_params(self, prompt_stack: PromptStack) -> dict:
             "seed": self.seed,
             **({"reasoning_effort": self.reasoning_effort} if self.is_reasoning_model else {}),
             **({"temperature": self.temperature} if not self.is_reasoning_model else {}),
+            "modalities": self.modalities,
+            "audio": self.audio,
```
vachillo, line 174: should these follow the same `**(...)` syntax?

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -204,44 +217,47 @@ def __to_openai_messages(self, messages: list[Message]) -> list[dict]:
...
+
+                for content in message.content:
+                    if isinstance(content, ActionCallMessageContent):
+                        if "tool_calls" not in openai_message:
+                            openai_message["tool_calls"] = []
+                        openai_message["tool_calls"].append(self.__to_openai_message_content(content))
+                    elif (
+                        isinstance(content, AudioMessageContent)
+                        and message.is_assistant()
+                        and time.time() < content.artifact.meta.get("expires_at", float("inf"))
```
vachillo, line 249: what is the scenario where there is no `expires_at`? are we just assuming no `expires_at` == try it anyway and see if its expired or not?

griptape/drivers/prompt/openai_chat_prompt_driver.py
```diff
@@ -282,6 +298,23 @@ def __to_openai_message_content(self, content: BaseMessageContent) -> str | dict
                 "type": "image_url",
                 "image_url": {"url": f"data:{content.artifact.mime_type};base64,{content.artifact.base64}"},
             }
+        elif isinstance(content, AudioMessageContent):
+            artifact = content.artifact
+
+            # We can't send the audio if it's expired.
+            if int(time.time()) > artifact.meta.get("expires_at", float("inf")):
```
vachillo, line 305: opposite logic here. wondering if we should just always try to send the transcript if there is no `expires_at`. i also probably dont understand when certain fields are set or unset, and if they are allowed to be None or not

griptape/tasks/actions_subtask.py
```diff
@@ -257,7 +265,7 @@ def __init_from_artifact(self, artifact: TextArtifact | ListArtifact) -> None:
             None
         """
         # When using native tools, we can assume that a TextArtifact is the LLM providing its final answer.
```
vachillo, line 267: update comment

griptape/drivers/prompt/base_prompt_driver.py
collindutter: OpenAi sometimes sends chunks with no content. For example, a final chunk that contains the usage.

griptape/drivers/prompt/openai_chat_prompt_driver.py
collindutter: They get ignored when using non-audio models. But I can see openai-compatible endpoints throwing a fit. I'll make conditional.

griptape/drivers/prompt/openai_chat_prompt_driver.py
collindutter: This logic is hard to follow. I refactored and added comments. Let me know if it's still unclear.

griptape/drivers/prompt/openai_chat_prompt_driver.py
collindutter: Same as above, hopefully cleared it up via refactor/comments.

griptape/drivers/prompt/base_prompt_driver.py
vachillo: and we dont care about the usage here because `AudioChunkEvent` is more concerned about the content?

griptape/drivers/prompt/openai_chat_prompt_driver.py
vachillo: yep makes sense, thanks

vachillo (approved):

griptape/drivers/prompt/base_prompt_driver.py
collindutter: Correct, it's just about the audio data. I'm now realizing there is gap in functionality of streaming publishing streaming events with the transcript chunks, but we can add that later as needed.

PR #1626, opened 2025-01-29
Refactor ActionsSubtask for more precise initialization logic

collindutter:

### Fixed
- `Answer:` being trimmed from LLM's final answer even when using native tool calling.

This PR changes the logic of `ActionsSubtask` to be more precise about how it parses the input Actions. Previously, it followed the decision tree of:
- if TextArtifact: Parse ReAct prompt generated from Prompt Driver with `use_native_tools=False`.
- else: Parse Artifact generated from Prompt Driver with `use_native_tools=True`.

The problem is that even Prompt Driver with `use_native_tools=True` eventually generates a TextArtifact. Now the decision tree looks like:
- if TextArtifact **and** the TextArtifact has been explicitly marked as a ReAct Prompt: Parse ReAct prompt generated from Prompt Driver with `use_native_tools=False`.
- else: Parse Artifact generated from Prompt Driver with `use_native_tools=True`.

The way ReAct Artifacts are "marked" is via a metadata field, `is_react_prompt`. Technically this change of behavior is a breaking change but I highly doubt anyone is passing inputs into `ActionsSubtask` directly.

NA

### DIFF
--- a/griptape/tasks/actions_subtask.py
+++ b/griptape/tasks/actions_subtask.py
@@ -83,12 +83,14 @@ class ActionsSubtask(BaseTask[Union[ListArtifact, ErrorArtifact]]):
         self._origin_task = parent_task
         self.structure = parent_task.structure
 
+        task_input = self.input
         try:
-            if isinstance(self.input, TextArtifact):
-                self.__init_from_prompt(self.input.to_text())
+            if isinstance(task_input, TextArtifact) and task_input.meta.get("is_react_prompt", False):
+                self.__init_from_prompt(task_input.to_text())
             else:
-                self.__init_from_artifacts(self.input)
+                self.__init_from_artifact(task_input)
 
+            # If StructuredOutputTool was used, treat the input to it as the output of the subtask.
             structured_outputs = [a for a in self.actions if isinstance(a.tool, StructuredOutputTool)]
             if structured_outputs:
                 output_values = [JsonArtifact(a.input["values"]) for a in structured_outputs]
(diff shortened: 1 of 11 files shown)


griptape/tasks/actions_subtask.py
```diff
@@ -33,6 +33,7 @@ class ActionsSubtask(BaseTask[Union[ListArtifact, ErrorArtifact]]):
     thought: Optional[str] = field(default=None, kw_only=True)
     actions: list[ToolAction] = field(factory=list, kw_only=True)
     output: Optional[BaseArtifact] = field(default=None, init=False)
+    use_native_actions: bool = field(default=True, kw_only=True)
```
collindutter, line 36: I don't love this name, suggestions? `actions_parse_strategy`?

griptape/tasks/prompt_task.py
```diff
@@ -184,9 +184,11 @@ def try_run(self) -> ListArtifact | TextArtifact | JsonArtifact | ErrorArtifact:
         if self.response_stop_sequence not in self.prompt_driver.tokenizer.stop_sequences:
             self.prompt_driver.tokenizer.stop_sequences.extend([self.response_stop_sequence])
 
-        result = self.prompt_driver.run(self.prompt_stack)
+        output = self.prompt_driver.run(self.prompt_stack).to_artifact(
+            meta={"is_react_prompt": not self.prompt_driver.use_native_tools}
```
dylanholmes, line 188: Might vote for propagating the same terminology instead of adding a new one (and avoid the not -- imo adds cognitive load):

```py
meta={"use_native_tools": self.prompt_driver.use_native_tools}
```
or
```py
meta={"using_native_tools": self.prompt_driver.use_native_tools}
```

dylanholmes (approved):
Nice work! -- Just one subjective comment

griptape/tasks/prompt_task.py
collindutter: I started with this but switched because I wanted to leave `ActionsSubtask` removed from the concept of native vs non-native tools. With the current approach we just handle the edge-case of TextArtifact ambiguity.  As we discussed this whole area is a mess and will hopefully be refactored away in the future.

PR #1658, opened 2025-02-06
feat(drivers-rerank-local): add LocalRerankDriver

### LINKED ISSUE #1657: Add LocalRerankDriver
### Discussed in https://github.com/orgs/griptape-ai/discussions/1558

<div type='discussions-op-text'>

<sup>Originally posted by **shhlife** January 11, 2025</sup>
Currently there's no way to Rerank using the RagEngine unless you have a Cohere account. It would be very helpful to have a way to run a local RerankDriver. Would it be possible to make one for Ollama? It has two reranking models available..

[linux6200/bge-reranker-v2-m3](https://ollama.com/linux6200/bge-reranker-v2-m3)
[zyw0605688/bge-reranker-v2-m3](https://ollama.com/zyw0605688/bge-reranker-v2-m3)
</div>

collindutter:

### Added
- `LocalRerankDriver` for reranking locally.
- Missing docs for rerank drivers.

Closes #1657


docs/griptape-framework/drivers/rerank-drivers.md
```diff
@@ -0,0 +1,29 @@
+---
+search:
+  boost: 2
+---
+
+## Overview
+
+Rerank Drivers can be used to rerank search results for a particular query. Every Rerank Driver implements the following methods:
+They can be used directly or with a [RagEngine's Rerank Module](../engines/rag-engines/#retrievalrerank-modules).
```
vasinov, line 9: Reword to something like "Every Rerank Driver implements the following methods that can be used directly [...]:".

docs/griptape-framework/drivers/src/local_rerank_driver.py
```diff
@@ -0,0 +1,11 @@
+from griptape.artifacts import TextArtifact
+from griptape.drivers.rerank.local import LocalRerankDriver
+
+rerank_driver = LocalRerankDriver()
+
+artifacts = rerank_driver.run(
+    "What is the capital of France?",
+    [TextArtifact("Hotdog"), TextArtifact("San Francisco"), TextArtifact("Paris")],
```
vasinov, line 8: Move to new lines for consistency with the previous example?

griptape/drivers/rerank/local_rerank_driver.py
```diff
@@ -0,0 +1,45 @@
...
+class LocalRerankDriver(BaseRerankDriver, FuturesExecutorMixin):
+    calculate_relatedness: Callable = field(default=lambda x, y: dot(x, y) / (norm(x) * norm(y)))
+    embedding_driver: BaseEmbeddingDriver = field(
+        kw_only=True, default=Factory(lambda: Defaults.drivers_config.embedding_driver), metadata={"serializable": True}
+    )
+
+    def run(self, query: str, artifacts: list[TextArtifact]) -> list[TextArtifact]:
+        query_embedding = self.embedding_driver.embed_string(query)
+
+        with self.create_futures_executor() as futures_executor:
```
vasinov, line 30: Should we handle the case where artifacts already have embeddings? Can't make `embedding_driver` optional, since we need it to embed the query, but may be we add a `generate_embeddings` flag?

If we do that, we'll need to explain in the docs that the embedding model used for generating artifacts has to be the same as the one in `LocalRerankDriver`.

vasinov (changes requested):

docs/griptape-framework/drivers/src/local_rerank_driver.py
collindutter: That formatting is a `ruff` decision. Will add some more examples to force the newlines.

vasinov (approved):

griptape/drivers/rerank/local_rerank_driver.py
collindutter: Agreed offline that this can come as a future optimization.

PR #1714, opened 2025-02-13
ci: automatically generate doc logs

collindutter:

## Describe your changes
This PR introduces automation to generate log previews in our documentation automatically.

We leverage our existing integration tests, which already execute all code snippets. Now, when a test passes, we copy the logs to `/tmp/logs`. Next, we scan for modified Python files in the `docs` directory and copy the relevant logs into a dedicated documentation log directory. Finally, we update any Markdown files that need to reference these logs with a new content tab format.

These changes are then automatically added to a new PR.

This:
````
```python
--8<-- "docs/griptape-framework/structures/src/tasks_1.py"
```
````
Is converted into:
````
=== "Code"
    ```python
    --8<-- "docs/griptape-framework/structures/src/tasks_1.py"
    ```

=== "Logs"
    ```shell
    --8<-- "docs/griptape-framework/structures/logs/tasks_1.txt"
    ```
````
image (7)
image (8)

NA

----
📚 Documentation preview 📚: https://griptape--1714.org.readthedocs.build//1714/

### DIFF
--- a/.github/workflows/docs-integration-tests.yml
+++ b/.github/workflows/docs-integration-tests.yml
@@ -144,7 +144,24 @@ jobs:
         uses: actions/checkout@v4
       - name: Init environment
         uses: ./.github/actions/init-environment
+      # Integration tests will run on all code snippets, outputting their logs to /tmp/logs
       - name: Run integration tests
-        run: make test/integration
-        env:
-          DOCS_ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files || '' }}
+        run: uv run pytest -n auto tests/integration/test_code_blocks.py
+      # If any code snippets have changed, we want to copy the logs to display in the docs
+      - name: Get Changed Files
+        id: changed-files
+        uses: tj-actions/changed-files@v42
+        with:
+          files: docs/**/*.py
+      - name: Run Snippet Logs 
+        run: uv run python .github/scripts/gen_snippet_logs.py docs/griptape-framework ${{ steps.changed-files.outputs.all_changed_files }}
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v7
+        with:
+          branch: "docs/snippet-logs-update"
+          title: "chore(docs): update code snippets logs"
+          body: "This PR was automatically generated to update the code snippets logs in the documentation."
+          commit-message: "chore(docs): update code snippets logs"
+          base: "main"
+          add-paths: |
+            docs/**/*
--- a/.github/scripts/gen_snippet_logs.py
+++ b/.github/scripts/gen_snippet_logs.py
@@ -1,0 +1,132 @@
+from __future__ import annotations
+
+import logging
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: .github/workflows/docs-integration-tests.yml [hunk 1]
```diff
@@ -2,7 +2,6 @@ name: Docs Integration Tests
 
 on:
   push:
-    branches: [ "main" ]
```
collindutter, line 5: Will remove

Makefile
```diff
@@ -36,7 +36,7 @@ test/unit/coverage:
 
 .PHONY: test/integration
 test/integration:
-	@poetry run pytest -n auto tests/integration/test_code_blocks.py
+	@poetry run pytest -n auto tests/integration/test_code_blocks.py
```
collindutter, line 39: Revert

.github/workflows/docs-integration-tests.yml
```diff
@@ -142,7 +141,24 @@ jobs:
...
-        run: make test/integration
-        env:
-          DOCS_ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files || '' }}
+        run: poetry run pytest -n auto tests/integration/test_code_blocks.py
+      # If any code snippets have changed, we want to copy the logs to display in the docs
+      - name: Get Changed Files
+        id: changed-files
+        uses: tj-actions/changed-files@v42
+        with:
+          files: docs/**/*.py
```
dylanholmes, line 152: This won't necessarily run again if the implementation of the code being demonstrated changes (I mean changes to the framework code). This could lead to out of date examples. Maybe its still better than the current process which is completely up to the PR requester.

If we already run all the examples as tests, does it take much more time to update the logs output while doing that?

If it does for some reason, a mitigation could be to re-run this on all the examples just before a release. 

What do you think?

dylanholmes (commented):
Super cool change 🤩 

Just one comment

.github/workflows/docs-integration-tests.yml
collindutter: The problem with regenerating the logs on every sample is that the logs may change in subtle ways each time. Feels odd to have docs that are shifting around on every push.

> Maybe its still better than the current process which is completely up to the PR requester.

We used to run integration tests on PRs, but it was a nightmare with external contributors. Github does not expose the secrets to fork PRs (for good reasons).

> If it does for some reason, a mitigation could be to re-run this on all the examples just before a release.

We also used to do this, but it was a headache having to fix tons of errors right before a release.


I think the best approach is to have some mechanism to explicitly regenerate the logs. Maybe we always process all python files, but only generate the logs if theres not already a logs section? That way if you wanted to regenerate logs for an example, you just remove the logs portion.

collindutter: @dylanholmes gentle bump on this

dylanholmes (approved):

PR #1743, opened 2025-02-17
feat(tools): allow setting allowlist and denylist on Tools

### LINKED ISSUE #1735: Can't Set Allowlist on DateTimeTool
https://github.com/griptape-ai/griptape/pull/1611 introduced some new activities for the DateTimeTool and `denylist`'d the old one. That breaks an example like this:

```python
from griptape.tools import DateTimeTool

ollama_agent = Agent(
    tools=[
        DateTimeTool(allowlist=["get_current_datetime"]),
    ],
)

ollama_agent.run("What is today's date?")
```

```
Traceback (most recent call last):
  File "/Users/collindutter/Documents/griptape/griptape-playground/src/misc/datetime_mistral.py", line 6, in <module>
    DateTimeTool(allowlist=["get_current_datetime"]),
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<attrs generated init griptape.tools.date_time.tool.DateTimeTool>", line 30, in __init__
    __attr_validator_allowlist(self, __attr_allowlist, self.allowlist)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/collindutter/Documents/griptape/griptape/griptape/mixins/activity_mixin.py", line 35, in validate_allowlist
    raise ValueError("can't have both allowlist and denylist specified")
ValueError: can't have both allowlist and denylist specified
```

Maybe we can remove the restriction on preventing allowlist and denylist? Go for a denied-by-default model?

collindutter:

## Describe your changes
Removes arbitrary restriction preventing setting allowlist and denylist on Tools. Both conditions must be satisfied to have the activity included.
Closes #1735

### DIFF
--- a/griptape/mixins/activity_mixin.py
+++ b/griptape/mixins/activity_mixin.py
@@ -31,9 +31,6 @@ class ActivityMixin:
         if allowlist is None:
             return
 
-        if self.denylist is not None:
-            raise ValueError("can't have both allowlist and denylist specified")
-
         for activity_name in allowlist:
             self._validate_tool_activity(activity_name)
 
@@ -41,9 +38,6 @@ class ActivityMixin:
     def validate_denylist(self, _: Attribute, denylist: Optional[list[str]]) -> None:
         if denylist is None:
             return
-
-        if self.allowlist is not None:
-            raise ValueError("can't have both allowlist and denylist specified")
 
         for activity_name in denylist:
             self._validate_tool_activity(activity_name)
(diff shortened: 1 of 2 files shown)


vachillo (approved):
so the outcome here is:
- if both are set, allowlist is the "source of truth"
- if both are set but there's an activity in both, that activity gets denied
is that right?

Issue #1749, opened 2025-02-19
Resolve 2.0 TODOs


collindutter:
We've added a number of TODOs that require breaking changes that need to be resolved before 2.0. Resolve these right before the release.

PR #1753, opened 2025-02-20
feat: add support for image embeddings

### LINKED ISSUE #1729: Image Embedding Drivers
### Discussed in https://github.com/orgs/griptape-ai/discussions/1528

<div type='discussions-op-text'>

<sup>Originally posted by **collindutter** December 31, 2024</sup>

I can't use image embedding models like [OpenAi's CLIP](https://github.com/openai/CLIP) in Griptape.
I want Embedding Drivers to support images.
- [fastembed](https://github.com/qdrant/fastembed)
- [transformers](https://huggingface.co/openai/clip-vit-base-patch32#use-with-transformers)

**Additional context**
The eventual goal is to use the embeddings with Vector Store Drivers which means I'd maybe need `upsert_image_artifact` and `query_image` methods. `upsert_vector` and [query_vector](https://github.com/griptape-ai/griptape/pull/1494) may be sufficient for now.
</div>

collindutter:

## Describe your changes
Sorry for big-ish PR, lots of overlapping changes here:
1. Add support for embedding images in Embeddings Drivers. Also refactors Embedding Drivers to have a single `embed` method rather than type-specific methods (embed_string, embed_image, etc).
a. Add support for embedding images in Amazon Bedrock Titan Embedding Driver.
b. Add support for embedding images in VoyageAi Embedding Driver.
1. Add support for querying/upserting with Vector Store Drivers. Vector Store Driver must be configured with a multi-modal Embedding Driver. Also refactors Vector Store Drivers to add single `upsert` method rather than type-specific methods (`upsert_text`, `upsert_text_artifact`, etc).
a. Add support to Local Vector Store Driver to save `Entry`s with Image Artifacts into the `persist_file`.

Closes #1729

### DIFF
--- a/griptape/drivers/embedding/base_embedding_driver.py
+++ b/griptape/drivers/embedding/base_embedding_driver.py
@@ -1,17 +1,18 @@
 from __future__ import annotations
 
+import warnings
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Optional
 
 import numpy as np
 from attrs import define, field
 
+from griptape.artifacts import ImageArtifact, TextArtifact
 from griptape.chunkers import BaseChunker, TextChunker
 from griptape.mixins.exponential_backoff_mixin import ExponentialBackoffMixin
 from griptape.mixins.serializable_mixin import SerializableMixin
 
 if TYPE_CHECKING:
-    from griptape.artifacts import TextArtifact
     from griptape.tokenizers import BaseTokenizer
 
 
@@ -32,21 +33,50 @@ class BaseEmbeddingDriver(SerializableMixin, ExponentialBackoffMixin, ABC):
         self.chunker = TextChunker(tokenizer=self.tokenizer) if self.tokenizer else None
 
     def embed_text_artifact(self, artifact: TextArtifact) -> list[float]:
-        return self.embed_string(artifact.to_text())
+        warnings.warn(
+            "`BaseEmbeddingDriver.embed_text_artifact` is deprecated and will be removed in a future release. `BaseEmbeddingDriver.embed` is a drop-in replacement.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.embed(artifact)
 
     def embed_string(self, string: str) -> list[float]:
+        warnings.warn(
+            "`BaseEmbeddingDriver.embed_string` is deprecated and will be removed in a future release. `BaseEmbeddingDriver.embed` is a drop-in replacement.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.embed(string)
+
+    def embed(self, value: str | TextArtifact | ImageArtifact) -> list[float]:
         for attempt in self.retrying():
             with attempt:
-                if self.tokenizer is not None and self.tokenizer.count_tokens(string) > self.tokenizer.max_input_tokens:
-                    return self._embed_long_string(string)
-                else:
-                    return self.try_embed_chunk(string)
-
+                if isinstance(value, str):
+                    if (
+                        self.tokenizer is not None
+                        and self.tokenizer.count_tokens(value) > self.tokenizer.max_input_tokens
+                    ):
+                        return self._embed_long_string(value)
+                    else:
+                        return self.try_embed_chunk(value)
+                elif isinstance(value, TextArtifact):
+                    return self.embed(value.to_text())
+                elif isinstance(value, ImageArtifact):
+                    return self.try_embed_artifact(value)
         else:
             raise RuntimeError("Failed to embed string.")
 
+    def try_embed_artifact(self, artifact: TextArtifact | ImageArtifact) -> list[float]:
+        # TODO: Mark as abstract method for griptape 2.0
+        if isinstance(artifact, TextArtifact):
+            return self.try_embed_chunk(artifact.value)
+        else:
+            raise ValueError(f"{self.__class__.__name__} does not support embedding images.")
+
     @abstractmethod
-    def try_embed_chunk(self, chunk: str) -> list[float]: ...
+    def try_embed_chunk(self, chunk: str) -> list[float]:
+        # TODO: Remove for griptape 2.0, subclasses should implement `try_embed_artifact` instead
+        ...
 
     def _embed_long_string(self, string: str) -> list[float]:
         """Embeds a string that is too long to embed in one go.
@@ -58,7 +88,7 @@ class BaseEmbeddingDriver(SerializableMixin, ExponentialBackoffMixin, ABC):
         embedding_chunks = []
         length_chunks = []
         for chunk in chunks:
-            embedding_chunks.append(self.try_embed_chunk(chunk.value))
+            embedding_chunks.append(self.embed(chunk.value))
             length_chunks.append(len(chunk))
 
         # generate weighted averages
--- a/griptape/drivers/vector/base_vector_store_driver.py
+++ b/griptape/drivers/vector/base_vector_store_driver.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
 import uuid
+import warnings
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, overload
 
 from attrs import define, field
 
 from griptape import utils
-from griptape.artifacts import BaseArtifact, ListArtifact, TextArtifact
+from griptape.artifacts import BaseArtifact, ImageArtifact, ListArtifact, TextArtifact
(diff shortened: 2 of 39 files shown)


griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -40,18 +36,80 @@ def to_artifact(self) -> BaseArtifact:
 
     def upsert_text_artifacts(
         self,
-        artifacts: list[TextArtifact] | dict[str, list[TextArtifact]],
+        artifacts: list[TextArtifact] | dict[str, list[TextArtifact] | list[ImageArtifact]],
```
emjay07, line 39: why do we want to support `ImageArtifact` in a method called `upsert_text_artifacts`? if we do keep it, should the first type also be `list[TextArtifact] | list[ImageArtifact]`

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -40,18 +36,80 @@ def to_artifact(self) -> BaseArtifact:
...
+            "`BaseVectorStoreDriver.upsert_text` is deprecated and will be removed in a future release. `BaseEmbeddingDriver.upsert` is a drop-in replacement.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.upsert(string, namespace=namespace, meta=meta, vector_id=vector_id, **kwargs)
+
+    @overload
+    def upsert_collection(
+        self,
+        artifacts: list[TextArtifact],
```
emjay07, line 86: should we support both text and image here?

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -40,18 +36,80 @@ def to_artifact(self) -> BaseArtifact:
...
+        self,
+        artifacts: dict[str, list[TextArtifact] | list[ImageArtifact]],
+        *,
+        meta: Optional[dict] = None,
+        **kwargs,
+    ) -> dict[str, list[str]]: ...
+
+    def upsert_collection(
+        self,
+        artifacts: list[TextArtifact] | dict[str, list[TextArtifact] | list[ImageArtifact]],
```
emjay07, line 103: same commont on first type, or `list[ImageArtifact]`

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -65,21 +123,23 @@ def upsert_text_artifacts(
...
                             )
                         )
 
                 return utils.execute_futures_list_dict(futures_dict)
 
-    def upsert_text_artifact(
+    def upsert(
         self,
-        artifact: TextArtifact,
+        value: str | TextArtifact | ImageArtifact,
```
emjay07, line 134: since we are allowing users to input `str` types, should we also allow `bytes` if a user has a loaded image already in memory they can upsert that may not be wrapped as an `ImageArtifact`?

griptape/drivers/vector/griptape_cloud_vector_store_driver.py
```diff
@@ -97,6 +98,8 @@ def query(
 
         Performs a query on the Knowledge Base and returns Artifacts with close vector proximity to the query, optionally filtering to only those that match the provided filter(s).
         """
+        if isinstance(query, BaseArtifact):
```
emjay07, line 101: why allow the type, just to raise an error later? what if someone tries to pass a `TextArtifact`? should the type be `ImageArtifact` instead?

griptape/drivers/vector/base_vector_store_driver.py
collindutter: `bytes` may represent something else in the future (audio files, video files, etc) so I'd rather not lock us into bytes == image.

griptape/drivers/vector/griptape_cloud_vector_store_driver.py
collindutter: Yes let's narrow the type to `ImageArtifact`, great call.

griptape/drivers/vector/base_vector_store_driver.py
emjay07: ah, fair point. I suppose we have type checking when creating an `ImageArtifact` that we don't want to duplicate here.

griptape/drivers/vector/marqo_vector_store_driver.py
```diff
@@ -165,9 +165,42 @@ def load_entries(self, *, namespace: Optional[str] = None) -> list[BaseVectorSto
...
+        } | kwargs
+
+        results = self.client.index(self.index).search(**params, context={"tensor": [vector], "weight": 1})
+
+        return self.__process_results(results, include_vectors=include_vectors)
+
     def query(
         self,
-        query: str,
+        query: str | BaseArtifact,
```
emjay07, line 203: nit: these types should be the same as BaseVectorStoreDriver

griptape/drivers/embedding/base_embedding_driver.py
```diff
@@ -32,21 +33,51 @@ def __attrs_post_init__(self) -> None:
...
+                elif isinstance(value, TextArtifact):
+                    artifact_value = value.to_text()
+                    if (
+                        self.tokenizer is not None
+                        and self.tokenizer.count_tokens(artifact_value) > self.tokenizer.max_input_tokens
+                    ):
+                        return self._embed_long_string(artifact_value)
+                    else:
+                        return self.try_embed_chunk(artifact_value)
+                elif isinstance(value, ImageArtifact):
```
emjay07, line 65: related to my other comment, I'd expect this then to call `try_embed_image`, which then determine support or not.

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -154,14 +204,14 @@ def query_vector(
...
-        query: str,
+        query: str | TextArtifact | ImageArtifact,
         *,
         count: Optional[int] = None,
         namespace: Optional[str] = None,
         include_vectors: bool = False,
         **kwargs,
     ) -> list[Entry]:
-        vector = self.embedding_driver.embed_string(query)
+        vector = self.embedding_driver.embed(query)
```
emjay07, line 214: I'm guessing that this call will fail to embed then if the user passes an image and the embedding driver doesn't support images. However, thinking a try catch that surfaces more useful error to the user about why the `query` on a `VectorStoreDriver` is related at all to the `embedding_driver` would be better here.

griptape/drivers/vector/marqo_vector_store_driver.py
```diff
@@ -195,22 +228,7 @@ def query(
         } | kwargs
 
         results = self.client.index(self.index).search(query, **params)
```
emjay07, line 230: For my own edification: Marqo (and others) will only support `ImageArtifacts` as a query param if the underlying embedding model that is configured with the driver supports them, correct? or will these always support images?

griptape/drivers/embedding/base_embedding_driver.py
```diff
@@ -32,21 +33,51 @@ def __attrs_post_init__(self) -> None:
...
+                    return self.try_embed_artifact(value)
         else:
             raise RuntimeError("Failed to embed string.")
 
+    def try_embed_artifact(self, artifact: TextArtifact | ImageArtifact) -> list[float]:
+        # TODO: Mark as abstract method for griptape 2.0
+        if isinstance(artifact, TextArtifact):
+            return self.try_embed_chunk(artifact.value)
+        else:
+            raise ValueError(f"{self.__class__.__name__} does not support embedding images.")
```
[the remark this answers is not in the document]
collindutter: Agreed, but we can't make it abstract (and defer to the child class) without it being a breaking change (if users have created their own Vector Stores). That's why I've added to the TODO which will be resolved in 2.0 [here](https://github.com/griptape-ai/griptape/issues/1749).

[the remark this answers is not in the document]
emjay07: ah, I see that comment now. got it.

griptape/drivers/embedding/base_embedding_driver.py
collindutter: I'd like to move away from having multiple methods that provide the same functionality but on different types of inputs. That's what the `@overload` decorator is for. So `try_embed_text` will go away in griptape 2.0 and I'd like to avoid introducing a `try_embed_image`.

griptape/drivers/vector/marqo_vector_store_driver.py
collindutter: That's correct. Though if you somehow already had the embeddings you could use `BaseVectorStore.query_vector` to skip the query embedding.

Issue #1756, opened 2025-02-21
Properly Handle Serialization of Pydantic Models from structured output


vachillo:
currently, using a pydantic model as a value to `output_schema` returns `GenericArtifact[BaseModel]`. this is an issue for downstream `EventListeners` that try to naively call `to_dict()`, as the `BaseModel` class needs special handling

```
class WordOutput(BaseModel):
    thought: str
    parts: List[str]
    combinations: List[str]

def setup_config():
      event_driver = GriptapeCloudEventListenerDriver(
          api_key=get_listener_api_key()
      )
      
      event_listener = EventListener(
          event_listener_driver=event_driver,
      )
      
      EventBus.add_event_listener(event_listener)

def create_word_agent() -> Agent:
    return Agent(
        output_schema=WordOutput,
    )

setup_config()
agent = create_word_agent()

agent.run("hello")
```
produces
```
                               File                                             
                             "/usr/local/lib/python3.11/site-packages/requests/s
                             essions.py", line 484, in prepare_request          
                                 p.prepare(                                     
                               File                                             
                             "/usr/local/lib/python3.11/site-packages/requests/m
                             odels.py", line 370, in prepare                    
                                 self.prepare_body(data, files, json)           
                               File                                             
                             "/usr/local/lib/python3.11/site-packages/requests/m
                             odels.py", line 510, in prepare_body               
                                 body = complexjson.dumps(json, allow_nan=False)
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                               File                                             
                             "/usr/local/lib/python3.11/json/__init__.py", line 
                             238, in dumps                                      
                                 **kw).encode(obj)                              
                                       ^^^^^^^^^^^                              
                               File "/usr/local/lib/python3.11/json/encoder.py",
                             line 200, in encode                                
                                 chunks = self.iterencode(o, _one_shot=True)    
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^    
                               File "/usr/local/lib/python3.11/json/encoder.py",
                             line 258, in iterencode                            
                                 return _iterencode(o, 0)                       
                                        ^^^^^^^^^^^^^^^^^                       
                               File "/usr/local/lib/python3.11/json/encoder.py",
                             line 180, in default                               
                                 raise TypeError(f'Object of type               
                             {o.__class__.__name__} '                           
                             TypeError: Object of type WordOutput is not JSON   
                             serializable 
[...]

Issue #1587, opened 2025-01-17
Serialization Woes


collindutter:
Issues like [this](https://github.com/griptape-ai/griptape/issues/1583) pop up because we maintain a [horrendous list of imports](https://github.com/griptape-ai/griptape/blob/f8d9bec12e5505243e12ab8b0fda40507432ff38/griptape/schemas/base_schema.py?plain=1#L145-L237) that needs to be updated any time a serializable field's type is exclusively imported in a `TYPE_CHECKING` block. Because they're only imported for the type checker, we get a name error when resolving the types unless we actually import it first.

I _think_ the only solution is to move serialized field's type's out of `TYPE_CHECKING` blocks, but this will come with some tricky circular import issues.

collindutter:
https://docs.astral.sh/ruff/settings/#lint_flake8-type-checking_runtime-evaluated-base-classes set to:

```toml
runtime-evaluated-base-classes = ["griptape.mixins.serializable_mixin.SerializableMixin"]
```

Looks like we may need to refactor `griptape.common` similar to what we did for https://github.com/griptape-ai/griptape/pull/1164

PR #1757, opened 2025-02-24
fix: properly serialize pydantic models

collindutter:

### Problem:
Pydantic models are not being handled by our serialization process. This results in events that look like this:

```
{
│   'type': 'FinishStructureRunEvent',
│   'id': '224441bcb1c449b391576081914dd4a2',
│   'timestamp': 1740425885.337445,
│   'meta': {},
│   'structure_id': None,
│   'output_task_input': None,
│   'output_task_output': {
│   │   'type': 'ModelArtifact',
│   │   'id': '49469681d5c049139a34f95e5bdda990',
│   │   'reference': None,
│   │   'meta': {},
│   │   'name': '49469681d5c049139a34f95e5bdda990',
│   │   'value': Output(text='hello') # not serializable!
│   }
}
```

### Solution:
1. Introduce a `ModelArtifact` that is used exclusively for wrapping Pydantic models. This should be a non-breaking change since it subclasses `GenericArtifact`. There are [hacks](https://discuss.python.org/t/getting-type-of-a-concrete-generic-inside-another-generic-at-runtime/52144) we could use for getting information out of `GenericArtifact`, but they are hacky and would require non-trivial refactors to implement.

2. During serialization, detect instances of `BaseModel`s and serialize them using a custom marshmallow field, `fields.Model`. Deserialization for this field has not been implemented at this time as it would require us to somehow store the Pydantic model with the field. I don't _think_ users will run into this during regular usage, but the workaround would be to manually deserialize the Pydantic model outside griptape.

Serialized events now look like:
```
{
│   'type': 'FinishStructureRunEvent',
│   'id': '8ab4850b69534589a36cc2254612a004',
│   'timestamp': 1740426782.152982,
│   'meta': {},
│   'structure_id': None,
│   'output_task_input': None,
│   'output_task_output': {
│   │   'type': 'ModelArtifact',
│   │   'id': 'b3c96262b9be4d78bb89399321f84f72',
│   │   'reference': None,
│   │   'meta': {},
│   │   'name': 'b3c96262b9be4d78bb89399321f84f72',
│   │   'value': {'text': 'hello'} # serializable!
│   }
}
```
Closes https://github.com/griptape-ai/griptape/issues/1756
YASW (yet another serialization woe): #1587

### DIFF
--- a/griptape/artifacts/model_artifact.py
+++ b/griptape/artifacts/model_artifact.py
@@ -1,0 +1,24 @@
+from __future__ import annotations
+
+from attrs import define, field
+from pydantic import BaseModel
+
+from griptape.artifacts.generic_artifact import GenericArtifact
+
+
+@define
+class ModelArtifact(GenericArtifact[BaseModel]):
+    """Stores Pydantic models as Artifacts.
+
+    Required since Pydantic models require a custom serialization method.
+
+    Attributes:
+        value: The pydantic model to store.
+    """
+
+    # We must explicitly define the type rather than rely on the parent T since
+    # generic type information is lost at runtime.
+    value: BaseModel = field(metadata={"serializable": True})
+
+    def to_text(self) -> str:
+        return self.value.model_dump_json()
--- a/griptape/artifacts/__init__.py
+++ b/griptape/artifacts/__init__.py
@@ -10,6 +10,7 @@ from .image_artifact import ImageArtifact
 from .audio_artifact import AudioArtifact
 from .action_artifact import ActionArtifact
 from .generic_artifact import GenericArtifact
+from .model_artifact import ModelArtifact
 
 
 __all__ = [
@@ -25,4 +26,5 @@ __all__ = [
     "AudioArtifact",
     "ActionArtifact",
     "GenericArtifact",
+    "ModelArtifact",
 ]
--- a/griptape/schemas/__init__.py
+++ b/griptape/schemas/__init__.py
@@ -6,5 +6,7 @@ from .bytes_field import Bytes
 
 from .union_field import Union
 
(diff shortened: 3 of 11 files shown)


griptape/artifacts/model_artifact.py
```diff
@@ -0,0 +1,24 @@
...
+    Attributes:
+        value: The pydantic model to store.
+    """
+
+    # We must explicitly define the type rather than rely on the parent T since
+    # generic type information is lost at runtime.
+    value: BaseModel = field(metadata={"serializable": True})
+
+    def to_text(self) -> str:
+        return self.value.model_dump_json()
```
vachillo, line 24: do we need `to_dict` as well?

griptape/schemas/model_field.py
```diff
@@ -0,0 +1,21 @@
...
+
+from typing import TYPE_CHECKING, Any, Optional
+
+from marshmallow import fields
+
+if TYPE_CHECKING:
+    from pydantic import BaseModel
+
+
+class Model(fields.Field):
```
vachillo, line 11: im onboard with `ModelArtifact` but should this be `PydanticModel`?

vachillo: or `PydanticBaseModel`?

griptape/artifacts/model_artifact.py
collindutter: That's covered by the custom field, `Model`.  We're overriding `to_text` in order to print the value out directly rather than the entire Artifact.

griptape/schemas/model_field.py
collindutter: I'm good with `PydanticModel`, updated.

Issue #1401, opened 2024-12-04
Replace Poetry With UV


collindutter:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
[uv](https://github.com/astral-sh/uv) is a package manager written by the same people as [ruff](https://github.com/astral-sh/ruff). Ruff has been lovely so I think we should explore using uv. The installation speed improvements could have a significant effect on our github actions.

https://github.com/fgmacedo/python-statemachine/commit/a5ef7feb432ca9bbaf5d9bede3ecf7b9a85b7f45

collindutter:
Blocked on https://github.com/dependabot/dependabot-core/pull/10040

PR #1760, opened 2025-02-24
Relax Deps

### LINKED ISSUE #1759: Incompatibility with VisionAgent
When trying to install Griptape in a project with VisionAgent using poetry, I get the following error due to incompatible requirements:
```
Because no versions of vision-agent match >0.2.245,<0.3.0
 and vision-agent (0.2.245) depends on pydantic (2.7.4), vision-agent (>=0.2.245,<0.3.0) requires pydantic (2.7.4).
And because griptape (1.3.5) depends on pydantic (>=2.10.6,<3.0.0)
 and no versions of griptape match >1.3.5,<2.0.0, vision-agent (>=0.2.245,<0.3.0) is incompatible with griptape (>=1.3.5,<2.0.0).
So, because griptape-object-detection depends on both griptape (>=1.3.5,<2.0.0) and vision-agent (^0.2.245), version solving failed.
```

collindutter:

## Describe your changes
1. Runs [poetry-relax](https://github.com/zanieb/poetry-relax) on the project. See their README for more context on why.
2. Lowers attrs/pydantic versions.
BEGIN_COMMIT_OVERRIDE
deps: relax dependencies by removing upper bound constraint
END_COMMIT_OVERRIDE
Closes #1759
Makes progress towards #1401 which does not support carets.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -7389,4 +7389,4 @@ loaders-sql = ["sqlalchemy"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.9"
-content-hash = "7eb86db91a4d720d57ddb3a28007a8622d7424a5db04399f5be6877ae2362233"
+content-hash = "24a70cd5a93f3e037a27e6013daefa2042fd312e4c458cc052a1db2d20c5d45b"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,64 +13,64 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.9"
-openai = "^1.1.1"
-attrs = "^25.1.0"
-jinja2 = "^3.1.4"
-marshmallow = "^3.21.3"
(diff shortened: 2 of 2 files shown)


vachillo:
what happens with major version changes? are we relying on unit/integ tests for those?

collindutter:
@vachillo yes. We can upper bound on a per-library basis if non-solvable issues arise.

vachillo:
also do we know if our code is forward compatible with the lower attrs/pydantic versions? similar to the issue with the lower `openai` version.

collindutter:
> also do we know if our code is forward compatible with the lower attrs/pydantic versions? similar to the issue with the lower `openai` version.

Verified with `attrs@23.2.0`. Can't verify with `pydantic@2.7.4` because `ollama` requires `>=2.9.0`. But our usage of `pydantic` is very minimal so I'm not too worried about it.

vachillo (approved):

PR #1791, opened 2025-03-03
docs(custom-tools): improve docs for creating custom tools

### LINKED ISSUE #1566: Update griptape tool docs for how to pass parameters
Currently in the custom tools documentation (https://docs.griptape.ai/stable/griptape-tools/custom-tools/) it demonstrates passing parameters through a `params` parameter:

```python
    def generate(self, params: dict) -> TextArtifact:
        return TextArtifact(str(round(random.random(), params["values"].get("decimals"))))
```

But in the past few releases it's been possible to pass the parameter directly:

```python
    def generate(self, decimals: int) -> TextArtifact:
        return TextArtifact(str(round(random.random(), decimals)))
```

we should update the docs to demonstrate both options

collindutter:

## Describe your changes
Improves docs for creating custom tools. In particular, explains the nuances of the various ways to declare activity parameters.
Closes #1566 

----
📚 Documentation preview 📚: https://griptape--1791.org.readthedocs.build//1791/


docs/griptape-framework/tools/custom-tools/index.md
```diff
@@ -21,7 +21,28 @@ Here is a simple random number generator Tool:
...
 
-Check out other [Griptape Tools](https://github.com/griptape-ai/griptape/tree/main/griptape/tools) to learn more about tool implementation details.
+### Tool Activities
+
+Tool can have multiple activities. Each activity is a method in the Tool class that is decorated with the [activity](../../../reference/griptape/utils/decorators.md#griptape.utils.decorators.activity) decorator.
+
+Each activity takes a `config` keyword argument that contains the configuration for the activity. The configuration can contain:
+
+- `description` (str): A plain text description of the activity. Ensure that the description is clear and concise as it will help inform the LLM of when to pick this activity.
+- `schema` (optional): An optional instance of `Schema` that defines the input values to the activity. This field can be omitted if the activity does not require any input values.
```
cjkindel, line 31: does not require or does not take? afaik you can have a tool with all optional params but still have a schema

"This field should be omitted if the activity does not accept any input values"

docs/griptape-framework/tools/custom-tools/src/index_2.py
```diff
@@ -1,28 +1,90 @@
...
 
-class RandomNumberGenerator(BaseTool):
+class RandomTool(BaseTool):
     @activity(
         config={
             "description": "Can be used to generate random numbers",
             "schema": Schema(
-                {Optional(Literal("decimals", description="Number of decimals to round the random number to")): int}
+                {
+                    "start": int,
```
cjkindel, line 18: Does @activity have magic that will apply the documentation description (L28-29) to the schema? If not, should we document how to provide description in the schema?

collindutter: Good call, will add back an example of `Literal`.

docs/griptape-framework/tools/custom-tools/index.md
collindutter: Good point, I will more clearly distinguish between an optional schema and a schema with optional values.

docs/griptape-framework/tools/custom-tools/index.md
```diff
@@ -21,7 +21,28 @@ Here is a simple random number generator Tool:
     --8<-- "docs/griptape-framework/tools/custom-tools/logs/index_2.txt"
     ```
 
-Check out other [Griptape Tools](https://github.com/griptape-ai/griptape/tree/main/griptape/tools) to learn more about tool implementation details.
+### Tool Activities
+
```
shhlife, line 25: It might be nice to explain what an "activity" is to an Agent. Not just a method with the activity decorator - but that it is a _thing_ the agent can do.

For example:

## Tool Activities

Activities are actions an Agent can perform with a various tool. They are described in natural language and are passed to the Agent when the tool is given to it. Some examples:

- "Can be used to create a random number"
- "Can be used to select a random item from a list"

Technically, each activity is a method in the tool class that's decorated with the `activity` decorator.

...
```

I know it's described down below that each activity contains a `config` keyword with a `description` and `schema` - but we jump straight to the technical without explaining the natural language first.

**Review by shhlife — CHANGES_REQUESTED** [review]
cool - just a couple of minor notes!

**Review by cjkindel — APPROVED** [review]

File: docs/griptape-framework/tools/custom-tools/index.md [hunk 4]
```diff
@@ -21,14 +21,50 @@ Here is a simple random number generator Tool:
...
+    Do not name any schema fields as `values` or `params`. They are reserved for the Activity method signature.
 
 ## Tool Dependencies
 
 Each Tool can also have its own dependencies. You can specify them in a `requirements.txt` file in the tool directory and Griptape will install them during Tool execution.
 To start, create a directory for your Tool inside your project. The directory must have the following structure:
 
-- `tool.py` file with a tool Python class.
-- `requirements.txt` file with tool Python dependencies.
+griptape/tools/calculator/
```
shhlife, line 63: I think this all needs to be wrapped in ```

image

collindutter: Good catch!

shhlife (approved):
nice work!

PR #1794, opened 2025-03-03
docs(recipes): add example of talking to audio file

### LINKED ISSUE #1793: Add Docs For Audio Modalities

collindutter:

## Describe your changes
Adds example showing off `gpt-4o-audio-preview` for audio inputs/outputs.
Closes #1793 

----
📚 Documentation preview 📚: https://griptape--1794.org.readthedocs.build//1794/


shhlife (commented):
A couple of quick thoughts..

I think it might be helpful to explain the modalities stuff? that's new and not necessarily something people are used to seeing in Griptape yet.

Is there a list of audio voices available we could link to? Maybe: https://platform.openai.com/docs/guides/text-to-speech#voice-options

is it necessary to use `cast` to force the audio artifact? Or is that just becasue we might not actually return an AudioArtifact?


Also, I didn't see the talk-to-an-audio showing up in the Recipes list on the left side of the docs? Does it need to get added?

shhlife (approved):
awesome, looks good to me!

PR #1845, opened 2025-03-17
fix: demote tokenizer log from warning to debug

### LINKED ISSUE #1842: Resolve `WARNING: :root:Model amazon.titan-embed-text-v1 not found...` when using Amazon Titan embedding models
Whenever using a Titan embedding model with the framework, the following warning is generated:

`WARNING:root:Model amazon.titan-embed-text-v1 not found in MODEL_PREFIXES_TO_MAX_OUTPUT_TOKENS, using default value of 1000.`

collindutter:

## Describe your changes
This is warning against functionality that will be removed in the future. Could remove the log, but will keep around as debug just in-case
Closes #1842


vachillo (approved):
nice work on addressing the future functionality warning. keeping the log for debugging is a smart move.

Issue #981, opened 2024-07-15
Parameter in framework to set max_retries on ToolTask and ToolkitTask error


shhlife:
- [ x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
There is not a way to tell the agent to limit retries when it encounters an error. You can use `max_subtasks`, but that will also limit successful subtasks, which you may not want to do.

**Describe the solution you'd like**
A `max_retries` parameter on an agent, and toolkitTask.

PR #1865, opened 2025-04-04
feat: add output schema validation subtask

collindutter:


docs/griptape-framework/drivers/prompt-drivers.md
```diff
@@ -75,6 +75,11 @@ You can specify `output_schema` using either `pydantic` or `schema` libraries, t
     --8<-- "docs/griptape-framework/drivers/src/prompt_drivers_structured_output_schema.py"
     ```
 
+!!! info
+
+    `PromptTask`s will automatically validate the LLM's output against the `output_schema` you provide. If the LLM fails to output valid JSON, the Task will automatically re-prompt the LLM to try again.
+    You can configure how many times it will retry by setting the `max_subtasks` parameter on the Task.
```
vachillo, line 81: this feels like it should be controlled by a different parameter. doesnt `max_subtasks` also control the number of tool calls as well? i would definitely want to control this independently

griptape/tasks/base_subtask.py
```diff
@@ -0,0 +1,56 @@
...
+
+logger = logging.getLogger(Defaults.logging_config.logger_name)
+
+T = TypeVar("T", bound=BaseArtifact)
+
+
+@define
+class BaseSubtask(BaseTask[T]):
+    @property
+    def origin_task(self) -> BaseTask:
```
dylanholmes, line 25: What is origin vs parent?

griptape/tasks/base_subtask.py
```diff
@@ -0,0 +1,56 @@
...
+if TYPE_CHECKING:
+    from griptape.common import PromptStack
+
+logger = logging.getLogger(Defaults.logging_config.logger_name)
+
+T = TypeVar("T", bound=BaseArtifact)
+
+
+@define
+class BaseSubtask(BaseTask[T]):
```
dylanholmes, line 23: A subtask is a task? Sounds like it'd be equivalent to creating a special "ChildTask" class, which would seem weird. Is a subtask just one that get generated and inserted into the graph "dynamically"?

dylanholmes (approved):
Gets the job done -- ship it!

griptape/tasks/base_subtask.py
collindutter: The Origin Task is the Task that created the Subtask. This terminology was already in-place with `ActionsSubtask`, just generalizing here.

docs/griptape-framework/drivers/prompt-drivers.md
collindutter: The introduction of `subtask_runners`:
https://github.com/griptape-ai/griptape/blob/487c5c9786e55251ffd85976cbeacedd1f1f17f1/griptape/tasks/prompt_task.py?plain=1#L80-L86

Makes this a bit trickier since there could be more than just these two subtasks now. Ideally we could configure granular limits (success, error, total) for arbitrary subtasks. Would you be ok with shipping as is and then following up with work on https://github.com/griptape-ai/griptape/issues/981?

vachillo (approved):

PR #1909, opened 2025-04-24
Add off_prompt=True to docs example

vachillo:

## Describe your changes
Fix the docs integration tests for these examples


vachillo:
> This is in the getting started flow. If we're going to make these changes I think we need to explain what off prompt does. I assume these examples are failing just by overloading the LLM with information?

thats my best guess. maybe we can leave one bad one in and have something along the lines of "see it failed but course-corrrected, use `off_prompt` and it wont have the failure`?

collindutter (approved):

PR #1916, opened 2025-04-28
Added support for username in RedisVectorStoreDriver and RedisConversationMemoryDriver

### LINKED ISSUE #1873: Add username support to Redis conversation memory driver
It's not currently possible to use a username other than default with the Redis Conversation Memory driver.

To fix this this, I think we need to add an additional kwonly attribute for username to the driver (not sure whether this should be serializable or not) and then pass that attribute as username when creating the Redis client instance.

torabshaikh:

Added support for username in RedisVectorStoreDriver and RedisConversationMemoryDriver
 - Updated RedisVectorStoreDriver and RedisConversationMemoryDriver to add support for redis username.
 - Updated relevant docs and tests.
 -  Added `INTEG_REDIS_USERNAME` parameter in  workflow. It will require a default value in environment variable.

Closes #1873  

----
📚 Documentation preview 📚: https://griptape--1916.org.readthedocs.build//1916/


collindutter:
Thanks @torabshaikh! We're going to run some tests on our side but PR looks great.

ian-griptape-ai:
I tested this.  It passed my tests. Thanks for contributing this.

collindutter (approved):

Issue #1590, opened 2025-01-17
Persist Summary In Conversation Memory Drivers


collindutter:
Summary Conversation Memory does not store/load its summary in Summary Conversation Memory Drivers. That means this information is lost in a new session and requires the user to "nudge" the Task in order to regenerate it. We can store this information in the metadata.

### Discussed in https://github.com/orgs/griptape-ai/discussions/1585

<div type='discussions-op-text'>

<sup>Originally posted by **shhlife** January 17, 2025</sup>
It would be wonderful to be able to maintain a summary of a threaded conversation in Threads - that way you can always get the summary without needing to execute a new run.</div>

torabshaikh:
Hi @collindutter  If you would like, I can take it up.

collindutter:
Sure @torabshaikh thank you!

PR #1924, opened 2025-04-30
Fixed: Summary is not persisting in conversation memory

torabshaikh:
- [*] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape/blob/main/CONTRIBUTING.md).

## #1590 

I used this code to test it. I ran it one time as it is and second time after commenting the agent.run and found that I was able to load the runs but not summaries.

`import os
from griptape.drivers.prompt.google import GooglePromptDriver
from griptape.memory.structure import SummaryConversationMemory
from griptape.drivers.memory.conversation.redis import RedisConversationMemoryDriver

prompt_driver = GooglePromptDriver(model="gemini-2.0-flash", api_key=os.environ["GOOGLE_API_KEY"])
redis_conversation_driver = RedisConversationMemoryDriver(
    host=os.environ["REDIS_HOST"],
    username=os.environ["REDIS_USERNAME"],
    port=int(os.environ["REDIS_PORT"]),
    password=os.environ["REDIS_PASSWORD"],
    index=os.environ["REDIS_INDEX"],
    conversation_id='torab_test',
)

conversation_memory = SummaryConversationMemory(
    offset=2,
    conversation_memory_driver=redis_conversation_driver,
    prompt_driver=prompt_driver
    )
agent = Agent(
    conversation_memory=conversation_memory,
    prompt_driver=prompt_driver
    )

agent.run("My name is Torab?")
agent.run("I am trying to resolve summary conversation issue.")
agent.run("What am I doing?")

print('conversation: ', conversation_memory.to_dict())
print('conversation summary: ', conversation_memory.summary)`

I did not update the `TestSummaryConversationMemory` since it does not uses persistent memory.


griptape/memory/structure/summary_conversation_memory.py
```diff
@@ -72,3 +72,11 @@ def summarize_runs(self, previous_summary: str | None, runs: list[Run]) -> str |
...
 
             return previous_summary
+
+    def after_add_run(self) -> None:
+        if self.max_runs:
+            while len(self.runs) > self.max_runs:
+                self.runs.pop(0)
+        self.meta["summary"] = self.summary
+        self.meta["summary_index"] = self.summary_index
+        self.conversation_memory_driver.store(self.runs, self.meta)
```
collindutter, line 82: Let's set `self.meta` and then call `super(). after_add_run()`

collindutter (changes requested):

collindutter:
Thanks @torabshaikh! Could you please share a reproducible code snippet that uses `LocalConversationMemoryDriver` with a `persist_file`? That will make it easy for me to verify whether things are working as expected.

torabshaikh:
Thanks for the review. As I understand this, It will be something like this, right?

```
   def after_add_run(self) -> None:
        self.meta["summary"] = self.summary
        self.meta["summary_index"] = self.summary_index
        super().after_add_run()
    
    def load_runs(self) -> list[Run]:
        runs, meta = self.conversation_memory_driver.load()
        self.runs.extend(runs)
        self.meta = dict_merge(self.meta, meta)
        self.summary = meta['summary']
        self.summary_index = meta['summary_index']
        return self.runs
```

We could have called super().load_runs() when overriding `load_runs` but we can't do it since it returns only `self.runs`. Right?

griptape/memory/structure/summary_conversation_memory.py
collindutter, line 1: @torabshaikh Moving [this](https://github.com/griptape-ai/griptape/pull/1924#issuecomment-2842974991) here so we can get a thread. Can we override with something like this?

```python
    def load_runs(self) -> list[Run]:
        runs = super().load_runs()
        self.summary = self.meta.get("summary")
        self.summary_index = self.meta.get("summary_index", 0)

        return runs
```

torabshaikh: I got it. We can directly get summary using `self.meta.get("summary")` once it is saved in the meta attribute. This will help us in avoiding using `conversation_memory_driver` in the `SummaryConversationMemory`.

collindutter:
@torabshaikh looks like tests are failing. You can run `make test/unit` locally to see why.

torabshaikh:
Yeah. I was able to replicate the issue locally. I will update the PR.

torabshaikh:
@collindutter I debugged this and found out that metadata was not getting initialized properly.

griptape/memory/structure/summary_conversation_memory.py
```diff
@@ -29,15 +29,18 @@ class SummaryConversationMemory(BaseConversationMemory):
         kw_only=True,
     )
 
+    def __attrs_post_init__(self) -> None:
+        self.meta["summary"] = self.summary
+        self.meta["summary_index"] = self.summary_index
```
collindutter, line 34: @torabshaikh can you please explain why this is needed? I would expect the other direction.

torabshaikh: This is what I also thought, but when we initialize `SummaryConversationMemory(summary="foobar")`, we set `self.summary='foobar'` but our `meta` object still remains empty. When `load_runs` is called, it will assign that empty summary to `self.summary` and set both of them `none`, so we need to set `meta['summary]` after initializing `self.summary`. If you would like, I can post the logs I used for debugging.

collindutter: @torabshaikh apologies for the late reply. Can you please add a comment explaining this behavior in this method? And maybe let's only assign if the values are set. I.e. `if self.summary is not None: self.meta["summary"] = self.summary`

collindutter:
@torabshaikh thanks for the continued updates. Could you please some unit tests to cover this new behavior? Then I think we can consider this done.

torabshaikh:
Hi @collindutter, Sorry for the late response. I had been busy. I added some test cases to track this behavior and to test the full run.

collindutter (approved):
Amazing, thank you @torabshaikh!

PR #1954, opened 2025-06-06
Support Griptape Cloud base_url with paths

cjkindel:

### DIFF
--- a/griptape/drivers/assistant/griptape_cloud_assistant_driver.py
+++ b/griptape/drivers/assistant/griptape_cloud_assistant_driver.py
@@ -5,7 +5,6 @@ import logging
 import os
 import uuid
 from typing import TYPE_CHECKING, Optional
-from urllib.parse import urljoin
 
 import requests
 from attrs import Factory, define, field
@@ -14,6 +13,7 @@ from griptape.artifacts import BaseArtifact, TextArtifact
 from griptape.configs.defaults_config import Defaults
 from griptape.drivers.assistant import BaseAssistantDriver
 from griptape.events import BaseEvent, EventBus
+from griptape.utils.griptape_cloud import griptape_cloud_url
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
@@ -72,7 +72,7 @@ class GriptapeCloudAssistantDriver(BaseAssistantDriver):
                 self.thread_id = thread["thread_id"]
 
     def _create_thread(self, thread_alias: Optional[str] = None) -> str:
-        url = urljoin(self.base_url.strip("/"), "/api/threads")
+        url = griptape_cloud_url(self.base_url, "api/threads")
 
         body = {"name": uuid.uuid4().hex}
         if thread_alias is not None:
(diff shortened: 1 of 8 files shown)


griptape/utils/griptape_cloud.py
```diff
@@ -27,6 +28,12 @@
     from griptape.observability.observability import Observability
 
 
+def griptape_cloud_url(base_url: str, route: str) -> str:
+    if not base_url.endswith("/"):
+        base_url += "/"
+    return urljoin(base_url, route.strip("/"))
```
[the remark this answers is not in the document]
cjkindel: I thought about that, but we don't actually want to accidentally let through an API route that is / terminated. Safety net may not be necessary though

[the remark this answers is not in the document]
vachillo: didnt know `strip` does both sides.

griptape/drivers/assistant/griptape_cloud_assistant_driver.py
```diff
@@ -72,7 +72,7 @@ def _create_or_find_thread(self, thread_alias: Optional[str] = None) -> None:
                 self.thread_id = thread["thread_id"]
 
     def _create_thread(self, thread_alias: Optional[str] = None) -> str:
-        url = urljoin(self.base_url.strip("/"), "/api/threads")
+        url = griptape_cloud_url(self.base_url, "api/threads")
```
collindutter, line 75: Do we need the util? Can we just remove the strip logic?

collindutter: I think we should just update the defaults to be correct (ending in a slash). Though this would technically be a breaking change, I think it's fine since it's correcting bad behavior.

collindutter: I don't love that the util does some magic behind the scenes that will only happen on GTC-specific things. I'd prefer if there was a validator that told me URLs in GTC must be formatted a specific way. And even better, I'd prefer if all URLs had this same requirement.

But ultimately this is a harmless change that solves a problem. So I won't hold it up.

collindutter (approved):

griptape/drivers/file_manager/griptape_cloud_file_manager_driver.py
```diff
@@ -120,7 +120,7 @@ def _get_asset_url(self, full_key: str) -> tuple[str, dict]:
 
     def _get_url(self, path: str) -> str:
         path = path.lstrip("/")
-        return urljoin(self.base_url, f"/api/{path}")
+        return urljoin(self.base_url, f"api/{path}")
```
vachillo, line 123: the base url in cloud has no trailing `/`, so it should be added here

vachillo (changes requested):

griptape/drivers/file_manager/griptape_cloud_file_manager_driver.py
cjkindel: Went back to the hlper util

vachillo (approved):

PR #1983, opened 2025-08-06
feat: Handle event stream error messages in GriptapeCloudPromptDriver

cjkindel:
- [X] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape/blob/main/CONTRIBUTING.md).

## Describe your changes
event stream data messages that contain the `error` key should raise an exception as a fatal error has occurred during the streaming of events.


collindutter (approved):
epic

PR #2003, opened 2025-09-25
fix: MCPTool import path

cjkindel:
- [X] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape/blob/main/CONTRIBUTING.md).


collindutter:
Why delete?

cjkindel:
> Why delete?

This was a confirmation that this breaks the unit tests/build for Python3.9, which it does.

PR #2056, opened 2026-01-08
Lazy Large 3rd party imports

kateforsberg13:

## Describe your changes
After analyzing import times, I picked the heaviest modules and updated griptape to lazy load these. On this branch, importtime overall is 47% of the original time. These modules will be imported once used, splitting up load time. 

The largest libraries were numpy, OpenAI, pydantic, and jinja2. I also included tiktokken, since that was noted as taking a long time as well.

### DIFF
--- a/griptape/drivers/assistant/openai_assistant_driver.py
+++ b/griptape/drivers/assistant/openai_assistant_driver.py
@@ -14,6 +14,8 @@ from griptape.utils.decorators import lazy_property
 if TYPE_CHECKING:
     import openai
     from openai import AssistantEventHandler
+    from openai.types.beta.threads import Text, TextDelta
+    from openai.types.beta.threads.runs import ToolCall, ToolCallDelta
 
 
 @define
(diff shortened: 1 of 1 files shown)


griptape/drivers/image_generation/openai_image_generation_driver.py
```diff
@@ -40,11 +41,13 @@ class OpenAiImageGenerationDriver(BaseImageGenerationDriver):
         output_format: Optional and only supported for gpt-image-1. Can be either 'png' or 'jpeg'.
     """
 
-    api_type: Optional[str] = field(default=openai.api_type, kw_only=True)
-    api_version: Optional[str] = field(default=openai.api_version, kw_only=True, metadata={"serializable": True})
+    # These defaults were changed from openai.api_type, openai.api_version, and openai.organization
```
kateforsberg13: this is also happening here. Do you want me to do the same thing @collindutter ?

griptape/drivers/text_to_speech/openai_text_to_speech_driver.py
```diff
@@ -19,17 +22,20 @@ class OpenAiTextToSpeechDriver(BaseTextToSpeechDriver):
...
+    # These defaults were changed from openai.api_type, openai.api_version, and openai.organization
+    # to None because those module-level attributes don't exist in OpenAI SDK v1.0+
+    api_type: Optional[str] = field(default=None, kw_only=True)
+    api_version: Optional[str] = field(default=None, kw_only=True, metadata={"serializable": True})
     base_url: Optional[str] = field(default=None, kw_only=True, metadata={"serializable": True})
     api_key: Optional[str] = field(default=None, kw_only=True)
-    organization: Optional[str] = field(default=openai.organization, kw_only=True, metadata={"serializable": True})
+    organization: Optional[str] = field(default=None, kw_only=True, metadata={"serializable": True})
     _client: Optional[openai.OpenAI] = field(
         default=None, kw_only=True, alias="client", metadata={"serializable": False}
```
kateforsberg13: also here

griptape/drivers/image_generation/openai_image_generation_driver.py
collindutter: We should be consistent for sure, but I don't understand the comment:

```
# These defaults were changed from openai.api_type, openai.api_version, and openai.organization
# to None because those module-level attributes don't exist in OpenAI SDK v1.0+
```

Are these no longer valid?

PR #2153, opened 2026-04-16
fix(tools-mcp): sanitize MCP tool names into valid activity names

### LINKED ISSUE #2152: MCPTool fails with MCP servers that expose tool names with hyphens
When `MCPTool` initializes against an MCP server whose tools include characters outside `[a-zA-Z0-9_]` (for example, hyphens), the agent raises `ValueError: Activity name can only contain letters, numbers, and underscores.` the first time a prompt driver tries to serialize the tool schema.

The MCP spec does not restrict tool names to that character set, and many real-world servers use hyphens. The reference `@modelcontextprotocol/server-everything` is one example. Its tools include `get-annotated-message`, `get-sum`, `trigger-long-running-operation`, etc., and every one of them trips `BaseTool.to_native_tool_name`'s regex check, making the server effectively unusable from Griptape.

Reproduction:

```python
from griptape.tools.mcp.sessions import StdioConnection
from griptape.tools.mcp.tool import MCPTool

connection: StdioConnection = {
    "transport": "stdio",
    "command": "npx",
    "args": ["-y", "@modelcontextprotocol/server-everything"],
}

Agent(tools=[MCPTool(connection=connection)]).run("Add 17 and 25 together and return the result.")
```

collindutter:
Closes #2152.

MCP tool names often include characters like `-` that are invalid as Griptape activity names, causing `BaseTool.to_native_tool_name` to raise `ValueError` the moment a prompt driver serializes the tool schema. This made otherwise conformant MCP servers (including `@modelcontextprotocol/server-everything`) unusable from Griptape.

`MCPTool` now routes each discovered tool name through a `_sanitize_activity_name` helper that replaces any character outside `[a-zA-Z0-9_]` with `_` before registering the activity. The original, unsanitized name is still passed to `_run_activity`, so the MCP call itself is unaffected.

### DIFF
--- a/griptape/tools/mcp/tool.py
+++ b/griptape/tools/mcp/tool.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import re
 import threading
 from types import MethodType
 from typing import TYPE_CHECKING, Any
@@ -150,7 +151,7 @@ class MCPTool(BaseTool):
 
         for tool in tools_response.tools:
             activity_handler = self._create_activity_handler(tool)
-            setattr(self, tool.name, MethodType(activity_handler, self))
+            setattr(self, self._sanitize_activity_name(tool.name), MethodType(activity_handler, self))
 
     def _get_session(self) -> _AsyncGeneratorContextManager[ClientSession, None]:
         return create_session(self.connection)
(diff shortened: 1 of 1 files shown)


griptape/tools/mcp/tool.py
```diff
@@ -150,7 +151,7 @@ async def _init_activities(self) -> None:
 
         for tool in tools_response.tools:
             activity_handler = self._create_activity_handler(tool)
-            setattr(self, tool.name, MethodType(activity_handler, self))
+            setattr(self, self._sanitize_activity_name(tool.name), MethodType(activity_handler, self))
```
cjkindel, line 154: And if sanitized names collide we have to accept that?

collindutter: I think so. That will always be a possibility without some deeper changes.

cjkindel (approved):

PR #2155, opened 2026-04-16
refactor(google): migrate all Google drivers to google-genai SDK

### LINKED ISSUE #1624: Use New Google GenAI SDK
Google appears to have a new SDK for genai: https://googleapis.github.io/python-genai/. This new SDK seems to solve a few quirks with the current one:

- [Hack](https://github.com/griptape-ai/griptape/blob/f59dce6c475d0c05a81c40a1bfea9d6548767f53/griptape/drivers/prompt/google_prompt_driver.py?plain=1#L143-L147) for setting system prompt at runtime. https://googleapis.github.io/python-genai/#system-instructions-and-other-configs
- [Hack](https://github.com/griptape-ai/griptape/blob/f59dce6c475d0c05a81c40a1bfea9d6548767f53/griptape/drivers/prompt/google_prompt_driver.py?plain=1#L194-L224) to convert into Google's schema format. This new SDK [supports pydantic](https://googleapis.github.io/python-genai/#json-response-schema) which will be used by Griptape 2.0.
- Missing Vertex support. https://googleapis.github.io/python-genai/genai.html#genai.client.Client.vertexai

collindutter:
Migrates every Google-backed driver off the deprecated `google-generativeai` package and onto the unified `google-genai` SDK (`from google import genai`), which is now the single Google dependency. `google-generativeai` is dropped from all extras, the `all` extra, the base schema resolver's `localns`, and the `INSTALL_MAPPING` for optional-dependency error messages.

`GooglePromptDriver` now uses `genai.Client(api_key=...)` and drives requests through `client.models.generate_content` / `generate_content_stream`, passing config via `types.GenerateContentConfig` (which also carries `system_instruction`, `tools`, and `tool_config`). This removes the `_system_instruction` mutation hack and the `MessageToDict` dance for `function_call.args`. `FunctionCallingConfig.mode` is uppercase in the new SDK. Native structured output is wired up too: when `structured_output_strategy="native"`, the driver sets `response_mime_type="application/json"` and forwards the prompt stack's JSON schema via `response_json_schema`. Gemini thinking models sometimes emit reasoning-only parts carrying just a `thought_signature`; those are filtered out in both `try_run` and `try_stream` since Griptape has no thought content type.

`GoogleTokenizer` uses the same client and calls `client.models.count_tokens(model=..., contents=...)`.

`GoogleEmbeddingDriver` moves to `client.models.embed_content(..., config=types.EmbedContentConfig(task_type=..., title=...))` and reads the vector from `response.embeddings[0].values`. It also gains a lazy `client` attribute mirroring the prompt driver and tokenizer, so callers can inject a custom `Client` for Vertex AI. End-to-end verified against Vertex with `gemini-embedding-001` returning 3072-dim vectors for `str`, `TextArtifact`, and both `RETRIEVAL_DOCUMENT` / `RETRIEVAL_QUERY` task types.

The `talk_to_a_video` recipe switches from `google.generativeai.files.{upload_file,get_file}` to `client.files.{upload,get}` via the prompt driver's client. Public attrs surfaces of all three drivers are unchanged.

Closes #1624

### DIFF
--- a/griptape/drivers/prompt/google_prompt_driver.py
+++ b/griptape/drivers/prompt/google_prompt_driver.py
@@ -36,9 +36,12 @@ from griptape.utils.decorators import lazy_property
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
-    from google.generativeai.generative_models import GenerativeModel
-    from google.generativeai.protos import Part
-    from google.generativeai.types import ContentDict, ContentsType, GenerateContentResponse
+    from google.genai import Client
+    from google.genai.types import (
+        Content,
+        GenerateContentResponse,
+        Part,
+    )
 
     from griptape.drivers.prompt.base_prompt_driver import StructuredOutputStrategy
     from griptape.tools import BaseTool
@@ -53,7 +56,7 @@ class GooglePromptDriver(BasePromptDriver):
     Attributes:
         api_key: Google API key.
         model: Google model name.
-        client: Custom `GenerativeModel` client.
+        client: Custom `google.genai.Client`.
         top_p: Optional value for top_p.
         top_k: Optional value for top_k.
     """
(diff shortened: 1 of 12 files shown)


cjkindel (approved):

cjkindel:
https://github.com/griptape-ai/griptape-nodes-library-standard/issues/177 may merit a change to this PR for streaming support, taking a look to confirm

griptape/drivers/prompt/google_prompt_driver.py
cjkindel: Also, why are we explicitly rejecting `native` structured_output_strategy?

collindutter: Likely because google did not support it at the time. Will re-evaluate.

collindutter: Likely because google did not support it at the time. Will re-evaluate.

cjkindel (approved):

PR #2183, opened 2026-05-23
Harden J2 utility: switch to ImmutableSandboxedEnvironment (defense against template injection)

ghost:
## Summary

Switch the default Jinja2 environment in `griptape/utils/j2.py` from `Environment` to `ImmutableSandboxedEnvironment`. This is a defense-in-depth hardening that prevents template-injection escalation (CWE-1336 / CWE-94) without changing how legitimate prompt templates render.

## Background

`J2` exposes a public `render_from_string(value, **kwargs)` API that calls `self.environment.from_string(value).render(kwargs)`. Several task classes pipe task input directly into it:

- `griptape/tasks/base_text_input_task.py:36`
- `griptape/tasks/prompt_task.py:351`
- `griptape/tasks/structure_run_task.py:48`
- `griptape/tasks/code_execution_task.py:32`
- `griptape/tasks/prompt_image_generation_task.py:41`
- `griptape/tasks/text_to_speech_task.py:34`
- `griptape/tasks/inpainting_image_generation_task.py:42`
- `griptape/tasks/outpainting_image_generation_task.py:42`
- `griptape/tasks/variation_image_generation_task.py:51`

When any of these task inputs originates from end-user input (the typical integration pattern for an AI agent framework), the rendered template string can carry Jinja2 syntax that the default `Environment` happily evaluates — including dunder-attribute walks that reach `__class__.__mro__.__subclasses__()` and module globals like `os`.

`ImmutableSandboxedEnvironment` blocks unsafe attribute access (`__init__`, `__globals__`, `__class__`, `__mro__`, …) and mutation operations on shared collections, which is the canonical Jinja2 defense for this class of issue. The same defense was recently adopted by sister AI-agent framework `PrefectHQ/marvin` (`marvin/utilities/jinja.py`, `SandboxedEnvironment`).

Safe template features used by griptape's shipped templates continue to work after the switch:

- variable substitution (`{{ name }}`)
- attribute / dict-key access (`{{ obj.value }}`)
- filter pipelines (`{{ x | upper }}`, `{{ data | tojson }}`)
- loops (`{% for x in items %}…{% endfor %}`)
- conditionals (`{% if cond %}…{% endif %}`)

## Patch

```diff
-from jinja2 import Environment, FileSystemLoader
+from jinja2 import Environment, FileSystemLoader
+from jinja2.sandbox import ImmutableSandboxedEnvironment

 ...

-            lambda self: Environment(loader=FileSystemLoader(self.templates_dir), trim_blocks=True, lstrip_blocks=True),
+            lambda self: ImmutableSandboxedEnvironment(loader=FileSystemLoader(self.templates_dir), trim_blocks=True, lstrip_blocks=True),
```

The `Environment` import remains because the `environment: Environment` type annotation on the field still references it — `ImmutableSandboxedEnvironment` is a subclass, so the annotation continues to type-check downstream consumers.

## Verification

Local smoke test against the patched `J2`:

| Case | Pre-patch | Post-patch |
|---|---|---|
| `render_from_string('{{ name }}', name='griptape')` | `'griptape'` | `'griptape'` ✓ |
| `render_from_string('{% for x in items %}{{ x\|upper }} {% endfor %}', items=['hi','world'])` | `'HI WORLD '` | `'HI WORLD '` ✓ |
| `render_from_string('{{ obj.value }}', obj={'value': 42})` | `'42'` | `'42'` ✓ |
| Untrusted template walking dunder attrs to reach `os` | RCE | `SecurityError: access to attribute '__init__' of 'type' object is unsafe.` |

Shipped templates under `griptape/templates/**/*.j2` use only the "safe" feature subset above; none of them depend on `__class__`, `__init__`, `__globals__`, mutation of shared collections, or any other surface that `ImmutableSandboxedEnvironment` restricts. The change is therefore non-breaking for current first-party usage.

## Why this is defense-in-depth

`render_from_string` is documented as accepting "a string" — there is no API contract that the string must come from a trusted source, and the typical integration shape (user-controlled task inputs feeding through `PromptTask`, `BaseTextInputTask`, etc.) makes it natural for downstream developers to pass user input through it. Today the framework relies on each integrator to sanitize that input themselves; after this change, the framework provides a sandbox by default and integrators get safe-by-construction behavior.

## Related

- Jinja2 sandbox docs: https://jinja.palletsprojects.com/en/stable/sandbox/
- Sister framework adoption: `PrefectHQ/marvin` `marvin/utilities/jinja.py` (SandboxedEnvironment)
- CWE-1336: Improper Neutralization of Special Elements Used in a Template Engine
- CWE-94: Improper Control of Generation of Code ("Code Injection")

### DIFF
--- a/griptape/utils/j2.py
+++ b/griptape/utils/j2.py
@@ -13,7 +13,9 @@ class J2:
     templates_dir: str = field(default=abs_path("templates"), kw_only=True)
     environment: Environment = field(
         default=Factory(
-            lambda self: ImmutableSandboxedEnvironment(loader=FileSystemLoader(self.templates_dir), trim_blocks=True, lstrip_blocks=True),
+            lambda self: ImmutableSandboxedEnvironment(
+                loader=FileSystemLoader(self.templates_dir), trim_blocks=True, lstrip_blocks=True
+            ),
             takes_self=True,
         ),
         kw_only=True,
--- a/griptape/utils/j2.py
+++ b/griptape/utils/j2.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from attrs import Factory, define, field
 from jinja2 import Environment, FileSystemLoader
+from jinja2.sandbox import ImmutableSandboxedEnvironment
 
 from .paths import abs_path
 
@@ -12,7 +13,7 @@ class J2:
     templates_dir: str = field(default=abs_path("templates"), kw_only=True)
     environment: Environment = field(
         default=Factory(
-            lambda self: Environment(loader=FileSystemLoader(self.templates_dir), trim_blocks=True, lstrip_blocks=True),
+            lambda self: ImmutableSandboxedEnvironment(loader=FileSystemLoader(self.templates_dir), trim_blocks=True, lstrip_blocks=True),
             takes_self=True,
         ),
         kw_only=True,


biswajeetdev (commented):
Good hardening — `Environment` with untrusted template input is a well-known SSTI vector, and switching to `ImmutableSandboxedEnvironment` correctly blocks attribute traversal attacks like `{{ "".__class__.__mro__[1].__subclasses__() }}`.

Two things worth verifying before merging:

**1. Feature compatibility** — `ImmutableSandboxedEnvironment` is more restrictive than `SandboxedEnvironment`. It blocks `{% set %}` for global state mutation and the `namespace()` function. If any existing Griptape templates use these features (e.g. accumulating values across a loop via a namespace), they will silently produce wrong output rather than raising an error. Worth a grep across `griptape/templates/` for `namespace(` and `{% set %}` to confirm nothing regresses.

**2. Test coverage** — a short test demonstrating that a SSTI probe like `{{ config.__class__ }}` now raises a `SecurityError` rather than leaking type information would make the security guarantee explicit and guard against future accidental downgrades.

PR #2251, opened 2026-07-15
fix(drivers-prompt): detect deprecated Claude sampling params by model version

### LINKED ISSUE #2250: Sampling-parameter deprecation guard is hardcoded to claude-opus-4-7, missing 4.8 and later
## Problem

Anthropic has deprecated the `temperature`, `top_p`, and `top_k` sampling parameters for Claude Opus 4.7 **and all later models** (per the [model deprecations page](https://platform.claude.com/docs/en/about-claude/model-deprecations)). Sending any of these to an affected model causes the request to be rejected with an upstream validation error.

The prompt drivers currently guard against this, but the guards are **hardcoded to the literal string `claude-opus-4-7`** and therefore only cover that single model. They do not cover `claude-opus-4-8` (or any future model that also deprecates these parameters), so sampling parameters are still forwarded to those models and their requests will be rejected the same way 4.7's were.

Concretely, in `griptape/drivers/prompt/`:

- `amazon_bedrock_prompt_driver.py` — `supports_temperature` returns `"claude-opus-4-7" not in self.model`. `claude-opus-4-8` does not contain that substring, so `temperature` is still sent.
- `anthropic_prompt_driver.py` — `supports_temperature`, `supports_top_p`, and `supports_top_k` all return `not self.model.startswith("claude-opus-4-7")`. `claude-opus-4-8` does not start with that string, so all three params are still sent.

The desired outcome is that sampling-parameter support is determined by which models have actually deprecated the parameters, not by a match against a single hardcoded model id — so that Opus 4.8 and any subsequent affected models are covered without a code change per model.

## Origin

Reported by James Clarendon. Surfaced while triaging user-facing "service unavailable"/chat failures on Opus models; the earlier 4.7 breakage was fixed in v1.11.0 (PR #2248) but that fix is a literal `claude-opus-4-7` match. Confirmed against the current `main` of both drivers.

## Impact

Any deployment routing traffic to `claude-opus-4-8` (native Anthropic or via Bedrock) will hit the same upstream validation rejection that made 4.7 unusable, because the deprecated sampling params are still forwarded. This is latent until 4.8 traffic flows, at which point chat/agent runs on 4.8 fail. The same gap applies to any future model on the deprecation list.

## Repro / evidence

- Bedrock guard: `amazon_bedrock_prompt_driver.py`, `supports_temperature` → `return "claude-opus-4-7" not in self.model`.
- Anthropic guard: `anthropic_prompt_driver.py`, `supports_temperature` / `supports_top_p` / `supports_top_k` → `return not self.model.startswith("claude-opus-4-7")`.
- Deprecation source: temperature, top_p, and top_k deprecated for Opus 4.7 and later.
- Note the Bedrock driver only guards `temperature` (it doesn't forward `top_p`/`top_k` by default), whereas the Anthropic native driver forwards and must guard all three.

## Possible direction

One option (not prescriptive): drive the `supports_*` checks off a set/pattern of affected model families rather than a single hardcoded id, so newly-deprecated models are picked up centrally. The right shape is up to whoever picks this up.

collindutter:
Closes #2250

### DIFF
--- a/griptape/utils/anthropic_utils.py
+++ b/griptape/utils/anthropic_utils.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+import re
+
+# Claude model families, and the minimum ``(major, minor)`` version within each family, that no
+# longer accept the ``temperature``, ``top_p``, and ``top_k`` sampling parameters. A model in one of
+# these families at or above the listed version has those parameters deprecated and must not receive
+# them. New families or version thresholds are picked up everywhere by editing this mapping.
+SAMPLING_PARAMS_DEPRECATED_MIN_VERSIONS: dict[str, tuple[int, int]] = {
+    "opus": (4, 7),
+}
+
+# Captures the family, ``major``, and optional ``minor`` version from a Claude model identifier,
+# e.g. ``claude-opus-4-7`` within ``us.anthropic.claude-opus-4-7-20251101-v1:0``. Plain aliases,
+# provider prefixes, geographic inference-profile IDs, ARNs, and dated variants all embed this
+# substring, so it is matched anywhere in the identifier. The minor version is optional because
+# major-only releases omit it (``claude-opus-5``) and pre-4.6 snapshots replace it with a date
+# (``claude-opus-4-20250514``); restricting it to one or two digits that are not followed by another
+# digit keeps an eight-digit date from being read as the minor.
+_CLAUDE_VERSION_PATTERN = re.compile(r"claude-([a-z]+)-(\d+)(?:-(\d{1,2})(?!\d))?")
+
+
+def supports_sampling_params(model: str) -> bool:
+    """Whether a Claude model accepts the temperature, top_p, and top_k sampling parameters.
+
+    Support is derived from the Claude model family and version embedded in ``model`` rather than a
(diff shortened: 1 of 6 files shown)


griptape/utils/anthropic_utils.py
```diff
@@ -0,0 +1,34 @@
...
+# them. New families or version thresholds are picked up everywhere by editing this mapping.
+SAMPLING_PARAMS_DEPRECATED_MIN_VERSIONS: dict[str, tuple[int, int]] = {
+    "opus": (4, 7),
+}
+
+# Captures the family and ``major``/``minor`` version from a Claude model identifier, e.g.
+# ``claude-opus-4-7`` within ``us.anthropic.claude-opus-4-7-20251101-v1:0``. Plain aliases, provider
+# prefixes, geographic inference-profile IDs, ARNs, and dated variants all embed this substring, so
+# it is matched anywhere in the identifier.
+_CLAUDE_VERSION_PATTERN = re.compile(r"claude-([a-z]+)-(\d+)-(\d+)")
```
cjkindel: What about patterns like `claude-fable-5`? No minor version

cjkindel (approved):

PR #2253, opened 2026-07-20
feat(ollama): expose api_key and headers on Ollama clients

### LINKED ISSUE #2238: OllamaPromptDriver: expose headers/api_key on ollama.Client for auth reverse proxy support
## Summary

`OllamaPromptDriver` constructs its `ollama.Client` with only `host`:

```python
return import_optional_dependency("ollama").Client(host=self.host)
```

The underlying `ollama.Client.__init__` accepts `**kwargs` forwarded to httpx, and the internal `_build_ollama_client` helper already handles an `api_key` kwarg → `Authorization: Bearer` header. But there is no way to reach that path through `OllamaPromptDriver`.

## Impact

Users running Ollama behind an authentication reverse proxy cannot forward credentials through the native driver. The only workaround today is to use the OpenAI-compatible driver instead — which drops tool calling support on Ollama (tools return blank output).

## Requested change

Expose an optional `headers: dict[str, str] | None` field (or `api_key: str | None`) on `OllamaPromptDriver` and forward it when constructing `ollama.Client`:

```python
return import_optional_dependency("ollama").Client(host=self.host, headers=self.headers)
```

This would allow auth-proxied Ollama setups to pass credentials while still using the native driver (and getting correct tool calling).

## Context

Discovered while switching Griptape Nodes Library from the OpenAI-compat driver to the native `OllamaPromptDriver` to fix blank tool-call output. The api_key present on hand-configured Ollama providers is silently dropped on the native path.

Solaris-star:
## Summary
`OllamaPromptDriver` (and the embedding driver) only passed `host` into `ollama.Client`, so reverse-proxy auth was impossible without dropping down to the OpenAI-compatible driver (and losing native tool calling).

## Change
- Add optional `api_key` and `headers` fields on `OllamaPromptDriver` and `OllamaEmbeddingDriver`
- Forward them to `ollama.Client(**kwargs)` (ollama already maps `api_key` → `Authorization`)
- Unit test covers client construction kwargs

Closes #2238.

### DIFF
--- a/griptape/drivers/embedding/ollama_embedding_driver.py
+++ b/griptape/drivers/embedding/ollama_embedding_driver.py
@@ -24,11 +24,18 @@ class OllamaEmbeddingDriver(BaseEmbeddingDriver):
 
     model: str = field(kw_only=True, metadata={"serializable": True})
     host: str | None = field(default=None, kw_only=True, metadata={"serializable": True})
+    api_key: str | None = field(default=None, kw_only=True, metadata={"serializable": False})
+    headers: dict[str, str] | None = field(default=None, kw_only=True, metadata={"serializable": False})
     _client: Client | None = field(default=None, kw_only=True, alias="client", metadata={"serializable": False})
 
     @lazy_property()
     def client(self) -> Client:
-        return import_optional_dependency("ollama").Client(host=self.host)
+        client_kwargs: dict = {"host": self.host}
+        if self.headers is not None:
+            client_kwargs["headers"] = self.headers
+        if self.api_key is not None:
+            client_kwargs["api_key"] = self.api_key
+        return import_optional_dependency("ollama").Client(**client_kwargs)
 
     def try_embed_chunk(self, chunk: str, **kwargs) -> list[float]:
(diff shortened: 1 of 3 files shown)


collindutter (dismissed):

collindutter (changes requested):
Thanks for the PR @Solaris-star. Code looks good but tests are failing, can you please fix?

Solaris-star:
Fixed in `e9078942`.

The new auth-forwarding test had accidentally inherited the `use_native_tools` / `structured_output_strategy` parametrization intended for `test_try_run`, which caused test collection errors. The parametrization decorators are now back on `test_try_run`.

Local verification:
- `uv run --no-sync pytest tests/unit/drivers/prompt/test_ollama_prompt_driver.py tests/unit/drivers/embedding/test_ollama_embedding_driver.py -q` — 22 passed
- targeted Ruff check/format — passed
- targeted Pyright for both changed drivers — 0 errors
- `git diff --check` — passed

Please rerun CI when convenient.

collindutter (approved):

collindutter:
Thank you!

PR #2276, opened 2026-08-23
fix(drivers-prompt): treat streamed usage deltas as running totals, not increments

### LINKED ISSUE #2275: Streaming token usage inflated for Anthropic/Gemini drivers: cumulative usage deltas are summed as increments
### Affected code (current main @ 2e97042)

`BasePromptDriver.__process_stream` aggregates streamed usage with:

```python
usage += message_delta.usage
```

`BaseMessage.Usage.__add__` (`griptape/common/prompt_stack/messages/base_message.py`) sums each field treating `None` as `0`, i.e. it assumes every delta carries a **per-delta increment**.

### But providers report running totals

- **Anthropic**: `AnthropicPromptDriver.try_stream` yields `Usage(output_tokens=event.usage.output_tokens)` on *every* `message_delta`. Per Anthropic's streaming API, that value is the **cumulative** number of output tokens so far.
- **Google (Gemini)**: `GooglePromptDriver.try_stream` yields `candidates_token_count` on every chunk, which is likewise cumulative; the code even works around this for the static prompt token count ("Only want to output the prompt token count once") but the output side still gets summed.
- OpenAI / Bedrock / Cohere / HuggingFace emit exactly **one** final usage delta, so summation happens to be correct for them.

### Result

For Anthropic and Gemini streams the final `Message.usage.output_tokens` (and therefore `FinishPromptEvent.output_token_count`) grows quadratically with chunk count: a 100-chunk response emitting cumulative totals 1..100 is reported as **5050** output tokens instead of **100**. Any cost accounting or telemetry built on streamed usage overcounts accordingly. As a side effect, `None` (unknown) fields are silently coerced to concrete `0`s by `__add__`.

### Minimal repro

```python
from griptape.common import DeltaMessage, PromptStack
from tests.mocks.mock_stream_usage_prompt_driver import MockStreamUsagePromptDriver

driver = MockStreamUsagePromptDriver(
    stream=True,
    stream_usage_deltas=[
        DeltaMessage.Usage(input_tokens=100),
        DeltaMessage.Usage(output_tokens=1),
        DeltaMessage.Usage(output_tokens=2),   # cumulative total is 2
    ],
)
message = driver.run(PromptStack())
print(message.usage.output_tokens)  # -> 3 (summed), expected 2 (latest running total)
```

### Suggested direction

Aggregate field-wise with last-reported-wins in `__process_stream` (equivalent for single-final-delta providers, correct for running-total providers), leaving the public `Usage.__add__` untouched. Happy to open a PR with regression tests if this sounds right.

feiiiiii5:
## Root cause

`BasePromptDriver.__process_stream` aggregated streamed usage with `usage += message_delta.usage`. `BaseMessage.Usage.__add__` sums each field (treating `None` as `0`), which assumes every delta carries a *per-delta increment* — but Anthropic (`message_delta` usage) and Gemini (per-chunk `usage_metadata`) report **running totals**. A 100-chunk Anthropic response emitting cumulative totals `1..100` was reported as `5050` output tokens instead of `100`; unknown fields were also silently coerced to `0`.

Details and repro in #2275.

## Fix

Aggregate field-wise with last-reported-wins in `__process_stream`, keeping the public `Usage.__add__` untouched:

```python
if message_delta.usage.input_tokens is not None:
    usage.input_tokens = message_delta.usage.input_tokens
if message_delta.usage.output_tokens is not None:
    usage.output_tokens = message_delta.usage.output_tokens
```

Equivalent for providers that emit one final usage delta (OpenAI, Bedrock, Cohere, HuggingFace), correct for running-total providers (Anthropic, Gemini). Unknown fields now stay honestly `None` when never reported.

## Test

- [x] Regression test: running-total sequence `[Usage(input=100), Usage(output=1), Usage(output=2)]` → final `Usage(input=100, output=2)` — **fails on main** (reports `output=3`)
- [x] Contract test: single final delta `[Usage(7, 9)]` → `(7, 9)` unchanged
- [x] New mock `tests/mocks/mock_stream_usage_prompt_driver.py` emits configurable usage-only deltas
- [x] `pytest tests/unit/drivers/prompt tests/unit/structures tests/unit/common` → 1045 passed; failures identical to clean-main baseline under the same environment (missing optional ollama deps), no new failures
- [x] `ruff check` + `ruff format --check` clean on changed files

## Diff scope

3 files: driver aggregation loop (+9/-1), one new mock (+27), two regression tests in `test_base_prompt_driver.py` (+31/-0).

Fixes #2275

### DIFF
--- a/griptape/drivers/prompt/base_prompt_driver.py
+++ b/griptape/drivers/prompt/base_prompt_driver.py
@@ -171,7 +171,15 @@ class BasePromptDriver(SerializableMixin, ExponentialBackoffMixin, ABC):
         # Aggregate all content deltas from the stream
         message_deltas = self.try_stream(prompt_stack)
         for message_delta in message_deltas:
-            usage += message_delta.usage
+            # Providers report usage as running totals over the stream -- e.g. Anthropic's
+            # cumulative output tokens on every `message_delta`, Gemini's per-chunk
+            # `usage_metadata` -- rather than as per-delta increments. Keep the latest
+            # reported value per field: summing would inflate those providers' totals,
+            # while being equivalent for providers that emit a single final usage delta.
+            if message_delta.usage.input_tokens is not None:
+                usage.input_tokens = message_delta.usage.input_tokens
+            if message_delta.usage.output_tokens is not None:
+                usage.output_tokens = message_delta.usage.output_tokens
             content = message_delta.content
 
             if content is not None:
--- a/tests/mocks/mock_stream_usage_prompt_driver.py
+++ b/tests/mocks/mock_stream_usage_prompt_driver.py
@@ -1,0 +1,27 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from attrs import define, field
+
+from griptape.common import DeltaMessage, PromptStack
+from tests.mocks.mock_prompt_driver import MockPromptDriver
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+
+@define
+class MockStreamUsagePromptDriver(MockPromptDriver):
+    """Emits a fixed sequence of usage-only deltas from try_stream.
+
+    Mimics providers that report usage as running totals over the stream rather than
+    per-delta increments (Anthropic's cumulative ``message_delta`` usage, Gemini's
+    per-chunk ``usage_metadata``).
+    """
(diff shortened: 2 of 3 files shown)


collindutter (approved):
Thank you!

PR #2284, opened 2026-08-26
fix(mcp): handle structured_content-only results and fix is_error field name (fixes #2281)

### LINKED ISSUE #2281: MCP tools returning only structuredContent give the agent an empty result
## Problem

When an MCP server returns its result as structured output, `MCPTool` hands the agent an empty result instead of the data.

Per the MCP spec, a tool may return its payload in `CallToolResult.structuredContent` (an object or array) and leave `content` empty. `MCPTool._convert_call_tool_result_to_artifact` only iterates `call_tool_result.content` and never reads `structuredContent`, so in that case the loop produces zero artifacts and the method returns `ListArtifact([])`.

The agent sees a successful tool call with no data. There's no error surfaced, so the failure is silent — the model just continues without the result it asked for, and typically hallucinates or reports that the tool returned nothing.

## Origin

Reported by Joshua Bretag, against `griptape/tools/mcp/tool.py` → `MCPTool._convert_call_tool_result_to_artifact`.

Affects any MCP server that returns structured-only output. Deterministic — not load- or timing-dependent.

## Impact

Structured-output MCP tools are effectively unusable through `MCPTool`. Because it fails silently rather than raising, the symptom surfaces as a model-quality problem (agent ignoring or inventing tool results) rather than as an obvious integration bug, which makes it expensive to diagnose from the outside.

Servers that populate both `content` and `structuredContent` are unaffected, since the `content` branch still yields artifacts.

## Repro / evidence

1. Point `MCPTool` at an MCP server with a tool that returns structured output — `structuredContent` set to an object or array, `content` an empty list.
1. Invoke that tool from an agent.
1. The activity returns `ListArtifact([])`; the agent receives no tool result data.

Read sites in `_convert_call_tool_result_to_artifact` for reference: the `isError` branch reads `content[0].text`, and the success path iterates `content` only. `structuredContent` is not referenced anywhere in the file.

Worth checking as part of the same fix: the `isError` branch indexes `call_tool_result.content[0]` unconditionally, which will raise `IndexError` on an error result that carries an empty `content` list — the same empty-`content` shape that triggers the main bug.

litterairplane:
## Summary

Fix `MCPTool._convert_call_tool_result_to_artifact` to handle `structured_content`-only results and two related bugs. Fixes #2281.

## Motivation

MCP servers may return their payload in `structured_content` while leaving `content` empty (per the MCP spec). The current code only iterates `content`, so those calls silently return `ListArtifact([])` — the agent receives nothing and typically hallucinates or reports the tool returned no data.

Two additional bugs were found in the same method while investigating:

- **Field name mismatch**: `call_tool_result.isError` is used, but the `mcp` SDK field is `is_error` (snake_case). Error results were silently falling through the success path.
- **`IndexError` on empty error content**: `content[0]` was accessed unconditionally in the error branch, which raises `IndexError` when `content` is an empty list — the same shape that triggers the main bug.

## Design

- After processing `content` items as before, if no artifacts were produced **and** `structured_content is not None`, serialize it as `json.dumps(...)` into a `TextArtifact`. This mirrors how `TextContent` results are handled and keeps the output consistent.
- `content` takes priority: servers that populate both fields are unaffected.
- The error-path `content[0]` access is guarded with a conditional.

## Changes

- `griptape/tools/mcp/tool.py` — three fixes in `_convert_call_tool_result_to_artifact`
- `tests/unit/tools/test_mcp_tool.py` — 6 new unit tests (was a stub `class TestMCPTool: ...`)

## Test plan

- `test_structured_content_only_returns_json_text` — the main regression case: structured-only result produces a `TextArtifact` with the JSON payload
- `test_content_takes_priority_over_structured_content` — both fields present; content wins
- `test_error_result_with_content` — error path with text returns `ErrorArtifact`
- `test_error_result_with_empty_content_does_not_raise` — empty content on error returns fallback message (no `IndexError`)
- `test_text_content_returned` — existing happy path still works
- `test_empty_result_returns_empty_list_artifact` — both fields absent/None returns empty `ListArtifact`

All 6 tests pass.

## Non-goals

- No changes to how `ImageContent`, `AudioContent`, or `EmbeddedResource` items are handled.
- No changes to the `structured_content` handling when `content` is non-empty (content always wins).

## Compatibility

Fully backwards-compatible. Servers that already return `content` are unaffected. The `is_error` fix corrects silent misbehaviour that was never intentional.

### DIFF
--- a/griptape/tools/mcp/tool.py
+++ b/griptape/tools/mcp/tool.py
@@ -224,8 +224,9 @@ class MCPTool(BaseTool):
     ) -> ListArtifact | ErrorArtifact:
         from mcp import types  # pyright: ignore[reportAttributeAccessIssue]
 
-        if call_tool_result.isError:
-            return ErrorArtifact(call_tool_result.content[0].text or "An unknown error occurred.")
+        if call_tool_result.is_error:
+            error_text = call_tool_result.content[0].text if call_tool_result.content else None
+            return ErrorArtifact(error_text or "An unknown error occurred.")
 
         response_artifacts: list[BaseArtifact] = []
         for content in call_tool_result.content:
@@ -242,5 +243,10 @@ class MCPTool(BaseTool):
                     response_artifacts.append(TextArtifact(content.resource.text))
                 elif isinstance(content.resource, types.BlobResourceContents):
                     response_artifacts.append(BlobArtifact(value=content.resource.blob))
+
+        if not response_artifacts and call_tool_result.structured_content is not None:
+            import json
+
+            response_artifacts.append(TextArtifact(json.dumps(call_tool_result.structured_content)))
 
         return ListArtifact(response_artifacts)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,6 +115,7 @@ loaders-pdf = ["pypdf>=5.0.1"]
 loaders-image = ["pillow>=11.0.0"]
 loaders-email = ["mail-parser>=4.0.0"]
 loaders-sql = ["sqlalchemy>=2.0.31"]
+tools-mcp = ["mcp>=1.10.0"]
 all = [
   "cohere>=5.11.2",
   "anthropic>=0.45.1",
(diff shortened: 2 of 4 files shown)


collindutter:
Thanks for the quick PR @litterairplane. This is important for us to get out quickly so I've extended your PR [here](https://github.com/griptape-ai/griptape/pull/2285).

PR #2285, opened 2026-08-26
fix(tools-mcp): return structuredContent when content is empty

### LINKED ISSUE #2281: MCP tools returning only structuredContent give the agent an empty result

collindutter:

## Describe your changes

`MCPTool` only read `CallToolResult.content`, so a server that returns its payload in `structuredContent` handed the agent an empty `ListArtifact` with no error. Fall back to `structuredContent` when no content blocks convert, and read error text from the first text block instead of indexing `content[0]`, which raised on error results whose content is empty or non-text.

Closes #2281

### DIFF
--- a/griptape/tools/mcp/requirements.txt
+++ b/griptape/tools/mcp/requirements.txt
@@ -1,2 +1,2 @@
-mcp[ws]>=1.10.0
+mcp[ws]>=1.10.0,<2
 exceptiongroup>=1.2.2
--- a/griptape/tools/mcp/tool.py
+++ b/griptape/tools/mcp/tool.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import json
 import re
 import threading
 from types import MethodType
@@ -224,7 +225,7 @@ class MCPTool(BaseTool):
     ) -> ListArtifact | ErrorArtifact:
         from mcp import types  # pyright: ignore[reportAttributeAccessIssue]
 
-        if call_tool_result.is_error:
+        if call_tool_result.isError:
             error_text = call_tool_result.content[0].text if call_tool_result.content else None
             return ErrorArtifact(error_text or "An unknown error occurred.")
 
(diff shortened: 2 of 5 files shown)


mcaloud (approved):
Looks good, though claude found another issue. [Filed a separate ticket for it](https://github.com/griptape-ai/griptape/issues/2286)

PR #2289, opened 2026-08-27
fix(tools-mcp): use removeprefix instead of lstrip for image/audio MIME format

### LINKED ISSUE #2286: MCPTool derives image/audio format with str.lstrip, corrupting formats like gif and aac
## Problem

`MCPTool` derives the artifact `format` from an MCP content block's MIME type with `str.lstrip`, which strips a *set of characters* rather than a prefix.

`griptape/tools/mcp/tool.py`:

```python
ImageArtifact(value=content.data, format=content.mimeType.lstrip("image/"), width=0, height=0)
AudioArtifact(value=content.data, format=content.mimeType.lstrip("audio/"))
```

`lstrip("image/")` removes any leading characters in `{i, m, a, g, e, /}`, so it keeps eating past the prefix whenever the subtype begins with one of those letters:

| `mimeType` | `format` | correct |
| --- | --- | --- |
| `image/png` | `png` | ✅ |
| `image/jpeg` | `jpeg` | ✅ |
| `image/gif` | `f` | ❌ |
| `image/apng` | `png` | ❌ |
| `audio/mpeg` | `mpeg` | ✅ |
| `audio/aac` | `c` | ❌ |

## Impact

`ImageArtifact.format` and `AudioArtifact.format` feed both the reconstructed MIME type and the generated filename:

```python
# griptape/artifacts/image_artifact.py
def mime_type(self) -> str:
    return f"image/{self.format}"

def name(self) -> str:
    return f"image_artifact_{fmt_time}_{entropy}.{self.format}"
```

So a GIF returned by an MCP server becomes `mime_type == "image/f"` and `name == "image_artifact_….f"`. Passing that artifact to a prompt driver sends `media_type: "image/f"` to the provider, which the Anthropic and OpenAI APIs reject — the run fails with a provider-side error that points nowhere near `MCPTool`. Saving the artifact writes a file with a meaningless extension.

Affects only subtypes whose first character is in the stripped set (`gif`, `apng`, `avif`, `aac`, …); `png`, `jpeg`, `webp`, `mp3`, `wav` happen to be unaffected, which is why this has gone unnoticed.

## Repro / evidence

```python
>>> "image/gif".lstrip("image/")
'f'
>>> "audio/aac".lstrip("audio/")
'c'
>>> "image/apng".lstrip("image/")
'png'
```

End to end: point `MCPTool` at a server whose tool returns an `ImageContent` block with `mimeType: "image/gif"`, invoke it from an agent, and inspect the resulting `ImageArtifact` — `format` is `"f"`.

## Suggested fix

Use a prefix-aware operation:

```python
format=content.mimeType.removeprefix("image/")
format=content.mimeType.removeprefix("audio/")
```

`str.removeprefix` is available on all supported interpreters (`requires-python = ">=3.10"`). `mimeType.split("/", 1)[1]` is the equivalent already used in `griptape/drivers/image_generation/leonardo_image_generation_driver.py:117`.

Worth adding a small table-driven test over `image/gif`, `image/apng`, and `audio/aac`, since the currently exercised subtypes all pass by coincidence.

## Origin

Found while reviewing #2281's fix — same function, unrelated cause. Not covered by that issue.

dineshyadav03:

## Describe your changes

`MCPTool._convert_call_tool_result_to_artifact` derived `ImageArtifact`/`AudioArtifact` `format` from the MCP result's `mimeType` using `str.lstrip("image/")` / `str.lstrip("audio/")`. `lstrip` strips any leading *characters* in the given set, not a literal prefix, so it silently corrupts any format whose subtype starts with a letter already in `{i,m,a,g,e,/}` (image) or `{a,u,d,i,o,/}` (audio):

- `"image/gif".lstrip("image/")` -> `"f"` (should be `"gif"`)
- `"image/apng".lstrip("image/")` -> `"png"` (should be `"apng"`)
- `"audio/aac".lstrip("audio/")` -> `"c"` (should be `"aac"`)

This corrupted `format` then flows into the `mime_type` sent back to LLM providers and into the file extension used when saving the artifact. It happened to go unnoticed because the existing test only covered `image/png`, which coincidentally survives `lstrip` unscathed.

Fix: use `str.removeprefix`, which already the established pattern elsewhere in this codebase (e.g. `griptape_cloud_assistant_driver.py`, `griptape_cloud_prompt_driver.py`) for exact-prefix stripping.

Added parametrized regression tests covering `image/gif`, `image/apng`, `audio/wav`, and `audio/aac`; confirmed they fail against the old `lstrip` implementation and pass with `removeprefix`. `ruff check` and `ruff format` are clean on both changed files.

Fixes #2286

### DIFF
--- a/griptape/tools/mcp/tool.py
+++ b/griptape/tools/mcp/tool.py
@@ -249,10 +249,12 @@ class MCPTool(BaseTool):
                 response_artifacts.append(TextArtifact(content.text))
             elif isinstance(content, types.ImageContent):
                 response_artifacts.append(
-                    ImageArtifact(value=content.data, format=content.mimeType.lstrip("image/"), width=0, height=0)
+                    ImageArtifact(value=content.data, format=content.mimeType.removeprefix("image/"), width=0, height=0)
                 )
             elif isinstance(content, types.AudioContent):
-                response_artifacts.append(AudioArtifact(value=content.data, format=content.mimeType.lstrip("audio/")))
+                response_artifacts.append(
+                    AudioArtifact(value=content.data, format=content.mimeType.removeprefix("audio/"))
+                )
             elif isinstance(content, types.EmbeddedResource):
                 if isinstance(content.resource, types.TextResourceContents):
                     response_artifacts.append(TextArtifact(content.resource.text))
--- a/tests/unit/tools/test_mcp_tool.py
+++ b/tests/unit/tools/test_mcp_tool.py
@@ -3,7 +3,7 @@ import json
 import pytest
 from mcp import types as mcp_types
 
-from griptape.artifacts import ErrorArtifact, ListArtifact, TextArtifact
+from griptape.artifacts import AudioArtifact, ErrorArtifact, ImageArtifact, ListArtifact, TextArtifact
 from griptape.tools.mcp.tool import MCPTool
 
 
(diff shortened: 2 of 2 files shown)


collindutter (approved):
Thanks!

Issue #995, opened 2024-07-18
Add More Ruff Rules


collindutter:
- [x] I have read and agree to the [contributing guidelines](https://github.com/griptape-ai/griptape#contributing).

**Is your feature request related to a problem? Please describe.**
Currently Griptape only uses a subset of the [available ruff rules](https://docs.astral.sh/ruff/rules/). To cut down on PR review time, we should enable as many ruff rules as possible, ignoring ones that we don't agree with or are prohibitive to fix.

**Describe the solution you'd like**
Incrementally add ruff rules. Eventually replace the rules with `ALL` and start removing ignored rules.

mayankdas2005:
Hi @collindutter , I'm looking to make my first contribution to this repository and I would love to take up this issue. Can I work on it?

collindutter:
Hey @mayankdas2005 thanks for offering! Feel free to start on some smaller rules, would love one PR per rule.

PR #2290, opened 2026-08-27
Enable Ruff rule B009 (get-attr-with-constant)

mohansree14:
## What
Enables the `B009` (`get-attr-with-constant`) Ruff rule, as suggested in #995.

## Why
`getattr(obj, "attr")` with a string-literal attribute name is not any safer than plain `obj.attr` access. With `B009` enabled, new instances are caught.

## Changes
- Removed `B009` from the `ignore` list in `pyproject.toml`.
- Suppressed the 14 existing violations with `# noqa: B009` (via `ruff check --select B009 --add-noqa`), leaving the `getattr()` calls unchanged. Most read attributes that are attached at runtime (the `@activity` decorator's `name`/`config`, or `audio`, which the OpenAI SDK does not type), so rewriting them as direct access would need `pyright: ignore` comments instead. Files:
  - `griptape/drivers/prompt/openai_chat_prompt_driver.py`
  - `griptape/memory/task/task_memory.py`
  - `griptape/mixins/activity_mixin.py`
  - `griptape/mixins/serializable_mixin.py`
  - `griptape/tasks/prompt_task.py`
  - `griptape/tasks/tool_task.py`
  - `griptape/tools/base_tool.py`

No behavior change: apart from the `pyproject.toml` line, the diff against `main` is comments only.

## Testing
- `ruff check .` (ruff 0.15.15, the locked version): all checks passed.
- `ruff format --check .`: 1235 files already formatted.

### DIFF
--- a/griptape/drivers/prompt/openai_chat_prompt_driver.py
+++ b/griptape/drivers/prompt/openai_chat_prompt_driver.py
@@ -439,9 +439,9 @@ class OpenAiChatPromptDriver(BasePromptDriver):
                     )
                 raise ValueError(f"Unsupported tool call delta: {tool_call}")
             raise ValueError(f"Unsupported tool call delta length: {len(tool_calls)}")
-        # OpenAi doesn't have types for audio deltas so we need to use hasattr and getattr.
-        if hasattr(content_delta, "audio") and content_delta.audio is not None:
-            audio_chunk: dict = content_delta.audio
+        # OpenAi doesn't have types for audio deltas so we need to use hasattr.
+        if hasattr(content_delta, "audio") and content_delta.audio is not None:  # pyright: ignore[reportAttributeAccessIssue]
+            audio_chunk: dict = content_delta.audio  # pyright: ignore[reportAttributeAccessIssue]
             return AudioDeltaMessageContent(
                 id=audio_chunk.get("id"),
                 data=audio_chunk.get("data"),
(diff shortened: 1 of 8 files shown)


mohansree14:
Added `pyright: ignore[reportFunctionMemberAccess]` (and one `reportAttributeAccessIssue`) annotations at the sites where the B009 cleanup replaced `getattr()` with direct attribute access. Pyright can't see these attributes because they're attached at runtime (via the `@activity` decorator, or absent from the upstream OpenAI SDK types), so direct access flags where `getattr()` didn't. No behavior changes — type-check only.

mohansree14 (commented):
Added pyright: ignore[reportFunctionMemberAccess] (and one reportAttributeAccessIssue) annotations at the sites where the B009 cleanup replaced getattr() with direct attribute access. Pyright can't see these attributes because they're attached at runtime (via the @activity decorator, or absent from the upstream OpenAI SDK types), so direct access flags where getattr() didn't. No behavior changes — type-check only.

collindutter (changes requested):
Thanks for the PR @mohansree14. The use of `pyright: ignore` that these changes necessitate makes me a little uneasy. I think we should just `noqa` away the existing violations.

mohansree14:
Thanks @collindutter, that makes sense. Switched to `noqa` in `eff369ac`: the `getattr()` calls are back to what's on `main`, the `pyright: ignore` comments are gone, and the 14 existing violations each carry `# noqa: B009`. B009 stays enabled in `pyproject.toml`, so new violations are still caught. Apart from that line, the diff is comments only. `ruff check .` and `ruff format --check .` are clean with the locked ruff 0.15.15.

collindutter (approved):
Thanks!

PR #2339, opened 2026-09-10
chore: enable Ruff PLW1641 rule

### LINKED ISSUE #995: Add More Ruff Rules

krisnaparahita:
Fixes #995

## Summary

- Remove PLW1641 (eq-without-hash) from the Ruff ignore list.
- Explicitly set BooleanArtifact.__hash__ to None, matching its existing unhashable behavior and the behavior of other Artifact subclasses.
- Keep the change scoped to one Ruff rule, following the maintainer preference in #995.

## Validation

- Ruff passes.
- Pyright passes.
- All 81 relevant unit tests pass.

### DIFF
--- a/griptape/artifacts/boolean_artifact.py
+++ b/griptape/artifacts/boolean_artifact.py
@@ -34,5 +34,7 @@ class BooleanArtifact(BaseArtifact):
     def __eq__(self, value: object) -> bool:
         return self.value == value
 
+    __hash__ = None
+
     def to_text(self) -> str:
         return str(self.value).lower()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -256,7 +256,6 @@ ignore = [
   "SLF001", # private-member-access
   "EM102",  # f-string-in-exception
   "PLC0415", # import-outside-top-level
-  "PLW1641", # eq-without-hash
   "A003",   # builtin-attribute-shadowing
   "FBT001", # boolean-typed-positional-argument
 ]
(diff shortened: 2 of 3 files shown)


collindutter (commented):
Hey thanks for the PR. I'm not super familiar with the implications of setting `__hash__` to `None`. Is this the right approach?

krisnaparahita:
Thanks for taking a look! 

Yes, __hash__ = None is the right call here.
it's actually the standard Python idiom for this (https://docs.python.org/3/reference/datamodel.html#object.__hash__): if a class overrides __eq__ and shouldn't be hashable, this is how you make that explicit.

collindutter (approved):

collindutter:
> Thanks for taking a look!
> 
> Yes, **hash** = None is the right call here. it's actually the standard Python idiom for this ([docs.python.org/3/reference/datamodel.html#object.__hash](https://docs.python.org/3/reference/datamodel.html#object.__hash)__): if a class overrides **eq** and shouldn't be hashable, this is how you make that explicit.

Right on, thanks for the explanation. Approved!

PR #2340, opened 2026-09-10
feat(drivers-vector-store): support regular inserts in BaseVectorStoreDriver

### LINKED ISSUE #50: Support regular inserts in `BaseVectorStorageDriver`

krisnaparahita:

## Describe your changes

Adds an `insert: bool = False` keyword argument to `BaseVectorStoreDriver.upsert()`. When enabled without an explicit `vector_id`, it generates a fresh `uuid4()` identifier instead of the deterministic content hash, allowing identical text to be inserted without overwriting the earlier entry.

The option works through `upsert_collection()` and concrete drivers through existing keyword forwarding. A shared `test_upsert_insert` test covers the new behavior; the full vector-driver suite passes (1,022 tests).

Fixes #50

### DIFF
--- a/griptape/drivers/vector/base_vector_store_driver.py
+++ b/griptape/drivers/vector/base_vector_store_driver.py
@@ -74,7 +74,7 @@ class BaseVectorStoreDriver(SerializableMixin, FuturesExecutorMixin, ABC):
         **kwargs,
     ) -> str:
         warnings.warn(
-            "`BaseVectorStoreDriver.upsert_text` is deprecated and will be removed in a future release. `BaseVectorStoreDriver.upsert` is a drop-in replacement.",
+            "`BaseVectorStoreDriver.upsert_text_artifacts` is deprecated and will be removed in a future release. `BaseVectorStoreDriver.upsert` is a drop-in replacement.",
             DeprecationWarning,
             stacklevel=2,
         )
@@ -138,6 +138,7 @@ class BaseVectorStoreDriver(SerializableMixin, FuturesExecutorMixin, ABC):
         namespace: str | None = None,
         meta: dict | None = None,
         vector_id: str | None = None,
+        insert: bool = False,
         **kwargs,
     ) -> str:
         artifact = TextArtifact(value) if isinstance(value, str) else value
@@ -145,8 +146,15 @@ class BaseVectorStoreDriver(SerializableMixin, FuturesExecutorMixin, ABC):
         meta = {} if meta is None else meta
 
         if vector_id is None:
-            value = artifact.to_text() if artifact.reference is None else artifact.to_text() + str(artifact.reference)
-            vector_id = self._get_default_vector_id(value)
+            if insert:
+                # Bypass the content-hash dedup below so re-adding the same text creates a new entry
+                # instead of overwriting the prior one.
+                vector_id = str(uuid.uuid4())
+            else:
+                value = (
+                    artifact.to_text() if artifact.reference is None else artifact.to_text() + str(artifact.reference)
+                )
+                vector_id = self._get_default_vector_id(value)
 
         meta = {**meta, "artifact": artifact.to_json()}
 
--- a/tests/unit/drivers/vector/test_base_vector_store_driver.py
+++ b/tests/unit/drivers/vector/test_base_vector_store_driver.py
@@ -26,6 +26,13 @@ class TestBaseVectorStoreDriver(ABC):
 
         assert len(driver.entries) == 2
 
+    def test_upsert_insert(self, driver):
+        id1 = driver.upsert(TextArtifact(value="foobar"), insert=True)
+        id2 = driver.upsert(TextArtifact(value="foobar"), insert=True)
+
+        assert id1 != id2
+        assert len(driver.entries) == 2
+
     def test_upsert_multiple(self, driver):
         driver.upsert_collection({"foo": [TextArtifact("foo")], "bar": [TextArtifact("bar")]})
 


griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -74,7 +74,7 @@ def upsert_text(
         **kwargs,
     ) -> str:
         warnings.warn(
-            "`BaseVectorStoreDriver.upsert_text` is deprecated and will be removed in a future release. `BaseVectorStoreDriver.upsert` is a drop-in replacement.",
+            "`BaseVectorStoreDriver.upsert_text_artifacts` is deprecated and will be removed in a future release. `BaseVectorStoreDriver.upsert` is a drop-in replacement.",
```
collindutter: Was this intentional? The original comment seems correct..

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -138,15 +138,23 @@ def upsert(
         namespace: str | None = None,
         meta: dict | None = None,
         vector_id: str | None = None,
+        insert: bool = False,
```
collindutter: Hmm I'm not sure about this approach. Using upsert to insert feels a bit off. I think we just want a proper `insert` method. e.g.

```python
   def insert(self, value, *, namespace=None, meta=None, **kwargs) -> str:
       return self.upsert(value, namespace=namespace, meta=meta, vector_id=str(uuid.uuid4()), **kwargs)

   def insert_collection(self, artifacts, *, meta=None, **kwargs): ...  # same fanout, calls self.insert
```

collindutter (commented):
Thanks for the PR! Some initial thoughts.

griptape/drivers/vector/base_vector_store_driver.py
krisnaparahita: Thanks for the feedback. I will replace the insert=True approach with dedicated insert() and insert_collection() methods, and Tests and PR description will be updated later. Please take another look once I update it.

griptape/drivers/vector/base_vector_store_driver.py
krisnaparahita: I will revert the unrelated deprecation warning change (that was unintentional).

griptape/drivers/vector/base_vector_store_driver.py
collindutter: Sure thing, please just re-request my review when it's ready.

krisnaparahita:
@collindutter could you please review again?

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -131,6 +131,57 @@ def upsert_collection(
...
+                    if not futures_dict.get(namespace):
+                        futures_dict[namespace] = []
+
+                    futures_dict[namespace].append(
+                        futures_executor.submit(
+                            with_contextvars(self.insert), a, namespace=namespace, meta=meta, **kwargs
+                        )
+                    )
+
+            return utils.execute_futures_list_dict(futures_dict)
```
collindutter, line 231: This is a full copy of `upsert_collection`. Can you please refactor this a bit?

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -131,6 +131,57 @@ def upsert_collection(
...
+    @overload
+    def insert_collection(
+        self,
+        artifacts: dict[str, list[TextArtifact]] | dict[str, list[ImageArtifact]],
+        *,
+        meta: dict | None = None,
+        **kwargs,
+    ) -> dict[str, list[str]]: ...
+
+    def insert_collection(
```
collindutter, line 177: Please add docs for these.

griptape/drivers/vector/base_vector_store_driver.py
```diff
@@ -154,6 +205,16 @@ def upsert(
...
+    def insert(
+        self,
+        value: str | TextArtifact | ImageArtifact,
+        *,
+        namespace: str | None = None,
+        meta: dict | None = None,
+        **kwargs,
+    ) -> str:
+        return self.upsert(value, namespace=namespace, meta=meta, vector_id=str(uuid.uuid4()), **kwargs)
+
```
collindutter: Nit: please place this method near the other insert methods.

tests/unit/drivers/vector/test_base_vector_store_driver.py
```diff
@@ -26,6 +27,30 @@ def test_upsert(self, driver):
 
         assert len(driver.entries) == 2
 
+    def test_insert(self, driver):
+        id1 = driver.insert(TextArtifact(value="foobar"))
+        id2 = driver.insert(TextArtifact(value="foobar"))
+
+        assert id1 != id2
+        assert UUID(id1).version == 4
+        assert UUID(id2).version == 4
```
collindutter: We don't need to test this, let's drop

collindutter (changes requested):
Thanks for the second revision. Looks much better, mostly cleanup nits.

krisnaparahita:
I have add the revisions, please review @collindutter

griptape/drivers/vector/base_vector_store_driver.py
collindutter: @krisnaparahita I still don't see any docs, did those get pushed up in your latest revision?

krisnaparahita: Sorry I was misunderstandin, I have push the latest commit. please check @collindutter

tests/unit/drivers/vector/test_base_vector_store_driver.py
collindutter: Looks like you removed the test for `insert`. Let's add that back please.

docs/griptape-framework/drivers/vector-store-drivers.md
```diff
@@ -7,13 +7,29 @@ search:
 
 Griptape provides a way to build drivers for vector DBs where embeddings can be stored and queried. Every Vector Store Driver implements the following methods:
 
-- `upsert()` for updating or inserting new text, [TextArtifact](../../reference/griptape/artifacts/text_artifact.md)s or [ImageArtifact](../../reference/griptape/artifacts/text_artifact.md)s into vector DBs. The method will automatically generate embeddings for a given value.
+- `upsert()` for updating existing entries or inserting new text, [TextArtifact](../../reference/griptape/artifacts/text_artifact.md)s or [ImageArtifact](../../reference/griptape/artifacts/text_artifact.md)s with an explicit ID into vector DBs. The method will automatically generate embeddings for a given value.
```
collindutter: >  with an explicit ID

This is not required.

```suggestion
- `upsert()` for inserting or updating text, [TextArtifact](../../reference/griptape/artifacts/text_artifact.md)s or
 [ImageArtifact](../../reference/griptape/artifacts/image_artifact.md)s in vector DBs by ID. If `vector_id` omitted, the method derives a deterministic ID from content. The method will automatically generate embeddings for a given value.
```

collindutter: Let's fix this incorrect link while we're in here:

`[ImageArtifact](../../reference/griptape/artifacts/text_artifact.md)` should point to `../../reference/griptape/artifacts/image_artifact.md`

docs/griptape-framework/drivers/vector-store-drivers.md
```diff
@@ -7,13 +7,29 @@ search:
...
 - `upsert_collection()` for performing an `upsert()` in parallel.
+- `insert()` for inserting new text, [TextArtifact](../../reference/griptape/artifacts/text_artifact.md)s or [ImageArtifact](../../reference/griptape/artifacts/text_artifact.md)s into vector DBs with an automatically generated ID. The method will automatically generate embeddings for a given value.
```
collindutter: Same comment as https://github.com/griptape-ai/griptape/pull/2340/changes#r4086218690

docs/griptape-framework/drivers/src/vector_store_drivers_13.py
collindutter: Thanks for adding this. This doesn't really show why you'd use insert over upsert. Upsert would have the same behavior here.

Maybe we replace with something like this:


 ```python
import os

from griptape.drivers.embedding.openai import OpenAiEmbeddingDriver
from griptape.drivers.vector.local import LocalVectorStoreDriver

embedding_driver = OpenAiEmbeddingDriver(api_key=os.environ["OPENAI_API_KEY"])
vector_store_driver = LocalVectorStoreDriver(embedding_driver=embedding_driver)
text = "Griptape is an enterprise AI framework."

vector_store_driver.upsert_collection({"upsert": [TextArtifact(text), TextArtifact(text)]})
vector_store_driver.insert_collection({"insert": [TextArtifact(text), TextArtifact(text)]})

print(f"upsert entries: {len(vector_store_driver.load_entries(namespace='upsert'))}")
print(f"insert entries: {len(vector_store_driver.load_entries(namespace='insert'))}")
 ```

So the output is
 
```text
upsert entries: 1
insert entries: 2
 ```

**Review by collindutter — CHANGES_REQUESTED** [review]
Looking better, just a few more minor things!

File: docs/griptape-framework/drivers/vector-store-drivers.md [hunk 9]
```diff
@@ -7,13 +7,29 @@ search:
 
 Griptape provides a way to build drivers for vector DBs where embeddings can be stored and queried. Every Vector Store Driver implements the following methods:
 
-- `upsert()` for updating or inserting new text, [TextArtifact](../../reference/griptape/artifacts/text_artifact.md)s or [ImageArtifact](../../reference/griptape/artifacts/text_artifact.md)s into vector DBs. The method will automatically generate embeddings for a given value.
+- `upsert()` for inserting or updating text, [TextArtifact](../../reference/griptape/artifacts/text_artifact.md)s or [ImageArtifact](../../reference/griptape/artifacts/image_artifact.md)s in vector DBs by ID. If `vector_id` omitted, the method derives a deterministic ID from content. The method will automatically generate embeddings for a given value.
```
collindutter:
```suggestion
- `upsert()` for inserting or updating text, [TextArtifact](../../reference/griptape/artifacts/text_artifact.md)s or [ImageArtifact](../../reference/griptape/artifacts/image_artifact.md)s in vector DBs by ID. If `vector_id` is omitted, the method derives a deterministic ID from content. The method will automatically generate embeddings for a given value.
```

tests/unit/drivers/vector/test_base_vector_store_driver.py
```diff
@@ -26,6 +27,32 @@ def test_upsert(self, driver):
 
         assert len(driver.entries) == 2
 
+    def test_insert(self, driver):
+        id1 = driver.insert(TextArtifact(value="foobar"))
+
+        assert UUID(id1).version == 4
```
collindutter: Looks like this assert got added back accidentally. We want the test, but not the assert. uuidv4 is an internal implementation detail, not something we want to assert on.

collindutter (changes requested):
Two _very_ minor points of feedback! Then I think we're good.

tests/unit/drivers/vector/test_base_vector_store_driver.py
krisnaparahita: done

krisnaparahita: done

docs/griptape-framework/drivers/vector-store-drivers.md
krisnaparahita: done

collindutter (approved):
Thank you again for the PR and for working through the review with me!

